PR #1, opened 2023-07-21
ucan interop with external services

Gozala:
[🪧 HTML View](https://github.com/web3-storage/RFC/blob/ucan-interop/rfc/ucan-interop.md)

I have been thinking how to make "sign the deal with wallet" API that would not require selling UCANs to spade team, yet avoid some secure channel between services (current spade approach) to provide some access control. I have captured some of this thinking in https://github.com/web3-storage/specs/pull/67/files

Talking with @vasco-santos earlier today I also realized we need similar thing for "failed deal" reporting, and unless misunderstood Spade is hoping storage providers can report failures directly to us instead of proxying such reports through some secure channel. So basically same challenge here.

I end up formalizing my thoughts around what are the ways ucan enabled system can interop with ucan disabled system and what are the tradeoffs. Would love your comments

### DIFF
--- a/rfc/ucan-interop.md
+++ b/rfc/ucan-interop.md
@@ -1,0 +1,101 @@
+# Abstract
+
+Services that have adopted UCANs are able to server requests to customers of their customers directly, reducing operational costs and complexity.
+
+However integrating with systems that have not adopted ucans can be challenging and forces following trade-offs:
+
+1. Privileged proxy services need to be introduced to allow requests from such actors.
+2. It is impossible to provide provable trace of accountibility in [recieipts](https://github.com/ucan-wg/invocation#225-receipt).
+
+In practice service is forced to introduce a less secure backdoor into the system.
+
+Here we explore various ways interoperabality could be achived with UCAN disabled systems and tradeoffs.
+
+## JWT Bridge
+
+UCAN enabled system identified by `did:web:system.io` creates an intermidiery principal e.g. `did:key:ZBridge` for interoperability with UCAN disabled systems.
+
+UCAN enabled system creates a principal `did:key:zOutsider` to represent UCAN disabled actor with-in UCAN system.
+
+System grants access to set of capabilities to an outside system by creating UCAN delegation chain
+
+```mermaid
+flowchart LR
+  did:web:system.io --> did:key:zOutsider --> did:key:zBridge
+```
+
+UCAN is formatted as JWT Token and passed to outside system that it can use as an authorization bearer token.
+
+Outside system attaches JWT Token to issued requests. The `did:key:zBridge` decodes request body and attached UCAN (JWT Token) and issues [UCAN Invocation](https://github.com/ucan-wg/invocation) forming following chain
+
+```mermaid
+flowchart LR
+  did:web:system.io --> did:key:zOutsider --> did:key:zBridge --> did:web:system.io
+```
+
+> ℹ️ Note that `did:key:zBridge` does not require access to any outside state besides own private key.
+
+The `did:key:zBridge` actor dispatches UCAN invocation to a UCAN enabled system `did:web:system.io` and responds with a UCAN Invocation Receipt back to the outside system.
+
+### Modes
+
+#### Session Token
+
+JWT Token can be long-lived and shared with outside system out of bound ahead of time. In this mode everything works as conventional JWT.
+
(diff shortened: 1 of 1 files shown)


rfc/ucan-interop.md
```diff
@@ -0,0 +1,101 @@
+# Abstract
+
+Services that have adopted UCANs are able to server requests to customers of their customers directly, reducing operational costs and complexity.
```
vasco-santos, line 3:
```suggestion
Services that have adopted UCANs are able to serve requests to customers of their customers directly, reducing operational costs and complexity.
```

rfc/ucan-interop.md
```diff
@@ -0,0 +1,101 @@
...
+
+However integrating with systems that have not adopted ucans can be challenging and forces following trade-offs:
+
+1. Privileged proxy services need to be introduced to allow requests from such actors.
+2. It is impossible to provide provable trace of accountibility in [recieipts](https://github.com/ucan-wg/invocation#225-receipt).
```
vasco-santos, line 8:
```suggestion
2. It is impossible to provide provable trace of accountability in [receipts](https://github.com/ucan-wg/invocation#225-receipt).
```

rfc/ucan-interop.md
```diff
@@ -0,0 +1,101 @@
...
+⚠️ Downside of this mode is that such tokens are easier to leak as they are likely to be passed and stored along with task requests. On the other hand amount of damage they can do has far more limited surface.
+
+### Tradeoffs
+
+- 💔 Security hinges on the ability of the outside system to keep JWT safe and secure.
+- 💔 Setup assumes trust in faceful operation of the `did:key:zBridge` actor.
+- 💚 The `did:key:zBridge` actor could be operated by third party or even outside system itself.
+- 💚 Outside system does not need to understand or worry about UCANs.
+- 💔 Impossible to hold outsyde system accountable.
+- 💔 No redelgation
```
vasco-santos, line 62: I think here is more around the UCAN adoption than a tradeoff of this solution? Would not expect a interop system would give full capability of using UCANs, otherwise this would replace the need for "selling" UCANs

vasco-santos, line 62:
```suggestion
- 💔 No redelegation
```

rfc/ucan-interop.md
```diff
@@ -0,0 +1,101 @@
...
+
+
+### Tradeoffs
+
+- 💚 JWT tokens can be shared in public
+- 💗 The `did:key:zBridge` can be hold accountable if it misbehaves.
+- 💚 The `did:key:zBridge` actor could be operated by anyone.
+- 💚 Outside system does not need to understand or worry about UCANs.
+- 💚 Outside system can be held accountable thanks to captured signatures.
+- 💔 Can not take advantage of redelgation
```
vasco-santos, line 99:
```suggestion
- 💔 Can not take advantage of redelegation
```

vasco-santos (commented):
@Gozala thanks for putting this together so quickly!

I would not expect that a interop layer with non UCAN entities would have redelegation has a requirement. If it had, there would even be less incentive to build UCAN based system for other parties.

With the above in mind, The PK solution seems quite great. That model seems to work quite nicely for all the use cases I have in mind, while offering all the security aspects (and even accountability)

Gozala:
I had been chatting with @gobengo on related thing which inspired another mode of operation that I'd like to incorporate here:

Lets assume there is an actor that simply wants oath like experience, that is public token and secret token. We could support that as follows:

1. Actor generates random seed token on their own.
2. Actor generates keypair form the seed token (we could teach w3 cli this trick)
3. They generate UCAN delegation with desired access level from their space (or account) to the public key derived from the secret token.
4. They POST to the RPC gateway endpoint passing UCAN delegation as JWT bearer token. They pass seed token as a secret token and capability `{ can, with, nb }` as JSON payload .
5. Gateway derives keypair from the seed.
6. Creates an invocation with passed capability and ucan as proof UCAN and signs it with the derived private key  handing off invocation to w3up API endpoint.
7. Gateway extracts receipt from from w3up response and passes it back to the actor

gobengo:
> They pass seed token as a secret token and capability { can, with, nb } as JSON payload .

could use `nonce` or `login_hint` in the oauth authorization-request

PR #2, opened 2023-07-21
Create group-by-space-provider.md

Gozala:


rfc/group-by-space-provider.md
```diff
@@ -0,0 +1,58 @@
+# Group by provider / sign by tenant
+
+by: @vasco-santos
+
+ https://filecoinproject.slack.com/archives/C02BZPRS9HP/p1689950046118329?thread_ts=1689873032.504899&cid=C02BZPRS9HP
+
+building a bit on top of this, and thinking about the flow of producers send us pieces, looks like this would fit as part of the w3-aggregation spec as a UCAN capability.
+I think later on (after migration) it can be done as an effect of `store/add` together with `piece/build` (or similar).
```
Gozala, line 8: I think capturing that in the effect is a cool ideal, as it will appear in the receipt and would allow recipient to look it up.

We should also consider how clients could submit `piece` CIDs to us. Perhaps we could have `piece/add` that users could invoke and have `piece/verify` instead of `piece/build` that way if we already have a car and piece we could avoid that effect.

In the future we could also use `piece/add` in place of S3 event to kick things off, and if user never did `piece/add` then it does not ends up in filecoin and that on them.

rfc/group-by-space-provider.md
```diff
@@ -0,0 +1,58 @@
...
+So here, I was thinking about having a `piece/offer` invocation where w3up after building the piece CID, as well as CF hosted APIs would invoke `piece/offer` that wouldbe handled by w3filecoin either responding queued or done if actually we already have it in an aggregate.
```
Gozala, line 9: That sounds good

rfc/group-by-space-provider.md
```diff
@@ -0,0 +1,58 @@
...
+What I still fail to capture is how would we tie this with a provider to set message group id. Probably this needs to be a `nb` field? I guess on older CF hosted APIs, it would just be hardcoded with the free provider for each product.
+w3up could be more intelligent and check which space `store/add` used for that CAR cid and then get registered provider for the space.
```
Gozala, line 11: I was thinking that users should interact with a system by tenant DID, meaning invocation `aud` is a tenant DID like `web3.storage` or `nft.storage`. That was also how I was thinking we could avoid having to [introduce tags](https://github.com/web3-storage/w3up/issues/167) for data stored / uploaded. It also ties nicely with [ucan invocation router](https://github.com/web3-storage/RFC/blob/ucan-invocation-router/rfc/ucan-invocation-router.md) idea.

However group resolution logic needs bit more though because `tenant:capability-provider` is `1:n` not `1:1` and `space:capability-provider` is also `1:n`.

I am actually not sure how are going to signal which capability-provider should handle the upload when space has multiple ones associated. Perhaps `aud` should be a `capability-provider` and not a tenant in which case everything would fit right in. E.g.  `free.web3.storage` and `basic.web3.storage` could be an `aud`. 

I'm actually bit fuzzy on DB details now worth asking @travis as that's probably in his headspace

rfc/group-by-space-provider.md
vasco-santos: ok, so `piece/add` followed by `piece/verify` that would trigger the pipeline if verified. Still need to think more on how we would skip verify in some steps. Maybe invocation is just triggered and the handler can decide to skip and state reasoning in receipt

rfc/group-by-space-provider.md
vasco-santos: @travis could we give us more details on what would you think here?

Gozala: @vasco-santos the linked issue captures what me and @travis came to agree was a good direction. Also I think @travis may not be seeing this thread in notifications so might be worth messaging him out of bound.

travis: Am I understanding correctly that the issue here is that we want to do "Filecoin stuff" differently depending on which plan a user has? 

From a user perspective I was imagining something like the following:

1) Alice comes in and signs up and gets a `free.web3.storage` plan by default, which gives them (eg) 100MB of storage
2) Alice uploads a 50MB file, Alice has one upload provider, we use it, everything works, Alice is happy
3) Alice uploads a 60 MB file, the system tells her "sorry, you don't have enough storage in your plan, but you can sign up for a basic plan for $10/mo that "
4) Alice signs up for a basic plan, retries the upload, it works!

I think the issue is that we would want to get the second upload (the 60MB one) into Filecoin ASAP since it is "covered" by the basic plan, but the first upload might not be as high a priority (or in the future might not get added to Filecoin at all?) since it was uploaded as part of the free plan - is that right or am I misunderstanding?

It seems like this comes down to whether we want to push the complexity of picking a "storage plan" that an upload should be associated with out to the clients, or whether we want to commit to heuristics that we would use on our end to make decisions like the one in the previous paragraph - does that sound right? I could see advantages either way, and I do think we could build our client libraries to make some of those heuristic decisions, but overall I think I prefer trying to hide this complexity from clients and keep the protocol API a little "simpler" for end-users?

That said I do think either could work! The one thing I really want to avoid is forcing `w3cli` or `w3ui` users to manually choose a plan for each upload - that feels like unnecessary and surprising friction that could be detrimental to people using and adopting `w3up` - I think we can avoid that whatever direction we go though, so that's not a hard blocker afaik.

Am I on the mark here or misunderstanding the conversation entirely?

vasco-santos: @travis yes that is the overall discussion here. 

Per https://github.com/web3-storage/w3up/issues/839 , proposal seem to be to have invocations like `store/add`, the `space/allocate` invocation would receive the tenant information related to the space where `store/add` is going to allocate space. This way, looks like option would be to on `store/add` invocation, the backend will look at the current plan for the provider (based on audience, so that we support space with multiple tenants) and:
- write in `store-table` the associated provider and plan
- return on `store/add` invocation that information for the receipt

That way, when we receive the CAR file, and `piece/add` is computed the information exists on our table and we can proceed with the Filecoin piece with the appropriate SLA

rfc/group-by-space-provider.md
```diff
@@ -0,0 +1,58 @@
...
+
+so w3filecoin would be aggregator, w3s/nft would be storefront, and:
+
+```mermaid
+flowchart LR
+
+user --store/add--> storefront
+storefront --piece/build--> storefront
+storefront --piece/offer--> aggregator
+aggregator --aggregate/offer--> authority
```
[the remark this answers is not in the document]
vasco-santos: yes

rfc/group-by-space-provider.md
travis: ok cool - that makes sense to me! I have a concern I haven't been able to fully articulate about capturing the "plan" in the store-table, but as long as the semantics are "this is the plan that was active/current when the file was uploaded" I think this sounds great.

PR #6, opened 2023-08-08
AgentDataExport "serverless" persistence strategy

travis:
Customers using w3up from "serverless" functions must currently invoke and execute `access/claim` during
every invocation of their function, wasting valuable time and network resources. We propose a solution involving
the encoding of the `AgentDataExport` datastructure into a string suitable to be used as the value of an environment variable.

👁️ [HTML View](https://github.com/web3-storage/RFC/pull/6/files?short_path=5dc9b02#diff-5dc9b020c16810694ba8565c46b9a482eec3dc5ba546d8cf47370b457ca0703b)


travis:
ah - I'm going to close this - there's a better way to accomplish this:

https://gist.github.com/alanshaw/e949abfcf6728f590ac9fa083dba5648

PR #7, opened 2024-01-25
proposal: consistent error handling with `Ucanto.Result`

travis:
@alanshaw @Gozala and I had a quick huddle about `Ucanto.Result` inspired by a discussion in `w3infra` (https://github.com/web3-storage/w3infra/pull/318#discussion_r1464989251) and I committed to writing up a proposal describing the tradeoffs of using _only_ `Result` for error handling vs using `Result` plus normal JavaScript error handling.

I've proposed we standardize on `Result` purity and offered to take a pass through the code to ensure we are wrapping all possible thrown errors in error `Result`s but want to get feedback from the team before I do that!

Comments sincerely requested - I, for one, could pretty easily be swayed to the "error `Result`s for known errors, thrown errors for unexpected stuff" camp, so if you feel strongly please let us know!

### DIFF
--- a/rfc/ucanto-result.md
+++ b/rfc/ucanto-result.md
@@ -1,0 +1,206 @@
+## Abstract
+
+Ucanto's `Result` type gives us the ability to use the TypeScript type system to specify the errors a function may return. We have started using `Result`s in a variety of different interfaces in our system, but have not been careful to catch and wrap runtime errors with specified error types in all cases. This has lead to functions that require callers to implement two different types of error handling - detection and handling of "error" `Result`s like `{error: {name: 'AnError', message: 'A message describing the error'}}` as well as standard `try...catch` error handling. We describe the tradeoffs and suggest that best practice is to coerce all errors into `Result` types in functions that return `Result`.
+
+## `Ucanto.Result`
+
+Ucanto `Result`s are described by the [UCAN invocation specification](https://github.com/ucan-wg/invocation/#6-result) and [implemented as a TypeScript `type` in `@web3-storage/ucanto`](https://github.com/web3-storage/ucanto/blob/main/packages/interface/src/lib.ts#L506). 
+
+Briefly, a `Ucanto.Result` is a TypeScript `Record` with EITHER an `ok` or `error` key. If `ok` is defined, `error` must be `undefined`, and vice-versa. Defining an function that returns `Result` allows us to specify possible error types for that function, which enables conveniences like IDE auto-complete and helps clients understand what types of errors they should expect to handle:
+
+```typescript
+interface PlanGetSuccess {
+  updatedAt: ISO8601Date
+  product: DID
+}
+
+interface PlanNotFound extends Ucanto.Failure {
+  name: 'PlanNotFound'
+}
+
+type PlanGetFailure = PlanNotFound
+
+interface PlansStorage {
+  /**
+   * Get plan information for a customer
+   *
+   * @param account account DID
+   */
+  get: (
+    account: AccountDID
+  ) => Promise<Ucanto.Result<PlanGetSuccess, PlanGetFailure>>
+}
+
+// using a Result from a normal TypeScript function
+async function getProductForAccount(plansStorage: PlansStorage, account: AccountDID): Promise<DID | null> {
+  const result = await plansStorage.get(account)
+  if (result.ok) {
+    return result.ok.product
+  } else if (result.error.name === 'PlanNotFound' ) {
+    return null
+  } else {
+    throw result.error
+  }
+}
+```
+
+## Mixed Error Handling Considered Annoying
+
+`Ucanto.Result` represents an alternative error handling paradigm for {Java|Type}Script that has some advantages in a typed environment. However, without care to wrap and handle unexpected errors in a function that returns a `Result` it can make comprehensive error handling annoying for callers. This becomes especially apparent when using multiple layers of `Result`-returning functions:
+
+```typescript
+interface CustomerGetSuccess {
+  email: string
+  plan: DID
+}
+
+interface CustomerNotFound extends Ucanto.Failure {
+  name: 'CustomerNotFound'
+}
+
+interface UnexpectedError extends Ucanto.Failure {
+  name: 'UnexpectedError'
+}
+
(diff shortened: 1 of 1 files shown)


travis:
Rendered HTML link for your perusing pleasure here: https://github.com/web3-storage/RFC/blob/proposal/ucanto-result/rfc/ucanto-result.md

vasco-santos (commented):
yes! I agree, we also did that already in Filecoin and I was hoping to get it the default everywhere

rfc/ucanto-result.md
```diff
@@ -0,0 +1,210 @@
...
+
+Briefly, a `Ucanto.Result` is a TypeScript `Record` with EITHER an `ok` or `error` key. If `ok` is defined, `error` must be `undefined`, and vice-versa. Defining an function that returns `Result` allows us to specify possible error types for that function, which enables conveniences like IDE auto-complete and helps clients understand what types of errors they should expect to handle:
+
+```typescript
+interface PlanGetSuccess {
+  updatedAt: ISO8601Date
+  product: DID
+}
+
+interface PlanNotFound extends Ucanto.Failure {
```
alanshaw, line 21: Import `Ucanto` for clarity on how we get hold of these types?

rfc/ucanto-result.md
```diff
@@ -0,0 +1,210 @@
...
+}
+
+type PlanGetFailure = PlanNotFound
+
+interface PlansStorage {
+  /**
+   * Get plan information for a customer
+   *
+   * @param account account DID
+   */
```
alanshaw, line 32: I don't think the param doc is adding anything useful here.

```suggestion
  /**
   * Get plan information for a customer
   */
```

rfc/ucanto-result.md
```diff
@@ -0,0 +1,210 @@
...
+// using a Result from a normal TypeScript function
+async function getProductForAccount(plansStorage: PlansStorage, account: AccountDID): Promise<DID | null> {
+  const result = await plansStorage.get(account)
+  if (result.ok) {
+    return result.ok.product
+  } else if (result.error.name === 'PlanNotFound' ) {
+    return null
+  } else {
+    throw result.error
+  }
```
alanshaw, line 47: Probably worth mentioning using `Result.try` to de-verboseify this.

travis: ah good idea - I'm not sure the result is much less verbose, but it's a good idea to highlight this utility!

here's what I came up with: 

```typescript
async function getProductForAccount(plansStorage: PlansStorage, account: AccountDID): Promise<DID | null> {
  try {
    return Result.try(await plansStorage.get(account)).product
  } catch (err) {
    if (err.name === 'PlanNotFound') {
      return null
    } else {
      throw err
    }
  }
}
```

travis:
I discovered an annoyance with the "Results only" error handling paradigm today: I was trying to implement a method in an interface (`PlansStorage#initialize`) and realized during the implementation that I want to return an error that was not specified in the interface. ie, I wanted to return a "invalid input" when the "external account ID" was not a `stripe:` URI:

https://github.com/web3-storage/w3infra/pull/318/files#diff-87f9a6d13813682bc1d417cd032ae3317f0cc2669d19ab5e5558f7ee48e73d1eR23

I ran into the same problem updating `PlansStorage#set` and opted to return not-yet-specified error types:

https://github.com/web3-storage/w3infra/pull/318/files#diff-87f9a6d13813682bc1d417cd032ae3317f0cc2669d19ab5e5558f7ee48e73d1eR66

Both of these approaches seem like reasonable short-term ways to get the code running. The second approach means I need to go back and update `w3up/upload-api` to specify more error results. The first gets my code typechecking, but violates the proposal above.

We _could_ solve this by saying that interfaces like this should always accept `Ucanto.Failure` (ie, specify error types like `type MyError = ErrorA | ErrorB | Ucanto.Error`) but this means we lose some of the benefits of typing.

Exhaustively specifying possible error types in interfaces seems like the cleanest way to do this, but is annoying, especially in a case where updating an upstream interface is non-trivial (ie, because you don't control it, or something similar) and in this case I genuinely don't know what to do - can't throw, can't return an unexpected error, don't want to just eat the error...

@Gozala any thoughts on this?

Gozala:
https://www.typescriptlang.org/play?#code/C4TwDgpgBAShDOBXANsAPAFShAHsCAdgCbxQDeAvlALzkUA0UAGtnoSXTVAKIBOvAe14A+GgCgokqAB9yUAQGsAXFAyMI-IQH4VBCADcNUChKmyy8hTqh7DvdZt4qWJsW7cBLAvl4AzAIYAxtBw-kR8grys+MSkEULkppIE-gC2ECoA5KHhjplirl4+AcFQAOq8HvjxUbgxHDWJUjZpGVCZFVUQNfmuRBCByP680L6IBIHAHgIEUCNhABRg-sAAFkrwwJUEAOYAlCpwSKhoAKpFABwAgvz+IIw5NcJi-YPDo+OT07MA7pX4SxW6ygm22O0YRBW-hU52811uIAOsAQKHQlEYnWqjmeYkCM02ql4IC4aBg0XYpCOqMYWDqFLojBYdNiPEcXFsGmECzeiHgbQWexoogA4oQNCshKSHiiTmpmMJhAcqbLGAsYABtTIaSKZAC6UAAZHRBbImKJqKIyEkoHiCATgP54AouDy+QLrcgIMAQQ78FwHU6AHR6PDu5o-VYeT1QBYAQk2KwggaIMwggqtzWaHl8MYT+ED+n8yEQSe1QnT1szUhGwEQvFmeaTheLEErUioEGQfKaVczjf9joUwbqC0bBaLJcDij2bckJma8+rXrrDd9TYnrdcLwGQxGUDGEymMyg4z+-jAUEw5JZ6OY14a2IWI2OwEOMvQcrNSNFel4Et4pLvpgjLCDSohuLaBKQg6XAYESCwHl8MwAFQxhWzSQd6AiIN6tAgB4nZEKhp5-mAT4QIsmQAPS8hoVH+AARthwCBsAeCZHsM7NPhhHEQQZ5kX8XQLNRtG8FR+CpGArHsYwzGce4zQ1iuciKCoABMxgFHsQA

Issue #4, opened 2023-07-27
Storefront endpoint where SPs can request CAR files for each segment of an aggregate


vasco-santos:
we need to work on a proposal to have an endpoint by Storefront where SPs can request CAR files of each segment of the aggregate
- this behind the scenes will redirect to roundabout presigned URLs
- previously we were considering to just compute presigned URLs for each SP storefront signs the deal proposal. However, this means that we would need to move JSON files around with potentially 100.000s entries of presigned URLs.
- we will be providing a base URI instead and a auth token that SPs can use. WIth that, we validate auth token (with expiration) and redirect to presigned URL

PR #8, opened 2024-01-25
feat: merkle structure

Gozala:


Gozala:
Few notes based on the conversation with @mikeal yesterday

- @mikeal suggested to not worry about the data types and basically only have bytes, leaving it up to the user to encode them into bytes.
   - 🤔 I think there is a merit to the idea of layering things in a way that base layer would not care about data types only bytes and ordered lists of bytes. On top you could have a layer that knows about data types and provides mapping to bytes / lists of bytes.
   -  I am not however persuaded by the argument that there is no consensus how to encode various types. In fact I'd argue that it is better to have a decision how to encode data for hashing than have a flexibility to decide encoding. Encoding floats in double precision for hashing does not imply you should encode them that way on disk or wire. In fact for this use case it makes sense to optimize encode for speed, while for disk / network it's likely better to optimize for size.
- [PKV] double hashes keys. Initially that was my plan, but while writing this up I could no longer see what that would solve so I no longer do that. I'm still able to provide proofs without revealing attribute names so maybe binary merkle tree gives us property that we'd use double hashing for ?
- I like that I can stick lists `[a, b]` and `[c, d]` just by looking at hashes. but I don't like that `[[a, b], [c, d]] = [a, b, c, d]`. Perhaps we could byte prefix list addresses and retain former without the later.
- I'm not sure structure is the list of attributes is a good idea. Perhaps we should differentiate two. Just like ☝️ we could just byteprefix hash. Tradeoff is that we conceal less information, if we don't prefix you can't tell if reference is to a scalar or complex structure, and you also can not join structs and lists. If you do prefix then you can join structs and lists, but now you can tell if reference is to a struct / list or a scalar.


[PKV]:https://github.com/mikeal/pkv

Gozala:
@mikeal provided more valuable feedback that I'd like to  capture here before I'm able to incorporate. Specifically, he raised a concern about encoding type semantics just through byte prefix (tag), because it is plausible that someone may have byte array `[2, 193, 15]` which when hashed with sha256 will produce `bbnelb523uc3jtdsqegf5odzm3qm3jsxrwg65e66arecvfkep4b2a` which is coincidentally address for `1985` according to this spec.

Changes of such a collision can be greatly reduced if instead of byte prefixing we prefix with a lot more distinct byte sequence. e.g. instead of prefixing integers with `2` we could prefix them with UTF8 encoded  bytes for `LEB 128 formatted integer value`. In fact we could reduce probability even more if we actually take LEB128 spec and use sha256 of it as a prefix. Probability of collision there is greatly reduced.

Alternatively we could still use tags we have here but in addition prefix each with the name of this spec.

I think this is a good suggestion and worth incorporating here

Gozala:
Another interesting insight from @mikeal's work on https://mikeal.github.io/cryptography/public-data.html where he uses double hashing allowing you to decide whether you want to reveal contents of the link or not. If we were to adopt it here we would have an ability to decide whether we want to reveal what the data is (in which case we share first order hash) or conceal it by sharing second order hash.

Reminds me a lot of onion routing

rfc/merkle-structure.md
```diff
@@ -0,0 +1,320 @@
...
+
+```json
+{
+  "message": {
+    "from": "@gazala",
+    "to": "@mikeal",
+    "payload": "Hi"
+  }
+}
+```
```
Gozala, line 26: You can think of this as a following AST in lisp notation

```clj
(Record
   (Attribute
      (String/UTF-8 "message")
      (Record
         (Attribute
             (String/UTF-8 "from")
             (String/UTF-8 "@gozala"))
         (Attribute
             (String/UTF-8 "to")
             (String/UTF-8 "@mikeal"))
          (Attribute
              (String/UTF-8 "payload")
              (String/UTF-8 "Hi"))))
```

rfc/merkle-structure.md
```diff
@@ -0,0 +1,320 @@
...
+
+And this is just one dimension of variability, if we were to use [DAG-CBOR] encoding instead we would end up with `bafyreie7pbuqqem5fvmrh4wv4joy737tmgmoo5tn4q7qwocd2a52si2m6m` when `message` is externalized and with `bafyreie7pbuqqem5fvmrh4wv4joy737tmgmoo5tn4q7qwocd2a52si2m6m` otherwise.
+
+And this is not the end, we could have a case where `message` is in [DAG-JSON] and rest is in [DAG-CBOR] in which case our data will be addressed `bafyreig5bmydxjpbq5j46hqs5cnay7xs4yef2nu4zts4rdmgxjzlsh4ume`.
+
+ℹ️ IPLD shifted question from where content is to how is it encoded. It is better problem to have, but still a problem because how you want to encode it will depend on deployed system and undoubtedly will vary from case to case basis. It is also very likely that choice of encoding will change over time and change in addresses is a major drawback.
```
expede, line 34: Yeah agreed — this is very annoying. I would like a _canonical_ hash, and make encoding a separate concern (heck, even tag the encoding you mean in the header). I think the intention was to make verification trivial, but then have one compact representation at the protocol layer. AFAIK the tension rises here when you want to represent data from other sources directly, but I haven't really seen that happen. You could always fall back to "it's just bytes" like in `dab-pb`, or select a structured representation as needed.

rfc/merkle-structure.md
```diff
@@ -0,0 +1,320 @@
...
+
+## Problems
+
+### Data Locality
+
+In most programming languages we build data structures out of primitive data types and more complex structures by nesting them into one another through [reference]s. In languages with managed memory you do not worry about whether you can dereference sub-structure reference in itself is implies access to it. Languages where you do need to worry about dereferencing tend to suffer from [null pointer] problems.
+
+Data structures in IPLD suffer from a similar set of problems as languages with
+[null pointer]s. In case of IPLD it is a locality problem. Nested sub-structures can either be **local**, inlined into a parent structure, or **external** linked using an IPLD link. Local references can always be dereferenced, but external references may require read from a disk, or another machine. In worse cases even searching the network to find machine that has it. Possibility of errors is inconvenient, but what's worse is the impact on the data model, imposing constraints of data exchange (block size) which are almost never universal.
```
expede, line 12: ...do CIDs impose limits on block size? I think this specific implementation does, but that's not a hard constraint, right?

rfc/merkle-structure.md
```diff
@@ -0,0 +1,320 @@
...
+## Proposition
+
+Now that we have outlined problems lets enumerate what would we desire instead:
+
+1. Data structure to be addressed by its contents regardless of location, whether some parts are inlined (local) or external. This will give us flexibility to decide which parts to inline vs externalize when storing and decide differently when transferring it.
+2. Data structure to be addressed by its content regardless of the encoding. This will give us flexibility to choose different encoding on disk from encoding on wire. In fact, we gain ability to negotiate content encoding per data exchange.
+
+### Proposed Addressing
+
+IPLD data model offers same data types as pretty much every mainstream programming language. We can discard link type due to issues it introduces. We also drop `null` because it is a bad idea and in most cases custom representation of "absence of value" is better choice.
```
expede, line 45: A single canonical representation that Merklizes _everything_, and has well defined Merklization of the data model types is a Very Good Idea:tm:

However, if I'm reading this correctly, you're saying that instead of an explicit link, we can a use can then break up the structure arbitrarily. I like that idea in general, but doesn't it mean that _any part of the structure_ can now have cache locality concerns? There's no way to express "if you have `a` you also definitely have `b`", right?

Hashes are very helpful in distributed systems, to the point that some top-tier DS researcher are working on a hash-based distributed operating system. Sure, everything is wrapped in a promise, requests can fail, time out, etc etc: that's how all DS work, and actor systems like Erlang & Elixir have been very successful with this model (using internal process IDs instead of CIDs, but close enough for this discussion).

I want to maintain structural sharing, both for usual data structure reasons, _but especially for caching._ For example: I cannot send the entire UCAN chain every time; I _must_ be able to omit parts of that payload. Being that it's signed over, I can't arbitrarily change the structure to suit serialisation needs.

# General Commentary

> We also drop `null` because it is a bad idea and in most cases custom representation of "absence of value" is better choice.

I'm not 100% sure that I agree. Yes, there's the old adage about the billion dollar mistake, but as you note: omitting a field is semantically equivalent (it's `undefined` by analogy to JS, which is arguably _worse_).

Absence is not always the same as marking something as empty. Empty strings, empty arrays, and empty maps are all examples of other valid "none" types that cause headaches (e.g. see the `NonEmpty` list type in Haskell and Rust). Falling back to hacks like `{field: ""}` to mean `None` seems like an unnecessary loss of expressive power.

Since this is a flat serialisation format, there probably isn't a requirement to have special serialisation for coproducts (which parser from this format to the application's type would do the conversion on), so having a `null` fits the role of `None` on required fields.

rfc/merkle-structure.md
```diff
@@ -0,0 +1,320 @@
+# Abstract
+
+Inter Planetary Linked Data (IPLD) is a fundamental building block of the IPFS. IPLD Links (a.k.a CIDs) are quintessential, yet I'd argue a major design flaw (on par with null pointers ?)
```
expede, line 3: > Links (a.k.a CIDs) are quintessential, yet I'd argue a major design flaw (on par with null pointers ?)

Woah, I could not disagree more. Yes, you have to reason about the network, _but this is explicitly a distributed system_. We can't reason about the tradeoffs the same as a local structure, and trying to merge the two always has tradeoffs.

> Much of the pain in traditional distributed programming comes from this mismatch:  programmers are expected to bridge from an ordered programming model into a disordered reality that executes their code.
> 
> — BloomLang website

The four approaches are:

1. Try to make everything feel local even if remote
  * Fails because not everything is actually local
  * Simple in simple cases, extremely confusing and frustrating in complex ones
2. Try to make everything feel remote even if local
  * More-or-less "like Erlang"
  * Simple cases always have to handle the "what if it's not there" case, but complex cases are consistent
3. Separate the cases
  * Have a distinction between things that are expected to be together versus could be unlinked (i.e. like most of the internet / HTTP _and also_ IPLD)
4. Monotonic / eventual consistency
  * i.e. Who cares if the remote doc isn't available, you can work with partial information just fine and progressively enhance 

I would generally claim that `#1` is not a fit for distributed systems — full stop.

`#2` is very workable, but inconvenient (programmers are unhappy about the unnecessary complexity when they _know_ something is local).

`#3` is what most systems do today, It's an extremely successful model. With Merklized data, I think the problem comes up when _humans_ don't want to work with pointer indirection or you have a hard requirement to access data that you can't fetch / data availability. I would argue that data availability should be solved at a different layer (like they do on the the non-Merklized web — AWS has many products aimed at this problem). Human factors can be solved many ways (such as inline-ipld, but the class of solution is certainly broader than that).

`#4` is kind of subtype / solution to `#3`. If you're in an eventually consistent domain (which sure, many things are not, e.g. streaming a movie), then you can largely side step the issue.

expede: Without going too deep into a pure theory discussion, I'd also like to gesture at Mark Miller's Vat concept as another example of `#3`, and maybe in which some ideas could be pulled.

From his thesis "Robust Composition", which is more about a programming model than a storage layer (but you can adapt the ideas):

There's obviously a distinction between local and remote:

Screenshot 2024-02-03 at 12 13 50

Local calls stay inside a Vat (in his terminology), which is a sequentially executing VM/container that guarantees that ant agent inside it can call an of the others inside it, in a totally predictable order. Calling across Vats may fail or introduce nondeterministic asynchrony. Two Vats can be on the same machine, or networked — they're just an abstraction of how you think some agents will expect others to be guaranteed responsive not.

Screenshot 2024-02-03 at 12 14 21.

I think of IPLD this way today ("these fields all are guaranteed to move together"), but I wonder if there's other ways to slice the problem that we could borrow from the Vat idea?

rfc/merkle-structure.md
Gozala: CIDs do not impose limits but in practice limits are implied. New version of the PR goes into more detail on this.

rfc/merkle-structure.md
Gozala: > However, if I'm reading this correctly, you're saying that instead of an explicit link, we can a use can then break up the structure arbitrarily.

Idea is data is broken up completely all the way to the scalar level. Whether you capture all the hashes or just parts you care about is up to you and in that sense it is arbitrary.

> I like that idea in general, but doesn't it mean that any part of the structure can now have cache locality concerns? There's no way to express "if you have a you also definitely have b", right?

That is accurate and a fare concern. I need to think more about it. My immediate reaction is that belongs in a different layer.

> I want to maintain structural sharing, both for usual data structure reasons, but especially for caching. For example: I cannot send the entire UCAN chain every time; I must be able to omit parts of that payload. Being that it's signed over, I can't arbitrarily change the structure to suit serialisation needs.

I'm not sure I follow what you are trying to say here. I agree with what you are saying and I believe proposal here should fully support it, do you suggest otherwise ?

> No null

I'm amendable to adding null type, even though I personally feel that using custom `None` structure is generally a better idea.

rfc/merkle-structure.md
Gozala: I'm not disagreeing with what you are saying here. That said, what I was trying to convey is that backing whether something is local or not into data identifier is not a good idea.

What's described in the proposal is probably `#4` or perhaps `#2`, definitely not `#1`. Hopefully newer version conveys the point more clearly.

rfc/merkle-structure.md
```diff
@@ -0,0 +1,503 @@
...
+
+Inter Planetary Linked Data (IPLD) is a fundamental building block of the IPFS. IPLD Links (a.k.a CIDs) are quintessential, yet arguably suffer from a major design flaw (on par with null pointers ?)
+
+# Introduction
+
+[Content identifiers (CIDs)][CID] claim that identify data not by _where it is stored_, but rather by the _content itself_. In our observation however, it moves addressing from "where content is stored" to "how content is stored" more specifically to how data is formatted / encoded. It is a significant improvement, but not quite as addressing by "what content is".
+
+Same data may be encoded differently depending on where it is stored, different yet when it is moved over the wire. It is also likely that encoding choices may evolve over time. Encoding choice affecting a content identifier is a major drawback and counter to goal of addressing content by "what content is".
+
+Identifying data by how it is encoded has various problems and we will call out major ones to illustrate motivation for this proposal. Then we will define goals for the proposed data identifiers and then proceed with the proposal.
```
Gozala, line 13:
```suggestion
Identifying data by how it is encoded has various problems and we will call out major ones to illustrate motivation for this proposal. Then we will define goals for the proposed data identifiers and proceed with the proposal.
```

rfc/merkle-structure.md
```diff
@@ -0,0 +1,503 @@
+# Merkle References
+
+# Abstract
+
+Inter Planetary Linked Data (IPLD) is a fundamental building block of the IPFS. IPLD Links (a.k.a CIDs) are quintessential, yet arguably suffer from a major design flaw (on par with null pointers ?)
```
Gozala, line 5:
```suggestion
Inter Planetary Linked Data (IPLD) is a fundamental building block of the IPFS. IPLD Links (a.k.a CIDs) are quintessential, yet arguably suffer from a major design flaw.
```

Gozala:
After reading @expede feedback I realized I over emphasized data locality point, which this proposal does not really address (parts of data might still be elsewhere). What I meant to emphasize is that partition choices affecting content identifiers is problematic. We could instead have way to derive identifier for any data giving us freedom to choose how to partition on read as opposed to write.

I tried to adjust text accordingly

rfc/merkle-structure.md
```diff
@@ -0,0 +1,492 @@
...
+[*] --> bmjrgvd75uynefn3hljzkl2lg4xqthymoqolc22qwtxl2crew27fa: bmj…27fa
+bmjrgvd75uynefn3hljzkl2lg4xqthymoqolc22qwtxl2crew27fa --> b46btqxqak3bstb7pmf2kkbl2ht6zat2xub5lovb4oavki4uysd2q_bmjrgvd75uynefn3hljzkl2lg4xqthymoqolc22qwtxl2crew27fa  : b46…sd2q
+b46btqxqak3bstb7pmf2kkbl2ht6zat2xub5lovb4oavki4uysd2q_bmjrgvd75uynefn3hljzkl2lg4xqthymoqolc22qwtxl2crew27fa: merkle-structureːfloat/double-precision
+bytes_bmjrgvd75uynefn3hljzkl2lg4xqthymoqolc22qwtxl2crew27fa: 0x9C 0xC4 0x20…0x08 0x32 0x40
+bmjrgvd75uynefn3hljzkl2lg4xqthymoqolc22qwtxl2crew27fa --> bytes_bmjrgvd75uynefn3hljzkl2lg4xqthymoqolc22qwtxl2crew27fa
+```
+
+#### Bytes
+
+Bytes is represented as a [merkle fold] of pair of leaf nodes. Left node is the cryptographic hash of the UTF-8 encoded string `merkle-structure:bytes/raw` and describes binary encoding format of the right node. Right node is a byte array of the bytes.
```
Gozala, line 204: 🤔 I'm starting to wonder if the right node should be an actual bao tree here

rfc/merkle-structure.md
Gozala: I have added `Null`. Even if I think it's better to model your data differently, in practice many languages use and have `null` and forcing them to not to is not an objective here.

rfc/merkle-structure.md
```diff
@@ -0,0 +1,505 @@
...
+```
+
+Merkle fold of a single leaf is a leaf itself
+
+```mermaid
+graph TB
+only(leaf)
+``````
+
+Merkle fold of zero leaves is an empty hash (hash of zero bytes)
```
Gozala, line 109:
```suggestion
Merkle fold of zero leaves is a leaf that is byte array with 0 bytes.
```

I think this might be a better choice. Consequence is that empty list becomes double hash of the list operator. Also null abstract data format can be defined as null tag and empty tree.

hannahhoward (commented):
this is a random set of reactions.

First, the encoding problem is already implicitly recognized and handled in a very primitive way in a lot of existing code, by indexing data by multihash -- so that way I can request whatever encoding I want and as long as the multihash matches, the peer can serve it back to me. Which doesn't solve the CBOR->JSON translation issue (since that will fail on decode) but does let you switch back and forth wth RAW bytes. Just an interesting observational note.

Second the partitioning problem is real too -- and I think it has parallels in the design of JSON APIs. I think back to the way people tried to build Hypermedia JSON APIs for years, encoding sub resource links alongside data, on the theory it would enable smart clients to consume data how they wanted. But ultimately fixed partitioning was inefficient, and hard for backend providers to evolve over time. GraphQL ate their lunch by allowing clients to specify exactly how much data they needed dereferenced. (btw, there's probably a world where you could save bandwidth by intentionally referencing everything but the scalar fields you care about, or paginating long list or map elements)

Re: Null -- the billion dollar mistake in Java is not the idea of a null type -- it's making all types implicitly able to be the type OR null. This type of issue really only is relevant IMHO at the schema layer of IPLD, if you want to go there (i.e. when you're dealing statically typed data, as opposed to the basic data model of IPLD which is more dynamic)

Ok so thinking through the @expede 's UCAN scenario, I would imagine the signature field would be derived by signing the root hash of the payload reference tree with the users private key. That wouldn't change if you replaced parts of the proof chain with reference tree hashes.

BTW, there's no reference to how one determines whether a node is a leaf or a reference tree -- is that encoded in the has in way similar to BAO?

Also, do we imagine a reader being able to request data in arbritrary formats and arbitrary levels of depth of de-referencing? Asking cause that's a non trivial amount of work to ask the person holding data to do. You'd essentially have to define a GraphQL like protocol to describe requests from the client.

I will say while this design appears to solve many problems, it's also a sufficient mind bender and compatibility break that it's a bit frightening to think about. In the UCAN example above, the signature would not match either the current signature you'd get from a UCAN with a full delegation chain OR the signature you get with a UCAN with references.

One way to consider this moving forward is to look at it an application level, rather than a universal data format level. The application a lot of people want for IPFS is "Store my files" which doesn't need all this. (Maybe directories are a candidate but honestly that could benefit from similicity too). On the other hand, if I wanted to use IPLD as distributed content addressed graph database to power say a social application, wow this seems amazing.

In that context, it might make sense to define this as kind of encoding format of its own. Personally, I don't think IPLD is useful when not working with structured data in an application context, rather than a file context. (UnixFS, while technically using IPLD, is a super weird format that only supports one kind of serialization and one set schema for blocks -- I wished we simply thought of it as a seperate thing)

BTW, I like turning all bytes blocks into a Bao Tree cause it creates a nice symmetry in the format.

Gozala:
> this is a random set of reactions.
> 

Thanks for taking a look @hannahhoward I'll try to provide some notes inline or request for clarification.

> First, the encoding problem is already implicitly recognized and handled in a very primitive way in a lot of existing code, by indexing data by multihash -- so that way I can request whatever encoding I want and as long as the multihash matches, the peer can serve it back to me. Which doesn't solve the CBOR->JSON translation issue (since that will fail on decode) but does let you switch back and forth wth RAW bytes. Just an interesting observational note.

I would summarize this that it is solved at block level, but not at the DAG level. That is if you have links from that block you're still very limited in kind of things you can do.

> 
> Second the partitioning problem is real too -- and I think it has parallels in the design of JSON APIs. I think back to the way people tried to build Hypermedia JSON APIs for years, encoding sub resource links alongside data, on the theory it would enable smart clients to consume data how they wanted. But ultimately fixed partitioning was inefficient, and hard for backend providers to evolve over time. GraphQL ate their lunch by allowing clients to specify exactly how much data they needed dereferenced. (btw, there's probably a world where you could save bandwidth by intentionally referencing everything but the scalar fields you care about, or paginating long list or map elements)
>

I think implicit proposition in here is that partition is often a read time decision and not a write time one.

> 
> Re: Null -- the billion dollar mistake in Java is not the idea of a null type -- it's making all types implicitly able to be the type OR null. This type of issue really only is relevant IMHO at the schema layer of IPLD, if you want to go there (i.e. when you're dealing statically typed data, as opposed to the basic data model of IPLD which is more dynamic)

I did take feedback from @expede and @ribasushi on this and have added `Null`. I still think there is a better way to model data than introducing `Null` but it is not the goal here so `Null` it is. 

> 
> Ok so thinking through the @expede 's UCAN scenario, I would imagine the signature field would be derived by signing the root hash of the payload reference tree with the users private key. That wouldn't change if you replaced parts of the proof chain with reference tree hashes.
>

Correct if I understand you correctly. In other words `hash([a, b, c])` will be same as `hash([a, b, hash(c)])`. If you meant something else please clarify

> 
> BTW, there's no reference to how one determines whether a node is a leaf or a reference tree -- is that encoded in the has in way similar to BAO?

All data types are mapped to pair of the tag (format) and corresponding node. Tag will tell you whether the right node is a leaf or a subtree.

It's bit tricky to compare to blake3 (which is what I think you meant). Because blake3 just chunks byte array into leaves and builds up a tree. Here we map data into list of nodes, primitives end up with nodes with depth 1 and recursive types can be arbitrary depth. However each node of the list gets merkle folded into effectively a leaf node from which we build BAO tree.

It is bit confusing because you can look at each node of the list as subtree or a leaf. That said you can tell if what you have is partial DAG or complete because you can differentiate primitives from references

> 
> Also, do we imagine a reader being able to request data in arbritrary formats and arbitrary levels of depth of de-referencing? Asking cause that's a non trivial amount of work to ask the person holding data to do. You'd essentially have to define a GraphQL like protocol to describe requests from the client.
> 

Short answer is NO, but longer answer is it really depend case by case basis. I think of this as a way that enables provider to choose level of indexing they want to support. It is really up to them and their customers to decide what kind of granularity they want to support. That said they provider could always reduce granularity on demand but they can not support more granular requests than they have chosen to index by.

From W3Up perspective, we find that as a provider we are not in a good position to decide what to index and what to announce on the network. It really is domain specific and are thinking of ways to enable clients to choose what to index support corresponding reads. What is proposed here is similar in spirit I imagine user uploading 4GiB of CBOR data and then publishing verifiable index for parts within it.


> I will say while this design appears to solve many problems, it's also a sufficient mind bender and compatibility break that it's a bit frightening to think about. In the UCAN example above, the signature would not match either the current signature you'd get from a UCAN with a full delegation chain OR the signature you get with a UCAN with references.
>

That is accurate. That said however UCANs are moving into direction of removing proof chains from delegations anyway so it may not be as bad.

Different way to view this proposal might be to treat is a canonical addressing schema, you still could have mapping from canonical to one being used. If you do so you suddenly become able to transcode things on the fly if requests are in canonical addressing scheme.

> 
> One way to consider this moving forward is to look at it an application level, rather than a universal data format level. The application a lot of people want for IPFS is "Store my files" which doesn't need all this. (Maybe directories are a candidate but honestly that could benefit from similicity too). On the other hand, if I wanted to use IPLD as distributed content addressed graph database to power say a social application, wow this seems amazing.
>

I agree. This is really taking what Iroh does for files and applying it to structured data. Although there is though that maybe two can converge to a degree, specifically I have considered mapping primitive strings and bytes types into BAO in which case bytes representing file would result in the exact same hash as blake3 assuming blake hashing is used.

> 
> In that context, it might make sense to define this as kind of encoding format of its own. Personally, I don't think IPLD is useful when not working with structured data in an application context, rather than a file context. (UnixFS, while technically using IPLD, is a super weird format that only supports one kind of serialization and one set schema for blocks -- I wished we simply thought of it as a seperate thing)
> 

Agreed. I think UnixFS suffers from the same issues outlined here, that is partitioning and layout is decided on write (well more like ingestion) and constrained in size by incremental variability. I think blake3 / bao makes far better tradeoffs here.

Honestly just switch in mindset of UnixFS as a particular view over file might get us really far. In that case you can start comparing tradeoffs offered by that view vs blake3/bao. 

> BTW, I like turning all bytes blocks into a Bao Tree cause it creates a nice symmetry in the format.

👍

rfc/merkle-structure.md
Gozala: I'm convinced now that it should be.

Gozala: I'm convinced now that it should be, that way hash of the raw bytes would come out same as the blake3 hash of that content.

Gozala: Blake3 splits byte array into 1024 byte size chunks (last chunk may be shorter). If there is only one chunk that chunk is a the root.

Gozala: I dug deeper into blake3 & bao and sadly compatibility is probably not going to work out. The problem is that `leaf_hash(chunk, n)` and `node_hash(left, right, is_root)` are not defined in terms of `blake3` hash so we would either have to define same algorithm in the merkle-folding or would instead define them using the hashing function (as per spec today).

Seems to me that we could take advantage of all the benefits that blake3/bao offers with a generic hashing function which seems like a better tradeoff here.

That said it is probably worth evaluating how why chunk index `n` is used and if it would make sense to incorporate that and same for the `is_root` parameter.

Gozala: For the reference bao hash and blake3 hash come out the same

https://github.com/oconnor663/bao/blob/e5f01f8c5d11f653eac861ee1a72f0fae5728460/src/encode.rs#L1206-L1208

Although that is because bao leverages blake3 internals to compute leaf hash

https://github.com/oconnor663/bao/blob/e5f01f8c5d11f653eac861ee1a72f0fae5728460/src/encode.rs#L559-L562

And to compute the node hash

https://github.com/oconnor663/bao/blob/e5f01f8c5d11f653eac861ee1a72f0fae5728460/src/encode.rs#L297-L300

PR #10, opened 2024-02-09
docs: store deliver capability rfc

vasco-santos:


olizilla (changes requested):
The idea here is good!

The text needs cleaning up but I'd rather go through it together in sync when you have time.

rfc/store-deliver.md
```diff
@@ -0,0 +1,278 @@
...
+
+Aiming to introduce verifiability into content storing, as well as to pave the way for a future proof of delivery, `store/deliver` capability is proposed. It would also make services not tied with store put events to execute side effects.
+
+To offer verifiability, `store/add` invocation should be associated with `store/deliver` via [effect]s. `store/deliver` would be invoked by the client to notify the service that content was stored, and the the service should also sign a confirmation that it is true.
+
+### store/delivery of not previously stored content
+
+Taking into account the above, we can look at the following diagram and flow walkthrough:
+
+![store-deliver-0](./store-deliver/store-deliver-0.svg)
```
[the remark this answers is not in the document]
vasco-santos: yeah, adding numbers seems a good idea. I will do that :)

I think mermaid would bring other challenges with the receipts and OK, ERR. I will add the numbers for now

rfc/store-deliver.md
```diff
@@ -0,0 +1,278 @@
...
+}
+```
+
+Note that content requested to store, may already be stored. If that is the case, the service may issue receipt with `"status": "done"`.
+
+On their schedule, the agent can post the bytes using the presigned URL. However, both the agent and the service never sign a receipt stating that this content is now stored. This is where `store/deliver` can help.
+
+While not relevant for verifiability, it is also important pointing out that the service MAY need to run some data computation, issue billing events or write indexes for retrievability. Today, web3.storage operated store service relies on "S3 Bucket events" to achieve this, but it should not be assumed that all stores web3.storage will rely on will provide such features.
+
+## store/delivery capability
```
alanshaw, line 52:
```suggestion
## `store/deliver` capability
```

rfc/store-deliver.md
```diff
@@ -0,0 +1,278 @@
...
+
+Aiming to introduce verifiability into content storing, as well as to pave the way for a future proof of delivery, `store/deliver` capability is proposed. It would also make services not tied with store put events to execute side effects.
+
+To offer verifiability, `store/add` invocation should be associated with `store/deliver` via [effect]s. `store/deliver` would be invoked by the client to notify the service that content was stored, and the the service should also sign a confirmation that it is true.
+
+### store/delivery of not previously stored content
```
alanshaw, line 58:
```suggestion
### `store/deliver` of not previously stored content
```

rfc/store-deliver.md
```diff
@@ -0,0 +1,278 @@
...
+  "can": "store/deliver",
+  "with": "did:key:abc...",
+  "nb": {
+    "link": "bag...",
+    "url": "https://..."
+  }
+}
+```
+
+5. service verifies if content was delivered by user, and queues a self invocation of `store/deliver` that is also provided as an effect.
```
alanshaw, line 110: I'd like this to be less complicated - in new invocation spec effects are CIDs to _tasks_ not _invocations_ so the server could issue this next task in the receipt and the client can invoke to complete the task and then we're done.

alanshaw (changes requested):

rfc/store-deliver.md
```diff
@@ -0,0 +1,310 @@
...
+    "ok": {
+      "status": "upload",
+      "url": "https://...",
+      "link": "bag...",
+      "allocated": 1234
+    }
+  },
+  "fx": {
+    "join": { "/": "bafy...storeConfirm" },
+    "fork": [{ "/": "bafy...storeDeliver" }]
```
vasco-santos, line 122: note that we won't be able to implement this until we implement invocation spec, so implementation would miss the fork for the start

rfc/store-deliver.md
```diff
@@ -0,0 +1,310 @@
...
+
+On their schedule, the agent can post the bytes using the presigned URL. However, both the agent and the service never sign a receipt stating that this content is now stored. This is where `store/deliver` and `store/confirm` can help. In short, we can see `store/deliver` as a way for the agent to sign that they did the work, and `store/confirm` for the service to sign that the bytes were received, stored and can now be retrieved.
+
+While not relevant for verifiability, it is also important pointing out that the service MAY need to run some data computation, issue billing events or write indexes for retrievability asynchronously. Today, web3.storage operated store service relies on "S3 Bucket events" to achieve this, but it should not be assumed that all stores web3.storage will rely on will provide such features.
+
+## New capabilities
+
+### `store/deliver` capability
+
+After the agent handled the asynchronous task of posting the bytes, the agent can invoke `store/deliver` to notify the service that the bytes were delivered. This invocation in the future MAY include a challenge for the service based on the posted data.
```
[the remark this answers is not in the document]
Gozala: Point I'm trying to make is that invocation spec has [promise pipelines](https://github.com/ucan-wg/invocation?tab=readme-ov-file#await) as a means for forming dependencies between tasks. Ideally we would embrace those so task coordination is captured as opposed to having them described in just a spec.

rfc/store-deliver.md
```diff
@@ -0,0 +1,310 @@
...
+  "nb": {
+    "link": "bag...",
+    "url": "https://..."
+  }
+}
+```
+
+### `store/confirm` capability
+
+Wwhen the service handles the invocation of `store/deliver`, and potentially, after some asynchronous computations/verifications, the service invokes `store/confirm` in order to issue a signed receipt that the flow is now finished.
```
Gozala, line 71:
```suggestion
When the service handles the invocation of `store/deliver`, and potentially, after some asynchronous computations/verifications, the service invokes `store/confirm` in order to issue a signed receipt that the flow is now finished.
```

rfc/store-deliver.md
```diff
@@ -0,0 +1,310 @@
...
+  "can": "store/deliver",
+  "with": "did:key:abc...",
+  "nb": {
+    "link": "bag...",
+    "url": "https://..."
+  }
+}
+```
+
+### `store/confirm` capability
```
Gozala, line 69: We really should have this 

1. In a different name space
2. Make resource service DID not user did.

Because it is service that can invoke this not the user and user has ownership of their space key not us.

rfc/store-deliver.md
```diff
@@ -0,0 +1,310 @@
...
+}
+```
+
+## Common flows
+
+Aiming to introduce verifiability into content storing, as well as to pave the way for a future proof of delivery, `store/deliver` capability is proposed. It would also make services not tied with store put events to execute side effects.
+
+To offer verifiability, `store/add` invocation should be associated with `store/deliver` and `store/confirm` via [effect]s. `store/deliver` would be invoked by the client to notify the service that content was stored, and the the service should also sign a confirmation that it is true via `store/confirm`.
+
+### `store/deliver` of not previously delivered content
```
Gozala, line 90: It may be a good idea to put this into a different namespace so that `with` is our DID not user DID, and `store/add` could delegate capability to invoke it. That would also enable use to perform the task e.g. on some time schedule etc...

General point is space DID here is not a resource, it's our bucket that is a resource.

rfc/store-deliver.md
```diff
@@ -0,0 +1,310 @@
...
+
+4. client invokes `store/deliver`
+
+```json
+{
+  "can": "store/deliver",
+  "with": "did:key:abc...",
+  "nb": {
+    "link": "bag...",
+    "url": "https://..."
```
Gozala, line 137: What is the URL here, pre-signed PUT url we handed prior ?

rfc/store-deliver.md
Gozala: I agree with this, that said there is still needs to be a pending task awaiting for `store/deliver` receipt that will issue receipt from us that workflow was complete. I tried to show this in the flowchart in https://github.com/web3-storage/RFC/pull/10/files#r1514851900

rfc/store-deliver.md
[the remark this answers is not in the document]
vasco-santos: @Gozala I definitely see this as a promising alternative. There is one critical point though that seems a hard requirement that this won't cover. We would like to have the `blob/close` + `assert/location claim`  to happen right away. While they do not happen the data is not available. This is a requirement that we have continuously not able to accomplish, and we always had reports of delays to get data ready to read (today in w3up is due to delay of replicator to CF).

The problem is:

> When service is made aware of this it can then perform blob/close or something along those lines which is more or less what store/confirm is. Here is the rough workflow diagram

By being the client to immediately tell that seems a more efficient way than service having a huge list of tasks to handle in a queue. Also ends up being less prone to errors and maintenance on failures.

We could potentially consider some event subscription and associated runs with the receipt reporting API. Even though this will also require us to thing on how we want to manage receipts being sent by client.

rfc/store-deliver.md
vasco-santos: yeah, I think resource here should probably be a did of the write target?

rfc/store-deliver.md
vasco-santos: yeah, I think it makes sense to be the same, rather than user deriving something. What do you think?

rfc/store-deliver.md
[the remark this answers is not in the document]
Gozala: > By being the client to immediately tell that seems a more efficient way than service having a huge list of tasks to handle in a queue. Also ends up being less prone to errors and maintenance on failures, and obviously way more expensive to run.

I think we need to do both (because we have no guarantee client will ever notify us).

1. Have run scheduled on pre-signed url expiry in case we have not got any update from the clients.
2. Have a handler on the receipt that will perform the closure.

> We could potentially consider some event subscription and associated runs with the receipt reporting API. Even though this will also require us to thing on how we want to manage receipts being sent by client.

ucanto agent messages are send both ways and they [can hold both invocations and receipts](
https://github.com/web3-storage/ucanto/blob/c24a99c3078b4c82f3133c5c1bcda6aa63667c84/packages/interface/src/lib.ts#L800-L821) which is to say that client can send receipt to a server after they have completed an upload.

You are correct right now our service router will not react to received receipts in any way, however we could [hook server up to do things with `incoming.report` which is currently ignored](https://github.com/web3-storage/w3infra/blob/main/upload-api/functions/ucan-invocation-router.js#L276-L277)

I think trickiest bit here would be to track dependencies that is `blob/close` would have to execute when we get receipt for `blob/write`. However we can even pun on that and simply add metadata in the receipt that will signal us to poll `blob/write`. Which is not perfect by any means but IMO better than to have no trace of the workflow capture in the receipts.

Alternatively we could just have a endpoint or even a capability to poll a task where handler will attempt to check dependencies and if they are available then run the task.

Both approaches would be general to all of the cases where we have workflows with non-linear dependency graphs as oppose to being on off workarounds for not having support for this yet.

> I would like to get closer to that, but would like to figure out how to not compromise better service for users and service cost

That is fare! I think there is a possible compromise here where we can do a just a bit more, not much and get workflows with traceable receipts for it. If we really want to aggressively cut scope we could even device something like

```js
{
   can: "system/poll",
   with: "did:key:zAlice",
   nb: {
      task: { '/': 'bafy...pending' },
      receipts: {
         'bafy...task': { '/'; 'bafy..receipt' }
      }
   }
}
```

Where handler will effectively just try to run the task we already created in the and put in the receipt and provide all the dependencies for it.

rfc/store-deliver.md
Gozala: Are we using that URL for anything ? Otherwise I would just leave it out.

vasco-santos: My main concern is a transition period where we change bucket and there were a few on going ops, where then this invocation is received after the switch. This would mean `store/add` handler returned presigned url for bucket A, then we swap to bucket B our back end, then `store/deliver` comes in and we do not know where to claim things. I would therefore like to keep this, unless there are other suggestions

rfc/store-deliver.md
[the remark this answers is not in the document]
vasco-santos: while we decided to go to simplest solution now, I think there is a lot of things here that indeed could be the future. We should save this info in an issue, maybe within the context of the `blob/*` proposal you put together?

For instance, the Job on expiration seems also something we should do, but I would like to not increase the scope here of dataanywhere, which by itself is already quite big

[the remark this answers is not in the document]
vasco-santos: @Gozala I think we can frame this differently to better be aligned toward implementing what you suggest later on, once we can. If we frame this into the context of the proof of delivery, where client will send a challenge that service needs to fulfil, this interaction as is needs to happen anyway. We can leverage it for the time being for submitting the index, until we have ease to implement this once UCAN updates land and we bubble them up

rfc/storage-deliver.md
```diff
@@ -0,0 +1,310 @@
+# extend store/* protocol with storege/deliver and storage/confirm
+
+## Authors
+
+- [Vasco Santos], [Protocol Labs]
+
+## Abstract
+
+`store/*` protocol is missing verifiability for content actually uploaded to the provided presigned URL. This factor together with intention to create a proof of delivery makes critical the introduction of a new capabilities in the `storage/*` namespace. In this RFC, we propose `storage/deliver` and `storage/confirm` capabilities.
```
vasco-santos, line 9: better names are welcome :)

hannahhoward (approved):
Just one question -- otherwise LGTM

PR #11, opened 2024-03-04
rfc: organizing work

alanshaw:
[Preview](https://github.com/web3-storage/RFC/blob/b52827a1cc0a06d88992c98420b3d9346f58d62e/rfc/organizing-work.md)

### DIFF
--- a/rfc/organizing-work.md
+++ b/rfc/organizing-work.md
@@ -1,0 +1,58 @@
+# Organizing work
+
+![status:draft](https://img.shields.io/badge/status-draft-yellow.svg?style=flat-square)
+
+## Editors
+
+- [Alan Shaw](https://github.com/alanshaw), [Protocol Labs](https://protocol.ai/)
+
+## Authors
+
+- [Alan Shaw](https://github.com/alanshaw), [Protocol Labs](https://protocol.ai/)
+
+## Abstract
+
+Formalize a process for working that ensures a consistent and up to date view of the work items in our pipeline. Serves to document a process that can be onboarded and observed by others.
+
+## Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC2119](https://datatracker.ietf.org/doc/html/rfc2119).
+
+## Out of scope
+
+* The tools or locations used to document work items. i.e. Github issues, Jira, Notion etc.
+* Roadmapping/general high level direction setting.
+
+## Pipeline
+
+Generally a work item would move from inception to completion through a pipeline in the following order:
+
+1. **Heap** - anyone can pile stuff onto the heap. It's just a place to track things that we may want to work on in the future. It can be of arbitrary size and detail.
+
+    Items that are of ambiguous _certainty_ SHOULD be accompanied by an [RFC](https://github.com/web3-storage/rfc) - a PR that allows others to comment in an open and structured way on the item. For example, an RFC may be a new idea or a way of achieving a goal that could be done in multiple ways. An RFC allows others to validate the idea, question or assumption to determine with more certainty if it should be worked on.
+    
+    All members SHOULD read RFC documents and provide feedback to allow the author to determine validity of the work item and adjust or clean up if necessary.
+
+2. **Stack** - prioritized work items. _Unlike_ a conventional stack, items are added at a position that denotes their relative priority, however, as per a conventional stack, items are usually "popped" from the top.
+
+    Items SHOULD be moved from the heap to the stack as part of a regular lightweight planning exercise that SHOULD include members of the group that will be working on the items. At this point a short description that details the _product benefits_ and _desired end state/outcome(s)_ of the system MUST be added if not already present.
+
+    Items on the stack SHOULD be prioritized according to the general roadmap as well as other factors, such as, but limited to, how on fire things are. Items in the stack may be re-prioritized during group planning.
+    
+    Multiple items in the heap MAY result in a _single_ item on the stack.
+
+    Larger items whose outputs are an implementation of some kind MUST also result in a PR to the [specs](https://github.com/web3-storage/specs/) repo.
+
+3. **Phase** - a short period of time where items on the stack are worked on. It focuses group effort on a particular theme and is analogous to a "sprint" in Scrum.
+
+    A phase MUST be created by group members who will be working on the items. Members decide on a theme and select a set of items from the top of the stack to work on. At this time a size estimate SHOULD be calculated per item to determine whether enough time exists to achieve the task within the phase period.
+    
+    In the case where a stack item is bigger in size than a phase period, members MAY select a set of sub-tasks that can be completed in time.
+    
+    Items MUST be assigned a DRI (Directly Responsible Individual) _and_ a support member. The support member is primarily responsible for reviewing PRs, but MAY assist/pair in other ways. Support members may the DRI for other tasks and vice versa. Larger items MAY involve more than two group members.
+
+    Phases SHOULD NOT last for more than one month. It is RECOMMENDED for phases to be _two weeks_ long.
+
+    Every phase SHOULD have a name, e.g. Crusty Koala.
+
+    It is not always possible for _all_ desired items in a phase to fit under the same umbrella/theme. It is RECOMMENDED in this case for one member of the team to be assigned to these tasks and for that member be immune from this assignment in the subsequent phase.
--- a/rfc/organizing-work.md
+++ b/rfc/organizing-work.md
@@ -31,7 +31,7 @@ Generally a work item would move from inception to completion through a pipeline
 
     Items that are of ambiguous _certainty_ SHOULD be accompanied by an [RFC](https://github.com/web3-storage/rfc) - a PR that allows others to comment in an open and structured way on the item. For example, an RFC may be a new idea or a way of achieving a goal that could be done in multiple ways. An RFC allows others to validate the idea, question or assumption to determine with more certainty if it should be worked on.
     
-    All members SHOULD read RFC documents and provide feedback to allow the author to determine validity of the work item and adjust or clean up if necessary.
+    All members MAY read RFC documents and provide feedback to allow the author to determine validity of the work item and adjust or clean up if necessary.
 
 2. **Stack** - prioritized work items. _Unlike_ a conventional stack, items are added at a position that denotes their relative priority, however, as per a conventional stack, items are usually "popped" from the top.
 


Gozala (approved):
Thanks for putting this together

gammazero (approved):
I am happy with this. It is very similar to other takes on agile development. If it works well then we have clear process, if not we can modify easily as what we are starting with is clear.

rfc/organizing-work.md
```diff
@@ -0,0 +1,58 @@
...
+## Out of scope
+
+* The tools or locations used to document work items. i.e. Github issues, Jira, Notion etc.
+* Roadmapping/general high level direction setting.
+
+## Pipeline
+
+Generally a work item would move from inception to completion through a pipeline in the following order:
+
+1. **Heap** - anyone can pile stuff onto the heap. It's just a place to track things that we may want to work on in the future. It can be of arbitrary size and detail.
```
travis, line 30: nerd. 🤓 ❤️ 🤣

rfc/organizing-work.md
```diff
@@ -0,0 +1,58 @@
...
+
+    Items that are of ambiguous _certainty_ SHOULD be accompanied by an [RFC](https://github.com/web3-storage/rfc) - a PR that allows others to comment in an open and structured way on the item. For example, an RFC may be a new idea or a way of achieving a goal that could be done in multiple ways. An RFC allows others to validate the idea, question or assumption to determine with more certainty if it should be worked on.
+    
+    All members SHOULD read RFC documents and provide feedback to allow the author to determine validity of the work item and adjust or clean up if necessary.
+
+2. **Stack** - prioritized work items. _Unlike_ a conventional stack, items are added at a position that denotes their relative priority, however, as per a conventional stack, items are usually "popped" from the top.
```
travis, line 36: https://en.wikipedia.org/wiki/Priority_queue

rfc/organizing-work.md
```diff
@@ -0,0 +1,58 @@
...
+
+    A phase MUST be created by group members who will be working on the items. Members decide on a theme and select a set of items from the top of the stack to work on. At this time a size estimate SHOULD be calculated per item to determine whether enough time exists to achieve the task within the phase period.
+    
+    In the case where a stack item is bigger in size than a phase period, members MAY select a set of sub-tasks that can be completed in time.
+    
+    Items MUST be assigned a DRI (Directly Responsible Individual) _and_ a support member. The support member is primarily responsible for reviewing PRs, but MAY assist/pair in other ways. Support members may the DRI for other tasks and vice versa. Larger items MAY involve more than two group members.
+
+    Phases SHOULD NOT last for more than one month. It is RECOMMENDED for phases to be _two weeks_ long.
+
+    Every phase SHOULD have a name, e.g. Crusty Koala.
```
travis, line 56: YES

rfc/organizing-work.md
```diff
@@ -0,0 +1,58 @@
...
+
+    Items on the stack SHOULD be prioritized according to the general roadmap as well as other factors, such as, but limited to, how on fire things are. Items in the stack may be re-prioritized during group planning.
+    
+    Multiple items in the heap MAY result in a _single_ item on the stack.
+
+    Larger items whose outputs are an implementation of some kind MUST also result in a PR to the [specs](https://github.com/web3-storage/specs/) repo.
+
+3. **Phase** - a short period of time where items on the stack are worked on. It focuses group effort on a particular theme and is analogous to a "sprint" in Scrum.
+
+    A phase MUST be created by group members who will be working on the items. Members decide on a theme and select a set of items from the top of the stack to work on. At this time a size estimate SHOULD be calculated per item to determine whether enough time exists to achieve the task within the phase period.
```
travis, line 48: Do you imagine multiple Phases in progress at the same time across the organization? It feels like that will inevitably be needed as the team grows and work streams diverge more, but maybe we're not there yet? Do you have any thoughts on how many people would be involved in a single phase or is that out of scope for this doc?

rfc/organizing-work.md
travis: this might be out of scope here, but it feels like it might be worth acknowledging that we will sometimes pull from further down the stack if, eg, the person pulling doesn't want to/isn't able to work on the thing on top of the stack - I'd love to encourage a culture of people picking up things outside their comfort zone, but sometimes that's just not the right move.

this may be obvious enough that it doesn't need to be memorialized here, just running my brain around it a bit!

travis (approved):
I love this!

rfc/organizing-work.md
reidlw: Agree @travis, and none of this should be considered super rigid. Guidelines, not laws.

rfc/organizing-work.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Pipeline
+
+Generally a work item would move from inception to completion through a pipeline in the following order:
+
+1. **Heap** - anyone can pile stuff onto the heap. It's just a place to track things that we may want to work on in the future. It can be of arbitrary size and detail.
+
+    Items that are of ambiguous _certainty_ SHOULD be accompanied by an [RFC](https://github.com/web3-storage/rfc) - a PR that allows others to comment in an open and structured way on the item. For example, an RFC may be a new idea or a way of achieving a goal that could be done in multiple ways. An RFC allows others to validate the idea, question or assumption to determine with more certainty if it should be worked on.
+    
+    All members SHOULD read RFC documents and provide feedback to allow the author to determine validity of the work item and adjust or clean up if necessary.
```
vasco-santos, line 34:
```suggestion
    All members MAY read RFC documents and provide feedback to allow the author to determine validity of the work item and adjust or clean up if necessary.
```

I would frame more as everyone can participate and have time to do so. But also can opt out from participation

rfc/organizing-work.md
```diff
@@ -0,0 +1,58 @@
...
+
+2. **Stack** - prioritized work items. _Unlike_ a conventional stack, items are added at a position that denotes their relative priority, however, as per a conventional stack, items are usually "popped" from the top.
+
+    Items SHOULD be moved from the heap to the stack as part of a regular lightweight planning exercise that SHOULD include members of the group that will be working on the items. At this point a short description that details the _product benefits_ and _desired end state/outcome(s)_ of the system MUST be added if not already present.
```
vasco-santos, line 38: I would love to work in a more metrics driven approach when possible. May be included in product benefits, but I would explicitly include metrics to take into account for the prioritization, like number of customers interested, their volume, etc

rfc/organizing-work.md
vasco-santos: I would also prefer to work in thematic way as much as possible. I think it can be mostly achieved if we have smaller phases, at least for a team of our size.

However, at the bare minimal to accomodate team growth, I think we should commit to:
- have non silos where one single member takes a thematic phase
- not have members spread into more than one thematic phase (unless of course it finishes half way through phase)

rfc/organizing-work.md
```diff
@@ -0,0 +1,58 @@
...
+    
+    In the case where a stack item is bigger in size than a phase period, members MAY select a set of sub-tasks that can be completed in time.
+    
+    Items MUST be assigned a DRI (Directly Responsible Individual) _and_ a support member. The support member is primarily responsible for reviewing PRs, but MAY assist/pair in other ways. Support members may the DRI for other tasks and vice versa. Larger items MAY involve more than two group members.
+
+    Phases SHOULD NOT last for more than one month. It is RECOMMENDED for phases to be _two weeks_ long.
+
+    Every phase SHOULD have a name, e.g. Crusty Koala.
+
+    It is not always possible for _all_ desired items in a phase to fit under the same umbrella/theme. It is RECOMMENDED in this case for one member of the team to be assigned to these tasks and for that member be immune from this assignment in the subsequent phase.
```
vasco-santos, line 58: See my comment above, I think we should avoid at all cost single man show for a set of tasks. If it is not enough work for a phase, we should just have 2 people that would both join the remaining team once done

rfc/organizing-work.md
joaosa: love it

joaosa (approved):
thank you for doing this 🙏

rfc/organizing-work.md
Gozala: If we going to stick to terminology borrowed from computer science, this should be called threads 😅

Jokes aside I do agree with @vasco-santos single person per theme / thread creates silos echo chambers and demotivation. I would also argue that goal should be more about a checkpoint and reevaluation than it is to drain an allocated stack. In our prior retrospectives we keep pushing timelines to drain the stack, it would be more rewarding to just call it re-evaluate and start another phase / thread.

rfc/organizing-work.md
Gozala: I agree with @vasco-santos participation is voluntary unless you're input was explicitly requested.

rfc/organizing-work.md
Gozala: I agree with @vasco-santos on this as well. If things don't fit the phase lets either cut it or give prioritize such that there is a group working on it.

It's not just that solo endeavors are unpleasant, it also creates knowledge gaps where whole teams end up depending on a single person with know how.

rfc/organizing-work.md
hannahhoward, line 30: Honestly, can we just use the regular names:

Heap -> Inbox
Stack -> Backlog

Prefer terms that a new dev coming in may already know

rfc/organizing-work.md
hannahhoward: If we just call it a backlog, we can treat it like a backlog, rather than try to figure out the right data structure that isn't actually an exact match for thing thing we want :P

rfc/organizing-work.md
hannahhoward, line 38: "Desired state / outcomes" - I like "Acceptance criteria" -- it's an easy way to define ahead of time how we measure whether the thing that got built does what we said it would.

rfc/organizing-work.md
```diff
@@ -0,0 +1,58 @@
...
+
+2. **Stack** - prioritized work items. _Unlike_ a conventional stack, items are added at a position that denotes their relative priority, however, as per a conventional stack, items are usually "popped" from the top.
+
+    Items SHOULD be moved from the heap to the stack as part of a regular lightweight planning exercise that SHOULD include members of the group that will be working on the items. At this point a short description that details the _product benefits_ and _desired end state/outcome(s)_ of the system MUST be added if not already present.
+
+    Items on the stack SHOULD be prioritized according to the general roadmap as well as other factors, such as, but limited to, how on fire things are. Items in the stack may be re-prioritized during group planning.
+    
+    Multiple items in the heap MAY result in a _single_ item on the stack.
+
+    Larger items whose outputs are an implementation of some kind MUST also result in a PR to the [specs](https://github.com/web3-storage/specs/) repo.
```
hannahhoward, line 44: I find this "implementation of some kind" a bit ambiguous. Do you mean a protocol we haven't implemented yet? Every ticket is an implementation of SOME kind right?

rfc/organizing-work.md
```diff
@@ -0,0 +1,58 @@
...
+
+3. **Phase** - a short period of time where items on the stack are worked on. It focuses group effort on a particular theme and is analogous to a "sprint" in Scrum.
```
hannahhoward, line 46: This is the one I have the most trouble with. I think we're trying to combine two seperate ideas into a single one.

The first is just a cycle for getting stuff done -- i.e. a sprint, though really I just mean the combination of backlog grooming, sprint planning, and reflection. A "rhythm" to work if you will. I think at least till we are larger, there should just be one of these.

The second is grouping work -- how we insure people can take on big themes together and not get silo'd alone or stuck doing a single thing. I really like the idea of temporary sub team. Essentially, we're all on web3.storage but we can form groupings that work closely for a period of time on a big deliverable -- that can span a month, or even, in some cases longer.

My anxiety about trying to combine both is that we end up making "what's happening" extremely difficult, and impose artificial time contraints on work that may take much longer than a sprint to do.

Am I missing something here in the explanation? Am I not getting what a phase is supposed to do?

hannahhoward (changes requested):
See comment about phases. I want to understand more. I put request changes cause I need more clarity, though perhaps once I have that no changes will be needed.

rfc/organizing-work.md
reidlw: I think this is a really important thing to tease apart.

rfc/organizing-work.md
reidlw: +1. The renaming doesn't appear to add anything, and so by default is probably to be avoided.

rfc/organizing-work.md
reidlw: And what Hannah says. 😅

rfc/organizing-work.md
reidlw: Can I just x100 this? :) Yes - couldn't agree more @vasco-santos. As we start populating the backlog with stuff, I'd like to push for a simple template that calls for articulating both the projected impact of the work, and the cost of doing it.

PR #12, opened 2024-03-05
feat: hierarchical capability organisation

Gozala:


Gozala:
@vasco-santos asking your feedback here as I think it might be worth introducing new capabilities already following proposed conventions

rfc/hierarchical-capabilities.md
```diff
@@ -0,0 +1,57 @@
...
+
+In addition we should capture depndency across namespaces in the way capabilities are orginalizated e.g. if upload task requires storing bytes we nest `store` capabilities under `upload` here is an example layout that may work
+
+
+```sh
+# write permission
+/space/content/add                          # upload/add
+/space/content/add/shard                    # store/add
+/space/content/add/shard/allocate           # store/allocate
+/space/content/add/shard/write              # S3/R2 PUT
```
vasco-santos, line 23: I am not sure about this proposal in specific. I would not consider `upload/add` to be a "parent" of `store/add` in a way. Probably I would put it into a different layer, like `space/collection` and drop `shard` from here instead?

vasco-santos (approved):
I like a lot this proposal and I am in for it. I would like to see in the RFC a more clear direction path for the migration. I see you mean we should keep both, but my main idea was: "we are probably doing a big breaking change when UCAN 1.0 and invocation spec, that seems a great timing", even though I know some motivation was to ease client refactor.

Maybe a combo of the above could be an option, so flag as deprecate and remove on the timing I mention.

vasco-santos (commented):
(I did not mean to approve, but comment at this point :/)

rfc/hierarchical-capabilities.md
```diff
@@ -0,0 +1,57 @@
...
+/space/content/remove/shard                 # store/remove
+
+# write+delete permission
+/space/content/edit/add                     # upload/add
+/space/content/edit/add/shard               # store/add
+/space/content/edit/add/shard/allocate      # store/allocate
+/space/content/edit/remove                  # upload/remove
+/space/content/edit/remove/shard            # store/remove
+
+/space/shard/add                            # store/add
```
hannahhoward, line 40: why would I use these two methods only, as opposed to the ones underneath `space/content`

asking cause I'm tempted to say just drop `content` entirely and make `/space/add`

I do wonder whether doing this kind of lower level hierarchy organization (as opposed to the /space & /accounts, which make 100% sense) is making assumptions about how folks want to delegate capabilities.

hannahhoward: like someone wants to just offer `/space/contend/add` (i.e.  `upload/add`) without delegating `/space/content/add/shard` -- maybe there is a usecase we just haven't thought of?

rfc/hierarchical-capabilities.md
```diff
@@ -0,0 +1,57 @@
+# Hierarchical Capabilities
+
+## Abstract
+
+Currently UCAN capabilities described in [W3Up specification](https://github.com/web3-storage/specs) do not follow consistent organizational structure. To make things worse accounts on login delegate `*` capabilities and spaces delegate set of namespaces.
+
+This creates a serious challenge for the programs (like w3up client) that attempt to infer set of methods to expose based no delegated capabilities.
+
+We roughly two roles in the sysntem `space` and `account` and their capability set do not overlap, so we can somewhat infer which methods to present in case of `*` ability by subject DID, but that hinges upon the fact that we use `did:key` for space and `did:mailto` for accounts. This assumbtion may be rendered obsolete in the future.
```
hannahhoward, line 9: If I understand, we're infering methods based on the did type of the subject? I 100% agree let's not do this.

hannahhoward (approved):
I am convinced on the top level hierarchy, not as clearly convinced on the lower level hiearchy, though I guess I don't strongly object.

Consider me non-blocking on this though I encourage others to discuss what I raised if it feels relevant to you.

rfc/hierarchical-capabilities.md
Gozala: Yeah the last on was the thinking here, in fact that is more or less our setup. That said I think what we arrived at https://github.com/web3-storage/RFC/pull/12#discussion_r1525222040 might be a better option, because `content/add` gives you blanket ability and you could have sub namespaces for shards and roots.

rfc/hierarchical-capabilities.md
vasco-santos: Yes, something more along the lines of `/space/content/add/root`, `/space/content/add/pack` I think would feel better

Gozala: > I also wonder if this is the correct order - I was thinking perhaps the verb should be at the end? Should the noun(s) drill down to the "action" (verb) from `/`? So you can read the capability right to left and it make sense?

While I do share the sentiment problem is that when user wants to share ability to add content they would have to delegate several capabilities as you need be able to add shards and uploads etc...

Which is also why `content/add` contained all kinds of writes so user can just delegate white capability, or limit it to specific write capability where needed.

Gozala: > It feels more likely to me that users would want to "wildcard-delegate" based on the thing (noun) not the action (verb)....probably not explaining that well, but I'd be hesitant to delegate /space/add/* - ability to add anything that exists or will exist in the future (narrow action, wide things), whereas /space/blob/* - any action with blobs specifically feels safer, despite widening of granted actions (narrow thing, wide actions).

Well that is why I had `/space/content/` to cover everything content related wheres `/space/content/add` covered all content writes and if you want to get very specific that `/space/content/add/blob/`.

So if we replace `blob` with `content` you get basically what you're suggesting here. However I wanted `content` to cover blobs (which are effectively shards) and DAGs consisting of blocks from shards.

Perhaps `content` is not best term for it ?

rfc/hierarchical-capabilities.md
```diff
@@ -0,0 +1,57 @@
...
+/space/content/list/shard                   # store/list
+
+# delete permission
+/space/content/remove                       # upload/remove
+/space/content/remove/shard                 # store/remove
+
+# write+delete permission
+/space/content/edit/add                     # upload/add
+/space/content/edit/add/shard               # store/add
+/space/content/edit/add/shard/allocate      # store/allocate
```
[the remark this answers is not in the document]
Gozala: Can you please elaborate what exactly does not ? Are you suggesting that no one would want to give `write + delete` or that calling it `edit` is misleading or something else ?

Gozala:
Thinking more about this I think it would be a good idea to try model what user space as KV looks like and then define commands accordingly. I suspect that our commands will very likely start to resemble KV operations hence it would make senes to do that to inform our design.

rfc/hierarchical-capabilities.md
vasco-santos: I still think content/blob would be important because we want to have caps to add side indexes and so one to the space? While they coul in theory be a content themselves, I think making distinction would be good so that we can treat those more as hidden files in console

rfc/hierarchical-capabilities.md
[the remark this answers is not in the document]
alanshaw: The fact that there are 3 different verbs in 1 capability - edit, add and allocate - I do not understand what being delegated this capability would allow me to do.

PR #13, opened 2024-03-07
 docs: datawherehouse location claim + store/publish

vasco-santos:
[HTML View](https://github.com/web3-storage/RFC/blob/docs/datawherehouse/rfc/datawherehouse.md)

This RFC proposes `datawherehouse` to keep track of location of bytes stored with write targets as part of data anywhere. In short, this is a datastore to store locations that also requires a flow such as the proposed on https://github.com/web3-storage/RFC/pull/10


rfc/datawherehouse.md
```diff
@@ -0,0 +1,132 @@
...
+
+## Background
+
+TL;DR
+
+1. We're missing a mapping of Data CID -> target URI (Bucket, Saturn Node, etc).
+2. We don't want clients to create location claims for internal bucket URLs because we might change the location in the future (reputation hit for client).
+3. We don't have bucket events in Cloudflare, so need the client to tell us when it has uploaded something to the provided write target.
+4. We want freeway code to be usable in Saturn nodes, so ideally it uses only content claims to discover locations.
+5. We want this information available as soon as content is written so that read interfaces can serve the content right away.
```
Gozala, line 17: Putting this comment here as I'm not sure what better place to put this would be. This is followup from the call this morning. I think we should issue a location claim when we verify content was delivered. Challenge however was how do we capture information like things like `bucket`, `node` etc.. not present in the URL so we could perform compute closer to a data.

I think there are few possible solution here:

1. We encode all that information as hints e.g. `https://w3s.link/car/bag...car?r2&bucket=foo`
    - 💚 Anyone will be able to perfrom read from that gateway URL
    - 💚 Gateway could take advantage of those hints or not care about them.
    - 💚 Other things in the system can recognize those hints and perform direct from bucket read as opposed to just going to that URL (backdoor / fastlane access)

2. We could publish claim with gateway URL and publish bucket and other info as a separate (non-location) claim
    - 💚 Anyone will be able to perform read from that gateway URL
    - 💚 Internal system can query both claims and to device most effecient access

3. We could publish claims with our custom URL schemas e.g. r2://bucket/.... 
    - ❤️‍🩹 Anyone who knows how to resolve it can read it, we can even publish resolver library
       - Our gateway could even support `https://w3s.link/r2://bucket/.../cid` URL
    - 💚 Internal system can query both claims and to device most efficient access

Broadly speaking I'm advocating for capturing whatever extra information we need to enable more optimal reads for internal use, yet make URLs publicly readable from the outside perhaps in less optimal mannery.

vasco-santos: Thanks for the write up. I now understand what you mean. The "Gateway could take advantage of those hints or not care about them" doing whatever other discovery they feel like/know about makes sense to me.

Yes, I think this a good suggestion

rfc/datawherehouse.md
```diff
@@ -0,0 +1,198 @@
...
+
+A location claim MAY look like:
+
+```json
+{
+  "op": "assert/location",
+  "rsc": "https://web3.storage",
+  "input": {
+    "content" : CID /* CAR CID */, 
+    "location": "`https:<CID>.dag.w3s.link?bucket=<bucketName>&region=<region>`",
```
alanshaw, line 108:
```suggestion
    "content" : { "/": "bafy...BLOBCID" }, // typically a CAR
    "location": "`https://w3s.link/ipfs/bafy...BLOBCID?origin=r2://region/bucketName/key",
```

alanshaw (approved):
Approve for "location claims with encoded params"!

rfc/datawherehouse.md
```diff
@@ -0,0 +1,198 @@
...
+
+> 2. We don't want clients to create location claims for internal bucket URLs because we might change the location in the future (reputation hit for client).
+
+Extending on the first point, making location claims to include "private/unavailable" URIs will make it harder for the service to move data to other places, given it would need to revoke a bunch of claims and re-write them with new location.
```
Gozala, line 27: Not pushing for this, just want use to get on the same page in regards to overall content claims design. Content claims are not designed as pub / sub, they are designed as point to point. That is user making a claim for us in no means imply that we have to republish as is to outside network. That is to say it is not unreasonable for user to make a that is true for us and no one else. When we verify the claim we can choose to re-publish it as-is but using our DID in the `with` field or make another claim derived from it e.g. move data to some public space and issue claim for it.

It is also worth calling out that in UCAN 1.0 we don't even create proof chains so capturing relation to original claim will be strictly optional and an on us.

rfc/datawherehouse.md
```diff
@@ -0,0 +1,198 @@
...
+
+> 3. We don't have bucket events in Cloudflare, so need the client to tell us when it has uploaded something to the provided write target.
+
+Actually we can even extend on this point by saying that today we have no verifiability on data being sent by the user, as well as received by the service. By having the client to sign that bytes were sent, and the service to check and also sign that is true will allow us to achieve that. Moreover, we also open the doors in this interaction for a challengen/proof of delivery.
+
+## High level flow of proposed solution
+
+* Simulate bucket event by having client submit `store/deliver` invocation.
```
Gozala, line 35: This is bikeshed at this point, but I would much rather reframe it as `block/publish` or something implying that user wants uploaded content to be discoverable / readable under that CID.

rfc/datawherehouse.md
```diff
@@ -0,0 +1,198 @@
...
+    * Effect linked by `store/add` receipt.
+    * Can be batched with other invocations like `filecoin/offer`
+* Handler writes to location claims with encoded information
+
+The following diagram presents the described flow, and is illustrated with the following steps:
+
+1. Client requests to store some bytes with the storefront service
+2. Service issues a receipt stating the URI of a write target for the client to write those bytes
+3. Client writes the bytes into the write target
+4. Client notifies the service that the bytes were written to the provided write target
```
Gozala, line 46: It would make more sense to me to reframe it as client tells service to serve written bytes under given CID. Server verifies that it has bytes corresponding to that CID and responds with location claim from which read can be performed.

Gozala: In fact it might be a good idea to also allow user to specify `TTL` for the produced read URL and maybe even consider future where this read URI might require permission (e.g. ucan authorization).

Not implying to do this now, just suggesting to reframe problem in a way that would allow these kinds of things

rfc/datawherehouse.md
```diff
@@ -0,0 +1,198 @@
...
+2. Service discovers where the requested bytes are stored relying on content claims service and the materialized claims from `datawherehouse`
+3. Service serves data stored on the discovered location.
+
+![datawherehouse2](./datawherehouse/datawherehouse-2.svg)
+
+## Location claims
+
+Content claims service is currently deployed implementing the [content claims spec](https://github.com/web3-storage/specs/pull/86). Among other claims, it provides [Location Claims](https://hackmd.io/IiKMDqoaSM61TjybSxwHog?view#Location-Claims) which MAY be used to claim that the bytes that hash to a given CID are available in a given URL.
+
+In w3s the service is responsible to deciding the write target, therefore service SHOULD be responsible for claiming the location of the bytes.
```
Gozala, line 64: Reframing this as user requesting read URL for their data seems to align well here.

rfc/datawherehouse.md
```diff
@@ -0,0 +1,198 @@
...
+
+While thinking about using location claims to record where bytes are stored by the service, there are a few characteristics we want to have:
+- location claim MUST resolve to a public and fetchable URLs
+- location in location claim SHOULD (ideally) not change recurrently given it MAY impact negatively the reputation of a party
```
Gozala, line 68: I really think we should let user choose how long the location claim should be valid when they request content to be published. If they're chosen TTL is to long we can refuse, if it is too short we can also refuse. But broadly speaking it should be a content owners call and not ours.

Again not suggesting to implement all this, but framing it through this terms seem like a good idea

rfc/datawherehouse.md
```diff
@@ -0,0 +1,198 @@
...
+
+Having a claim that cannot be used by every potential client (i.e. needs some extra permissions) is also a disadvantage that may represent penalties in a reputation system. Moreover, rate limits can have a negative impact on reputation as well.
+
+URIs that include all necessary information to enable derivation for smart clients that could rely on Worker R2 Bindings or to generate Presigned URLs are critical for several use cases. URIs that minimize Egress costs can also be preferred by smart clients.
+
+Per the above, considering S3, it looks like we should rely on S3 Object URL (e.g. `https://<BUCKET_NAME>.s3.<AWS_REGION>.amazonaws.com/<CID>/<CID>.car`).
+
+But with R2, there is no perfect fit. The only good option would be the format used by presigned URLs, but it should only be created on request given their expiration. The custom domain offers better retrievability than the Public Object URL for Dev, but has not enough information encoded for smart clients (i.e. no way to know bucket name). For R2, we will likely need to add 2 location URIs:
```
Gozala, line 175: I think this also fits nicely with the idea of user "request read URL" for the content as we can ensure that user requested TTL is aligned with pre-signed R2 URL TTL.

rfc/datawherehouse.md
```diff
@@ -0,0 +1,198 @@
...
+- Custom domain object URL (e.g. `https://<CUSTOM_DOMAIN>.web3.storage/<CID>/<CID>.car`)
+  - can be used out of the box to fetch content
+- Presigned URL like URI without query params (e.g. `https://<ACCOUNT_ID>.r2.cloudflarestorage.com/<BUCKET_NAME>/<CID>/<CID>`)
+  - won't really work, but smart clients can see if it is available and rely on its encoded info to use CF Worker Bindings / create presigned URLs, etc.
+
+Alternatively, we can require `<CUSTOM_DOMAIN>` to become bucket name, in order to make it work as a single location URI. Main disadvantage besides the extra requirement, is that there is no url mention of being a R2 bucket, which would mean hard coded assumptions. We could also consider to mimic S3 URL for R2 here as well.
+
+The Public Object URL for Dev should not be adopted, as it is heavily rate limited, we do not know what CF may do with it in the future, and also does not even have information about the Bucket name.
```
Gozala, line 183: I think it is safe to assume that no matter what schema we may arrive to given known constraints is not going to hold with constraints that may be introduced in the future. Which is why I would suggest dumb stateless bridge similar to UCAN bridge that simply decodes URL and redirects to an appropriate location. This would allow us to put same logic into client and save us from hop to a bridge. It is also something we could upgrade over time as our constraints change.

rfc/datawherehouse.md
```diff
@@ -0,0 +1,198 @@
...
+
+#### Proposal
+
+Nothing prevents us from claiming multiple location URIs for a given content, however we may need to also be careful on having multiple claims for the same location as if it is not fully available it MAY be ranked badly in whatever reputation system we may create. However, some smart clients MAY benefit of cost savings or faster retrievals if they have extra information encoded in the URI.
```
Gozala, line 187: I am confused now, I thought service was issuing content claims and not the client, but here you seem to go other way round.

Gozala (changes requested):
I provided some feedback inline. I did not approve PR only because reading this document I'm left with an impression that two conflicting proposals are made. Thinking more on this and digesting all the shared insights I think I'm coming to a following conclusion:

1. Client could issue receipt that they uploaded content (IN THE FUTURE), but that should not trigger pipeline, meaning it should not make that content available on IPFS or put it into filecoin.
    - I think this would also align with potential private data requirement that has been brought up by product.
2. Client should explicitly request read URL for the content, which is effectively reframing of `store/deliver`. When user requests content identified by CID to be readable we can verify that we have it and either issue location claim or produce an error saying we don't have it.
    - Furthermore we probably should consider incorporating TTL in the user request because forever is impossible in finite bounds of universe. In fact accepting TTL is a commitment and we probably could / should reflect this in billing.
    - This also may set us up on a path of supporting request URLs that are not public, which I think it's matter of time. I imagine such URLs could be under space DID and require UCAN token to perform reads. Not something I propose we tackle now, but it is good to consider how things would fit together.
3. When client requests content read URL we should respond with a location claim from us (audience should be DID of the `with` of the request). That means client can publish claim on our behalf wherever they wish, because it is commitment from us.
4. I think we really should design it with `RAW` CIDs as opposed to CAR CIDs.  Because we do not verify those are CARs. I'm not suggesting rewriting everything here, but I would encourage to do it for the new capabilities as it is easy to go from CAR <->  RAW CID it's just switching a codec byte. Alternatively we could use multihash instead.

rfc/datawherehouse.md
vasco-santos: this is deprecated part by now. But in this context. smart clients mean clients reading location claims such as frreeway/roundabout

rfc/datawherehouse.md
```diff
@@ -0,0 +1,198 @@
...
+5. We want this information available as soon as content is written so that read interfaces can serve the content right away.
+
+### Extended details
+
+> 1. We're missing a mapping of Data CID -> target URI (Bucket, Saturn Node, etc).
+
+We considered at first location claims directly to data bytes would be the solution here. But when we got closer to put that into practise we realized this was not a good idea. We need this mapping so that w3s Read/Write Interfaces can discover where the bytes are. Where the bytes are stored may actually be private write targets (for instance, a R2 bucket), which location is not public. We consider that location claims MUST be retrievable, have public access and not heavily rate limited. Finally, some read interfaces (for instance Roundabout, Freeway) require some information encoded in the URI (like bucket name), which would not be available in a public URL of R2 bucket. All things considered, Location claims should include URIs like `https://bafy...data.ipfs.w3s.link`, and we need a mapping of `bafy...data` to where its bytes are actually stored internally.
```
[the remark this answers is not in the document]
vasco-santos: I mean we considered location claim to `r2://...`

rfc/datawherehouse.md
prodalex: 100%.

prodalex:
> Client could issue receipt that they uploaded content (IN THE FUTURE), but that should not trigger pipeline, meaning it should not make that content available on IPFS or put it into filecoin.
I think this would also align with potential private data requirement that has been brought up by product.

Yeah. The client will need to be able to decide eventually if content should be public, private, kept in hot storage (maybe even number of replicas/store targets but to be considered in billing) or actually just be archived in filecoin and then deleted from hot storage once the CommP and pieceID is available. 

We should also think about Object lifecycle management. The customer might want to set up rules when content is automatically deleted from hot storage after X time or X time without reads, etc. and only kept in filecoin.

rfc/datawherehouse.md
vasco-santos: we decided on `store/publish`

rfc/datawherehouse.md
```diff
@@ -0,0 +1,236 @@
...
+    "ok": {
+      "link" : { "/": "bafy...BLOBCID" }, // RAW CID - go from CAR <-> RAW CID it's just switching a codec byte
+      "location": "`https://w3s.link/ipfs/bafy...BLOBCID?origin=r2://region/bucketName/key"
+    }
+  },
+  "fx": {},
+}
+```
+
+TODO: Should return the actual location claim? How would we do that?
```
vasco-santos, line 93: TODO: would love your help here @Gozala ?

vasco-santos: or maybe a CID for it? but client can't really read one today? Maybe receipt CID that claim was made?

rfc/datawherehouse.md
```diff
@@ -0,0 +1,236 @@
...
+
+This capability can be specified as follows:
+
+```json
+{
+  "op": "store/publish",
+  "rsc": "did:key:abc...space",
+  "input": {
+    "link": { "/": "bafy...BLOBCID" }, // RAW CID - go from CAR <-> RAW CID it's just switching a codec byte
+    "url": "https://..." // write target presignedurl previously provided
```
vasco-santos, line 73: maybe I should include some info for this. My reasoning to put this is to let service know what actually was the write target so that service can check it is there. Otherwise, we would need to find out by storing presigned URLs somewhere. Thoughts?

rfc/datawherehouse.md
```diff
@@ -0,0 +1,236 @@
...
+      "link" : { "/": "bafy...BLOBCID" }
+    }
+  },
+  "fx": {},
+}
+```
+
+## Location claims encoding location hints
+
+Content claims service is currently deployed implementing the [content claims spec](https://github.com/web3-storage/specs/pull/86). Among other claims, it provides [Location Claims](https://hackmd.io/IiKMDqoaSM61TjybSxwHog?view#Location-Claims) which MAY be used to claim that the bytes that hash to a given CID are available in a given URL.
```
Gozala, line 112: I think fact that size of the content / blob is not captured (or is optional) is an unfortunate oversight. I would suggest changing `range` field into required that way size of the content is clear from the claim.

rfc/datawherehouse.md
```diff
@@ -0,0 +1,236 @@
...
+```
+
+Return on success the following receipt:
+
+```json
+{
+  "ran": "bafy...storePublish",
+  "out": {
+    "ok": {
+      "link" : { "/": "bafy...BLOBCID" }, // RAW CID - go from CAR <-> RAW CID it's just switching a codec byte
```
Gozala, line 85: In the new capabilities I'd suggest using `blob` or `content` as filed name instead of `link`.

rfc/datawherehouse.md
```diff
@@ -0,0 +1,236 @@
...
+
+## Location claims encoding location hints
+
+Content claims service is currently deployed implementing the [content claims spec](https://github.com/web3-storage/specs/pull/86). Among other claims, it provides [Location Claims](https://hackmd.io/IiKMDqoaSM61TjybSxwHog?view#Location-Claims) which MAY be used to claim that the bytes that hash to a given CID are available in a given URL.
+
+In w3s the service is responsible to deciding the write target, therefore service SHOULD be responsible for claiming the location of the blob on user request to have it published.
+
+While thinking about using location claims to record where bytes are stored by the service, there are a few characteristics we want to have:
+- location claim MUST resolve to a public and fetchable URLs
+- location in location claim SHOULD (ideally) not change recurrently given it MAY impact negatively the reputation of a party. However, we should consider letting client choose how long the location claim should be valid for.
```
Gozala, line 118: I would say it should not change within the commitment window of the claim.

rfc/datawherehouse.md
```diff
@@ -0,0 +1,236 @@
...
+
+As a way to store the location of this bytes, we discussed relying on a "private" location claims concept, or even on location claims for a gateway that have hints as encoded params in the URL that the read interface can decide if want to try to use. This would allow us to already have the infra and datastores we have, leaving the decentralization of content claims for a completely different problem.
+
+### _private_ location claims
+
+_private_ location claims would enable us to not expose these claims directly to the user, given their sole purpose at the moment is internal routing. This would enable queries of w3s read/write interfaces to know where the bytes for a CID are stored.
+
+With this building block we can issue claims that MAY not be public and fetchable URLs, as well as not have worries on a potential future data migration.
+
+A _private_ location claim MAY look like:
```
Gozala, line 130: I'm starting to think that private location claims are obsolete. Content claim should include whatever metadata it needs in the claim to make anyone with it able to perform a read. We can simply leverage UCAN auth system for the rest.

In other words query params ≈ ucan auth header. Later in addition gives us ability to choose who can exercise it while former is public by default.

rfc/datawherehouse.md
```diff
@@ -0,0 +1,236 @@
...
+
+```json
+{
+  "op": "assert/location",
+  "aud": "did:key:abc...space",
+  "rsc": "did:web:private.web3.storage",
+  "input": {
+    "content" : CID /* // RAW CID */, 
+    "location": "`https://<BUCKET_NAME>.<REGION>.web3.storage/<CID>/<CID>.car`",
+    "range"   : [ start, end ] /* Optional: Byte Range in URL
```
Gozala, line 140: We should make this required field so size of the content can be inferred from the claim.

Gozala (approved):
I have proposed various things here, but I think we can call this a consensus and can iterate on remaining details in the spec.

rfc/datawherehouse.md
vasco-santos: embed would mean we need to implement in ucanto what we talked about in Istanbul to attach things via receipt builder, but that seems good

rfc/datawherehouse.md
vasco-santos: > Thinking more about this I feel like this is not a good requirement, because client may have to do request /add in one session and do /publish in the other. Having to maintain state across these invocations seems like incidental complexity and I would advocate against it.

Not sure what you meant. My point of having the URL was precisely what I think you mean here. Not require state on our end across multiple invocations. Note that I am also assuming that we can derive write target details via presigned URL, which I am not entirely sure it will be true.

rfc/datawherehouse.md
```diff
@@ -0,0 +1,236 @@
...
+> 3. We don't have bucket events in Cloudflare, so need the client to tell us when it has uploaded something to the provided write target.
+
+Actually we can even extend on this point by saying that today we have no verifiability on data being sent by the user, as well as received by the service. By having the client to sign that bytes were sent, and the service to check and also sign that is true will allow us to achieve that. Moreover, we also open the doors in this interaction for a challengen/proof of delivery.
+
+## High level flow of proposed solution
+
+* Client stores some blob with web3.storage
+* Client requests this blob to be published with `store/publish`
+* Service handler verifies requested blob was stored on the client space, issues a location claim with encoded information for location and returns it to the client
+  * Can be batched with other invocations like `filecoin/offer`
```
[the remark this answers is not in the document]
vasco-santos: not important, but was more towards the idea that we should batch all follow up invocations of `store/add` that can happen in parallel

rfc/datawherehouse.md
vasco-santos: well, for this case we are meaning blobs (or CAR files), which in theory is the entire thing, but yeah, since we need to do a HEAD request to see if it is in the write target we can make this required as it is really no extra cost

rfc/datawherehouse.md
vasco-santos: since we are all in agreement to not use this, I will just drop it from the RFC. FWIW, this is currently there as "alternative", but later is specified what this RFC proposes

rfc/datawherehouse.md
```diff
@@ -0,0 +1,236 @@
...
+Extending on the first point, making location claims to include "private/unavailable" URIs will make it harder for the service to move blob to other places, given it would need to revoke a bunch of claims and re-write them with new location.
+
+> 3. We don't have bucket events in Cloudflare, so need the client to tell us when it has uploaded something to the provided write target.
+
+Actually we can even extend on this point by saying that today we have no verifiability on data being sent by the user, as well as received by the service. By having the client to sign that bytes were sent, and the service to check and also sign that is true will allow us to achieve that. Moreover, we also open the doors in this interaction for a challengen/proof of delivery.
+
+## High level flow of proposed solution
+
+* Client stores some blob with web3.storage
+* Client requests this blob to be published with `store/publish`
```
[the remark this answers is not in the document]
vasco-santos: `/space/content/add/blob/publish` yeah, I think that I agree with this direction + start with blob work for new path

rfc/datawherehouse.md
Gozala: You are thinking about state we (service) needs to maintain and I was referring to the state client will need to maintain. Requiring client to be stateful or a dependency on them providing right url is not good choice.

Point I was making is that we (service) needs to lookup state in DB anyway so extending a (state) record with a URL so client will be stateless was my recommendation. That also reduces surface for errors where client fails to provide a URL or a correct URL for that matter.

vasco-santos: 👌🏼

PR #14, opened 2024-03-15
rfc: github org migration

joaosa:
[Preview](https://github.com/web3-storage/RFC/blob/rfc/github-org-migration/rfc/github-org-migration.md)

The tl;dr is we're leaving everything as is and moving only what we're sure to be working on in the next 6mo. 
Migration is a reversible process, so we can adjust accordingly. The concrete list of repos to migrate is:

- [w3up](https://github.com/web3-storage/w3up);
- [w3cli](https://github.com/web3-storage/w3cli);
- [console](https://github.com/web3-storage/console);
- [w3infra](https://github.com/web3-storage/w3infra);
- [w3filecoin-infra](https://github.com/web3-storage/w3filecoin-infra);
- [www](https://github.com/web3-storage/www);
- [content-claims](https://github.com/web3-storage/content-claims);
- [RFC](https://github.com/web3-storage/RFC);
- [specs](https://github.com/web3-storage/specs).

Reasoning: we know for sure we'll need to work on these repos in the next 6 months in the new venture. We should evolve (not fork) them for the new venture, as they are considered a valuable foundation on which to build, and we should bring existing customers without big breaking changes or migrations;

New repos should be created in the w3s-project org;

Other important and in-use support repos (e.g. libraries) move to w3s-project org;

Other repos may never migrate, or may migrate as and when needed.


rfc/github-org-migration.md
```diff
@@ -0,0 +1,154 @@
...
+- [ ] [add-to-web3](https://github.com/web3-storage/add-to-web3)
+- [ ] [files-from-path](https://github.com/web3-storage/files-from-path)
+- [ ] [multipart-parser](https://github.com/web3-storage/multipart-parser)
+- [ ] [example-image-gallery](https://github.com/web3-storage/example-image-gallery)
+- [ ] [broker](https://github.com/web3-storage/broker)
+- [ ] [web3-schema](https://github.com/web3-storage/web3-schema)
+- [ ] [example-forum-dapp](https://github.com/web3-storage/example-forum-dapp)
+- [ ] [fauna-dump](https://github.com/web3-storage/fauna-dump)
+- [ ] [db-migration-pipeline](https://github.com/web3-storage/db-migration-pipeline)
+- [ ] [fauna-dumpify](https://github.com/web3-storage/fauna-dumpify)
```
vasco-santos, line 72: these can really be archived and dropped from migration as we used them for old DB migrations of old products

rfc/github-org-migration.md
```diff
@@ -0,0 +1,154 @@
...
+- [ ] [admin.storage](https://github.com/web3-storage/admin.storage)
+- [ ] [hydra-booster](https://github.com/web3-storage/hydra-booster)
+- [ ] [ipns-publisher](https://github.com/web3-storage/ipns-publisher)
+- [ ] [specs](https://github.com/web3-storage/specs)
+- [ ] [consul-cluster-go-ipfs](https://github.com/web3-storage/consul-cluster-go-ipfs)
+- [ ] [repin-unpinned-cids](https://github.com/web3-storage/repin-unpinned-cids)
+- [ ] [parse-link-header](https://github.com/web3-storage/parse-link-header)
+- [ ] [js-replica](https://github.com/web3-storage/js-replica)
+- [ ] [w3up-archived](https://github.com/web3-storage/w3up-archived)
+- [ ] [upload-protocol](https://github.com/web3-storage/upload-protocol)
```
vasco-santos, line 82: probably also should be archived/ignored

vasco-santos (commented):
note that we also have some things in nftstorage org that should also be migrated https://github.com/nftstorage

joaosa:
> note that we also have some things in nftstorage org that should also be migrated https://github.com/nftstorage

thanks for the heads-up! can you add me to the org, so I can list all repos? 🙏

vasco-santos:
@joaosa I also noticed a few repos are missing, noticed at least `w3infra` and `w3filecoin-infra`. Reason was I was also going to callout that we need to check what is the impact of moving these repos under a new org with their deployments with seed.run. Probably we should setup a mini project with sst + seed.run and migrate the repo to see what happens to the infra

joaosa:
> @joaosa I also noticed a few repos are missing, noticed at least `w3infra` and `w3filecoin-infra`. Reason was I was also going to callout that we need to check what is the impact of moving these repos under a new org with their deployments with seed.run. Probably we should setup a mini project with sst + seed.run and migrate the repo to see what happens to the infra

Well spotted! All repos should be accounted for now. Also got the ones from `nft.storage` 👍 and added your recommendations.

rfc/github-org-migration.md
```diff
@@ -0,0 +1,415 @@
...
+
+Ensure a common understanding on how repositories should be migrated on a case-by-case basis.
+
+Provides recommendations and fixes to make this migration process as smooth and bulletproof for everyone.
+
+## Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC2119](https://datatracker.ietf.org/doc/html/rfc2119).
+
+Migration is defined as the act of transferring ownership from either the [Filecoin Saturn](https://github.com/filecoin-saturn) or [Web3 Storage](https://github.com/web3-storage) orgs into [w3s](https://github.com/w3s-project) and ensuring all dependent systems continue operating transparently. If such a goal is not possible, then the necessary changes to either prevent the system from stopping its operation or recover it will be proposed.
```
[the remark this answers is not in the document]
gobengo: For context, one of the hardest things for me when grokking the IPFS/PL ecosystem is that seemingly the majority of times you click a documentation link to a github repo, it redirects you to some repo that is archived because the repo got passed around between a couple teams or something. If people had just left the public repos alone, those links would still link to repos with coherent docs and repo state, just maybe not committed to in awhile, but without the frustration of 'oh well i was eager to use this but now that it's explicitly abandoned I'm going to reevaluate things'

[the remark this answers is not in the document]
joaosa: @gobengo you make a very good point which I had not considered yet. I have experienced that confusion myself, so let's avoid it when we can. thank you!

In the case of public repos, I suppose forking and marking the "old" version as archived while linking to the new repo, for reference and maintenance reasons, should do it. As you said, there are no cost savings to be had here and this makes navigation clearer for end users. Any new changes will land in the fork.

For private repos (there are few sensitive ones on Saturn's side at least), I would still transfer them, as repo dependencies are solely defined internally and this would reduce the amount of artifacts we leave behind.
Also wrt to cost savings, given that "unless you are on a free plan, adding an outside collaborator to a private repository will use one of your paid licenses." ([ref](https://docs.github.com/en/organizations/managing-user-access-to-your-organizations-repositories/managing-outside-collaborators/adding-outside-collaborators-to-repositories-in-your-organization)), it makes sense to consolidate them into a single final org, so as to minimize the number of paid seats across the four orgs we will inherit in terms of maintenance (ideally, only one of them will need to evolve over time).

what do you think?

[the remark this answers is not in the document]
gobengo: > In the case of public-facing repos, I suppose forking and marking the "old" version as archived while linking to the new repo, for reference and maintenance reasons, should do it.

For https://github.com/web3-storage/w3up
If we can get to the point where the web3-storage org is on a free github plan, is there any reason at all there needs to be a w3s-project/w3up repo?

IMO we don't even need to have a new repo for w3up in a new org, so we wouldn't even need to do the 'marking the "old" version as archived' in that case. (I could be wrong, I just don't understand why w3up inherently needs to be migrated to a new github org when it could just stay put).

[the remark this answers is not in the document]
joaosa: > If we can get to the point where the web3-storage org is on a free github plan, is there any reason at all there needs to be a w3s-project/w3up repo?

Given that, as I see it, the primary goal here is enable cost savings (I would say a secondary goal is consolidation while minimizing complexity), it all comes down to the set of Github features we need to operate efficiently. 

The web3-storage org is using Github Enterprise and if we can do away with all the features listed [here](https://docs.github.com/en/get-started/learning-about-github/githubs-plans#github-team) and with no private repos, then it could be downgraded to a free org. That would enable us to leave things as they are.
I can't really tell what the nft.storage org is using as I'm not an org owner, but Saturn is using Github Teams.

Maybe we should have this discussion first, so as to figure out where to head next.
If we need any of those features then we'll need to move things around, as we'll need Github Teams and we only wanted to billed for that in a single org. 

I don't think we need Github Enterprise for web3-storage, but I'll do my best to verify that.
I'll also try to analyze and come up with a more concrete rationale on which features we need based on usage, but for now here's the list of Github Teams features:

- [ ] GitHub Support via email
  - This will likely need to exist in the org(s) we actually use. Have you all ever used this feature?
- [ ] 3,000 GitHub Actions minutes per month
  - Will need to verify this a bit deeper wrt to usage
- [ ] 2 GB GitHub Packages storage
  - same
- [ ] Advanced tools and insights in private repositories:
  - likely not need
- [ ] Required pull request reviewers
  - needs to be verified
- [ ] Multiple pull request reviewers
  - this is needed
- [ ] Draft pull requests
  - useful but maybe not needed
- [ ] Team pull request reviewers
  - needs to be verified
- [ ] Protected branches
  - this is needed
- [ ] Code owners
  - needs to be verified
- [ ] Scheduled reminders
  - needs to be verified
- [ ] GitHub Pages (private)
  - not needed
- [ ] Wikis
  - needs to be verified
- [ ] Repository insights graphs: Pulse, contributors, traffic, commits, code frequency, network, and forks
  - nice to have?
- [ ] The option to enable or disable GitHub Codespaces
  - needs to be verified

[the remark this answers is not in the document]
joaosa: Another argument in favor of a migration is the fact we can't add issues across orgs to project tracking. If we were to keep things as they are, we'd need to have a layer of project-specific issues on project tracking referring to the concrete ones in the final repos.

rfc/github-org-migration.md
```diff
@@ -0,0 +1,425 @@
...
+- [ ] [js-replica](https://github.com/web3-storage/js-replica)
+- [ ] [w3up-archived](https://github.com/web3-storage/w3up-archived)
+- [ ] [upload-protocol](https://github.com/web3-storage/upload-protocol)
+    - It SHOULD be archived.
+    - It SHOULD NOT be migrated.
+- [ ] [ucanto](https://github.com/web3-storage/ucanto)
+- [ ] [w3up](https://github.com/web3-storage/w3up)
+- [ ] [ucanto-name-system](https://github.com/web3-storage/ucanto-name-system)
+- [ ] [alt-gateway](https://github.com/web3-storage/alt-gateway)
+- [ ] [link.web3.storage](https://github.com/web3-storage/link.web3.storage)
```
alanshaw, line 104: @vasco-santos do you know what this is?

rfc/github-org-migration.md
```diff
@@ -0,0 +1,425 @@
...
+- [ ] [dag-check](https://github.com/web3-storage/dag-check)
```
alanshaw, line 105: Archive/ignore

rfc/github-org-migration.md
```diff
@@ -0,0 +1,425 @@
...
+- [ ] [pickup](https://github.com/web3-storage/pickup)
+- [ ] [workers](https://github.com/web3-storage/workers)
```
alanshaw, line 107: I don't _think_ we use this. Maybe @vasco-santos knows?

rfc/github-org-migration.md
```diff
@@ -0,0 +1,425 @@
...
+- [ ] [minibus](https://github.com/web3-storage/minibus)
+- [ ] [w3name](https://github.com/web3-storage/w3name)
+- [ ] [reads](https://github.com/web3-storage/reads)
+- [ ] [claudio](https://github.com/web3-storage/claudio)
+- [ ] [minibus-client](https://github.com/web3-storage/minibus-client)
+- [ ] [w3link](https://github.com/web3-storage/w3link)
+- [ ] [dagula](https://github.com/web3-storage/dagula)
+- [ ] [dagula-gateway](https://github.com/web3-storage/dagula-gateway)
+- [ ] [fast-unixfs-exporter](https://github.com/web3-storage/fast-unixfs-exporter)
```
alanshaw, line 116: Archive/ignore

rfc/github-org-migration.md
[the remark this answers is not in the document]
gobengo: @joaosa I don't know the backstory on why we use the more expensive GitHub Enterprise. We do make heavy-ish use of github actions.

>  That would be a decent win, but that would not solve project tracking 

One thing I'm sort of sad to lose that we have in web3-storage org but not w3s-project org is https://docs.github.com/en/issues/managing-your-tasks-with-tasklists/about-tasklists

rfc/github-org-migration.md
```diff
@@ -0,0 +1,425 @@
...
+- [ ] [parse-link-header](https://github.com/web3-storage/parse-link-header)
+- [ ] [js-replica](https://github.com/web3-storage/js-replica)
+- [ ] [w3up-archived](https://github.com/web3-storage/w3up-archived)
+- [ ] [upload-protocol](https://github.com/web3-storage/upload-protocol)
+    - It SHOULD be archived.
+    - It SHOULD NOT be migrated.
+- [ ] [ucanto](https://github.com/web3-storage/ucanto)
+- [ ] [w3up](https://github.com/web3-storage/w3up)
+- [ ] [ucanto-name-system](https://github.com/web3-storage/ucanto-name-system)
+- [ ] [alt-gateway](https://github.com/web3-storage/alt-gateway)
```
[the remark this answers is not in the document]
joaosa: does this mean it can be archived or should it be kept around / migrated as is?

rfc/github-org-migration.md
vasco-santos: it was some tests we did in our team week in Miami to test the service worker deployment. I think we can archive

rfc/github-org-migration.md
vasco-santos: We use this for the workers for things like loki/logs

vasco-santos: see https://github.com/web3-storage/reads/blob/main/packages/edge-gateway/package.json#L21

rfc/github-org-migration.md
```diff
@@ -0,0 +1,755 @@
...
+- [ ] [claudio](https://github.com/web3-storage/claudio)
+    - Original MVP for hoverboard.
+    - It SHOULD be archived.
+    - It SHOULD NOT be migrated.
+- [ ] [minibus-client](https://github.com/web3-storage/minibus-client)
+    - The cli for minibus.
+    - It SHOULD NOT be archived.
+    - It SHOULD NOT be migrated.
+- [ ] [w3link](https://github.com/web3-storage/w3link)
+    - [ ] check with @vascosantos10 if it can be archived
```
vasco-santos, line 205: no this has the deployed worker for w3s.link

rfc/github-org-migration.md
```diff
@@ -0,0 +1,755 @@
...
+    - It SHOULD NOT be archived.
+    - It SHOULD NOT be migrated.
+- [ ] [dagula-gateway](https://github.com/web3-storage/dagula-gateway)
+    - Experiment to back a GW with Elastic IPFS over Bitswap, but it didn't work out with the amount of traffic.
+    - It SHOULD be archived.
+    - It SHOULD NOT be migrated.
+- [x] [fast-unixfs-exporter](https://github.com/web3-storage/fast-unixfs-exporter)
+    - already archived.
+- [ ] [telemetry-grafana-agent](https://github.com/web3-storage/telemetry-grafana-agent)
+    - [ ] check with @vascosantos10 if it can be archived
```
vasco-santos, line 217: no, this holds a bunch of our Grafana metrics, including w3up, gateways, cloudflare metrics, etc

vasco-santos: note this is deployed in a digital ocean app

rfc/github-org-migration.md
```diff
@@ -0,0 +1,755 @@
...
+    - Blog for the old website.
+    - It SHOULD NOT be archived.
+    - It SHOULD NOT be migrated.
+- [ ] [w3ui-website](https://github.com/web3-storage/w3ui-website)
+    - It SHOULD NOT be archived.
+    - It SHOULD NOT be migrated.
+- [x] [attach-write-to-read](https://github.com/web3-storage/attach-write-to-read)
+    - already archived.
+- [ ] [sigv4](https://github.com/web3-storage/sigv4)
+    - [ ] check with @vascosantos10 if it can be archived
```
vasco-santos, line 266: no, it is used by w3up to generate presigned urls

rfc/github-org-migration.md
```diff
@@ -0,0 +1,755 @@
...
+    - It SHOULD NOT be archived.
+    - It SHOULD NOT be migrated.
+- [ ] [diffusionbee-stable-diffusion-ui](https://github.com/web3-storage/diffusionbee-stable-diffusion-ui)
+    - It SHOULD be archived.
+    - It SHOULD NOT be migrated.
+- [ ] [ai-artwork-uploader](https://github.com/web3-storage/ai-artwork-uploader)
+    - It should be left behind as a demo.
+    - It SHOULD NOT be archived.
+    - It SHOULD NOT be migrated.
+- [ ] [sst-monorepo](https://github.com/web3-storage/sst-monorepo)
```
vasco-santos, line 291: yes

rfc/github-org-migration.md
```diff
@@ -0,0 +1,755 @@
...
+    - It SHOULD NOT be archived.
+    - It SHOULD NOT be migrated.
+- [ ] [backup-progress](https://github.com/web3-storage/backup-progress)
+    - It SHOULD be archived.
+    - It SHOULD NOT be migrated.
+- [ ] [secrets](https://github.com/web3-storage/secrets)
+    - It SHOULD NOT be archived.
+    - It SHOULD NOT be migrated.
+- [ ] [carpark-backfill](https://github.com/web3-storage/carpark-backfill)
+    - [ ] check with @vascosantos10 if it can be archived.
```
vasco-santos, line 359: yes

rfc/github-org-migration.md
```diff
@@ -0,0 +1,755 @@
...
+- [ ] [ipni](https://github.com/web3-storage/ipni)
+    - To talk to IPNI.
+    - It SHOULD NOT be archived.
+    - It SHOULD NOT be migrated.
+- [x] [dealer](https://github.com/web3-storage/dealer)
+    - already archived.
+- [ ] [carpark-bucket-diff](https://github.com/web3-storage/carpark-bucket-diff)
+    - [ ] check with @vascosantos10 if it can be deleted.
```
vasco-santos, line 367: yes

rfc/github-org-migration.md
```diff
@@ -0,0 +1,755 @@
...
+- [x] [docs](https://github.com/web3-storage/docs)
+    - already archived.
+- [ ] [go-w3up](https://github.com/web3-storage/go-w3up)
+    - It SHOULD NOT be archived.
+    - It SHOULD NOT be migrated.
+- [ ] [migrate-to-w3up](https://github.com/web3-storage/migrate-to-w3up)
+    - It SHOULD NOT be archived.
+    - It SHOULD NOT be migrated.
+- [ ] [piece-compute-worker](https://github.com/web3-storage/piece-compute-worker)
+    - [ ] Ask @vascosantos10 if it can be deleted or archived.
```
vasco-santos, line 461: not really, we may still want a worker to compute Filecoin Pieces

rfc/github-org-migration.md
```diff
@@ -0,0 +1,773 @@
...
+- [ ] [ucanto](https://github.com/web3-storage/ucanto)
+    - UCAN encoding implication for creating UCANs and sending them to a service and verifying delegations.
+    - It SHOULD NOT be archived.
+    - It SHOULD be migrated.
+- [ ] [w3up](https://github.com/web3-storage/w3up)
+    - Implementation of the protocols and specs repos. The current web3.storage system.
+    - It SHOULD NOT be archived.
+    - It SHOULD be migrated.
+- [ ] [ucanto-name-system](https://github.com/web3-storage/ucanto-name-system)
+    - [ ] check with @gobengo what should be done
```
gobengo, line 174: I don't know of anything that relies on or depends on this. It would be nice if that URL stays resolving. I think that means a archive or migration to other org is safe.

rfc/github-org-migration.md
```diff
@@ -0,0 +1,773 @@
...
+    - Cli for devs.
+    - It SHOULD NOT be archived.
+    - It SHOULD NOT be migrated.
+- [ ] [console](https://github.com/web3-storage/console)
+    - Web UI for w3up.
+    - It SHOULD NOT be archived.
+    - It SHOULD NOT be migrated.
+- [ ] [one-webcrypto](https://github.com/web3-storage/one-webcrypto)
+    - Fork
+    - [ ] Ask @gobengo if it can be deleted or archived.
```
gobengo, line 460: `@web3-storage/access` (part of w3up-client) depends on this https://github.com/web3-storage/w3up/blob/6fa7797ba6f5c0bbd359c807e257882a7d5d6fb8/packages/access-client/package.json#L113

joaosa: thank you 🙏 . great to know

rfc/github-org-migration.md
joaosa: Archiving is probably the safest, so I'll suggest that

rfc/github-org-migration.md
```diff
@@ -0,0 +1,770 @@
...
+    - It SHOULD NOT be archived.
+    - It SHOULD NOT be migrated.
+- [ ] [pieces](https://github.com/web3-storage/pieces)
+    - Cli for devs.
+    - It SHOULD NOT be archived.
+    - It SHOULD NOT be migrated.
+- [ ] [console](https://github.com/web3-storage/console)
+    - Web UI for w3up.
+    - It SHOULD NOT be archived.
+    - It SHOULD NOT be migrated.
```
alanshaw, line 455:
```suggestion
    - It SHOULD be migrated.
```

rfc/github-org-migration.md
```diff
@@ -0,0 +1,770 @@
...
+- [ ] [one-webcrypto](https://github.com/web3-storage/one-webcrypto)
+    - Fork which w3up depends upon. See [here](https://github.com/web3-storage/w3up/blob/6fa7797ba6f5c0bbd359c807e257882a7d5d6fb8/packages/access-client/package.json#L113).
+- [ ] [www](https://github.com/web3-storage/www)
+    - Current website.
+    - It SHOULD NOT be archived.
+    - It SHOULD NOT be migrated.
```
alanshaw, line 461:
```suggestion
    - It SHOULD be migrated.
```

alanshaw (approved):
LGTM!

rfc/github-org-migration.md
```diff
@@ -0,0 +1,770 @@
...
+> (...)
+> "All links to the previous repository location are automatically redirected to the new location. When you use git clone, git fetch, or git push on a transferred repository, these commands will redirect to the new repository location or URL. However, to avoid confusion, we strongly recommend updating any existing local clones to point to the new repository URL."
+
+This ensures migration disruptions are minimal and that we can keep operating normally and migrate on a repo-by-repo basis. Still, it is crucial to look at each repo individually to ascertain migration consequences which may impact other systems outside of Github and our local development machines.
+Additionally, it is RECOMMENDED to perform the [post-migration instructions](#post-migration-instructions) below.
+
+CI pipelines were checked to make sure nothing relies on a particular repo naming structure. Github was used and the results were manually checked using the following search commands:
```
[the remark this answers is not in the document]
joaosa: One thing I was worried about was that if a particular component relied on another one through its repo name that something could go amiss. Either way, this would result in a Github HTTP redirect.

rfc/github-org-migration.md
```diff
@@ -0,0 +1,770 @@
...
+
+New repositories MUST be created in [w3s](https://github.com/w3s-project).
+
+The goal of this section is to migrate repos we are going to be actively working on, so we're able to downgrade source orgs to the Github free tier.
+Ideally, migrations should happen when we are working on repos whose purpose overlaps between w3s and the previous project.
+
+We have a total of 193 candidate repositories to migrate (out of a total of 203 of which 10 were already archived).
+We SHOULD migrate repos one-by-one with an opportunistic approach after we have established it is safe to do so.
+Once the actions for a repository are successfully identified, it will be marked as [x].
+If a given repository has no checkbox, than means it was sorted prior to this document.
```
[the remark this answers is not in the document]
joaosa: I meant "dealt with" and more specifically archived. I'll rephrase.

rfc/github-org-migration.md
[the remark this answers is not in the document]
joaosa: As for the seed.run side of things, state is tided to the app on seed.run's side and afaict hidden in the depths of CloudFormation. the repo is only regarded as the source of truth to compute the infra state. if a repo is moved, then reading from it will stop working and all attached infra will remain as is. Reading can be fixed by reattaching the new repo located in the new org. Hope this makes sense

Gozala (approved):
Looks good to me. Only think I would suggest to create release for each repo we migrate to validate that migration was successful.

vasco-santos (approved):
Great work! ❤️

PR #15, opened 2024-03-15
rfc: ipni for indexing w3c stored data

gammazero:
[HTML View](https://github.com/web3-storage/RFC/blob/ba472a10ac6aab142f19d85fd012590502fc1dd8/rfc/ipni-w3c.md)

Describes the ipni/offer capability and how data is indexed and retrieved without relying on centralized services.

- This is necessary for the write-anywhere initiative.
- Depends on https://github.com/web3-storage/specs/pull/85
- Derived from spec proposal [w3-ipni](https://github.com/web3-storage/specs/blob/40fb9981225dc8b6180e9e4586f6a31b970508e8/w3-ipni.md)


rfc/ipni-w3c.md
```diff
@@ -0,0 +1,346 @@
...
+
+type EntryChunk struct {
+    Entries [Bytes]
+    Next optional Link
+}
+```
+- `Addresses` the most recent value of this is returned with lookups for any CID in any advertisement. It will contain the multiaddr form of URL claims service? TBD.
+- `ContextID` must be the byte encoded form of the CAR CID.
+- `Entries` must be the CID of an `EntryChunk` for a subset (or all) of the multihashes in the CAR.
+- `Metadata` encodes the location of the content claims bucket and the inclusion claim that ties the CAR CID to the CARx-index CID.
```
gammazero, line 271: I think it would be better to not put any content claim here, but rather just the minimal needed to get the content claims bucket. 

The advertisement is signed so it cannot be altered without invalidating the signature/CID, so if the ad contains the CID of the claims bucket and the claims bucket has an inclusion claim that contains the advertisement CID, then that should provide a chain of provable authenticity.

rfc/ipni-w3c.md
```diff
@@ -0,0 +1,346 @@
...
+  -   The context id bytes provide the CAR CID for any block look up. The CAR CID can then be used to find the CAR index via our content-claims API.
+  - We could delete the IPNI records by CAR CID if the CAR is deleted.
+- Make IPNI advertising an explicit UCAN capability that clients can invoke rather than a side-effect of bucket events
+  - With this we are free to write CARs anywhere. The user's agent invokes a `ipni/offer` capability to ask us to publish an IPNI advertisement for the blocks in their CAR.
+  - This empowers the user to opt-in or out as they need, and allows us to bill for the (small) cost of running that service.
+- Provide a way to retrieve a CAR index, that has (CID, offset, length) tuples, without retrieving the CAR file.
+    - Eliminates the need to rely on the CARPark database. Block location information can now be stored on the w3s network.
+    - Clients can request specific data item and only receive data for that item.
+- Put the source inclusion claim in the IPNI advert metadata.
+  - We have to sign IPNI Adverts as the provider. Providing a signed source claim allows more nuanced reputation decisions.
```
hannahhoward, line 51: Could the source inclusion claim be signed by the client, if IPNI supported UCAN delegations?

rfc/ipni-w3c.md
```diff
@@ -0,0 +1,346 @@
...
+            "ID": "QmUA9D3H7HeCYsirB3KmPSvZh3dNXMZas6Lwgr4fv1HTTp",
+            "Addrs": [
+              "/dns4/dag.w3s.link/tcp/443/https"
+            ]
+          }
+        }
+]}]}
+```
+
+web3.storage publishes the blocks it can provide by encoding a batch of multihashes as an IPLD object and writing it to a bucket as an `Advertisement`, addressed by it's CID.
```
hannahhoward, line 98: I am familiar with a lot of this but it might be helpful to offer a synchronous "brown bag lunch" type learning on how IPNI works, the design goals, etc. In particular, why things like ContextIDs are critical for scale is useful. I'm not 100% sure how much this fully explains what IPNI is intending to do, purely from text.

rfc/ipni-w3c.md
hannahhoward: I agree. As you know I think it might make sense to put a location claim for the car-index in here for parallelism of round trips.

hannahhoward: Truth be told this will all work better though if we can get away from car files one day :P

hannahhoward: But even if we do, I assume we will be stuck supporting car files in perpetuity.

rfc/ipni-w3c.md
```diff
@@ -0,0 +1,346 @@
...
+sequenceDiagram
+    actor Alice
+    Alice->>w3s: put CAR + CARx-index
+    activate Alice
+    Alice-->>Alice: create location claims
+    Alice->>w3s: ipni/offer (inc, loc claims)
+    deactivate Alice
+    activate w3s
+    w3s-->>w3s: fetch & verify CARx-index
+    w3s-->>w3s: create/store claims bucket
```
hannahhoward, line 165: this would also possibly move up if the client isn't making location claims

rfc/ipni-w3c.md
```diff
@@ -0,0 +1,346 @@
...
+```json
+{
+  "content": CID, // CAR CID
+  "includes": CID // CARx-index Index CID
+}
+```
+
+When `ipni/offer` is invoked, encoded inclusion claim and encoded location claims for the CAR and the CARx-index are all sent with the invocation.
+
+_Alternatively, the w3s service could create the location claims. The user would still need to provide the locations, so might as well do this as a location claim. If w3s moves the data, then it can replace the location claims with new ones_
```
hannahhoward, line 226: right where it lives is owned by w3s, so w3s should create the locaiton claims -- but only on user trigger. I don't think the user needs to provide the locations per-se -- W3S is already tracking that.

rfc/ipni-w3c.md
```diff
@@ -0,0 +1,346 @@
...
+
+The service must fetch he CARx-index and parse it to find the set of multihashes included in the CAR. see: [Verifying the CARv2 Index](#verifying-the-carv2-index)
+
+_Alternatively the service could generate the CARx-index and store it, but it would be better to have the clients do that work of generating and storing the index._
```
hannahhoward, line 230: Strong push for clients to do this ONLY.

My basic rubric is:
authority over what the data is is ALWAYS the client
authority over where the data lives is ALWAYS w3s

rfc/ipni-w3c.md
```diff
@@ -0,0 +1,346 @@
...
+
+A user can use a CID that identifies data stored on w3c to lookup the location where the data can be retrieved and content claims describing that data.
+
+The data's location can change and content claims about the data can change, at anytime during the data's storage lifetime, and queried for the data CIDs will be able to get the most locations and claims.
+
+This RFC describes how IPNI indexing can be used to lookup data locations and content claims for any CID within a CAR stored in w3c.
+
+## Abstract
+
+A user stores data as a CAR file. This CAR contains a set of multihashes for individual pieces of data within the CAR.
```
Gozala, line 17: While that is the case today I would highly encourage to reframe this differently. That is users store content as opaque blobs containing series of blocks addressed vie multihashes. It is still true but allows us to treat CARs as currently implementation detail

rfc/ipni-w3c.md
```diff
@@ -0,0 +1,346 @@
...
+
+The data's location can change and content claims about the data can change, at anytime during the data's storage lifetime, and queried for the data CIDs will be able to get the most locations and claims.
+
+This RFC describes how IPNI indexing can be used to lookup data locations and content claims for any CID within a CAR stored in w3c.
+
+## Abstract
+
+A user stores data as a CAR file. This CAR contains a set of multihashes for individual pieces of data within the CAR. 
+
+The user creates a an external CAR index [CARx-index](https://github.com/web3-storage/RFC/pull/9/files) that contains tuples of (CID, offset, length) of any items they intend to make individually retrievable from the car.
```
Gozala, line 19: I think here also it may make sense to have byte range indexes within blob addressed by multihashes as opposed to cids

rfc/ipni-w3c.md
hannahhoward: I should add I think we should absolutely make the entire ipni/offer process, plus generating the CAR-index, etc, entirely optional (front client software perspective, it's on by default to start). This leaves the ability to not index things potentially in the clients hands.

This design for example, enables Bitswap via using IPNI to map to CAR CID for any CID. If people don't want Bitswap, potentially eventually with flat data not CAR-ified, they might not need all this. They might just need to index the root and other CIDs of interest.

rfc/ipni-w3c.md
```diff
@@ -0,0 +1,346 @@
...
+    w3s-->>w3s: fetch CAR
+    w3s-->>w3s: extract data from CAR
+    w3s-->>Alice: data item(s)
+    deactivate w3s
+```
+        
+**Note**: One round trip can be avoided if the CARx-index location is in the IPNI response, instead of having to read it out of a location claim. This may not work if the CARx-index location claim is too big for the metadata (recommended size: 100 bytes, max size: 1024 bytes).
+
+**Question**: 
+Will the claims bucket be located at some arbitrary URL, or is it stored in a location known to w3s? If the latter then only the CID of the cliams bucket or CID of the CAR file is needed to fetch claims.
```
hannahhoward, line 333: Let's assume for the moment that the retriever is a public client, and THEY don't know the location absent a public URL, even if W3S does.

hannahhoward (changes requested):
My main ask off the bat is to align with #13, which says location claims don't get made on the client.

rfc/ipni-w3c.md
hannahhoward: Additionally, this all assumes public retrieval is an intended outcome -- for private retrieval, not advertising at all seems like a given.

rfc/ipni-w3c.md
```diff
@@ -0,0 +1,346 @@
...
+
+## Abstract
+
+A user stores data as a CAR file. This CAR contains a set of multihashes for individual pieces of data within the CAR. 
+
+The user creates a an external CAR index [CARx-index](https://github.com/web3-storage/RFC/pull/9/files) that contains tuples of (CID, offset, length) of any items they intend to make individually retrievable from the car.
+
+With an [inclusion claim](https://hackmd.io/@gozala/content-claims?utm_source=preview-mode&utm_medium=rec#Inclusion-Claims), the user who created the CAR asserts that CAR contains a given set of multihashes, transitively, via the CARx-index.
+
+After storing the CAR and CARx-index, w3s creates a [location claim](https://hackmd.io/@gozala/content-claims?utm_source=preview-mode&utm_medium=rec#Location-Claims) for the CAR file and the CARx-index. Each location claim maps an item's CID to a URL where the item can be retrieved.
```
Gozala, line 23: I think linked inclusion claims were not a very good idea as in they don’t provide enough information to be all that useful and tie us to CARs. I think multihash inclusion claims that follow are better take.

specifically I would suggest adopting variant of inclusion claims where we claim that some that (multi)hash A is a byte range of content addressed by (multi)hash B

This also makes equivalence  claims obsolete because covering full range allows you to say this sha256 is same as this blake3 hash.

I’m still not sure if raw cid is better than multihash however. It seems more flexible but it also becomes vague if you start using e.g unixfs cids

rfc/ipni-w3c.md
```diff
@@ -0,0 +1,346 @@
...
+
+w3s creates a Content Claims Bucket. This is a container, stored on w3s network, that holds all content claims associated with the stored data. The location claims and inclusion claim are written to the claims bucket. The claims bucket is then stored.
```
Gozala, line 25: This is a case but also something we’d like to move away from in favor of storing claims in user spaces instead. That way costs associated with storing and serving claims is legged by users as opposed to be unpredictable expense for a service

rfc/ipni-w3c.md
```diff
@@ -0,0 +1,346 @@
...
+
+[IPNI](https://github.com/ipni/specs/blob/main/IPNI.md) will allow a client to lookup any of the multihashes, that are in the CARx-index, to get information sufficient to recover the claims bucket. The location claims from the claims bucket can then be used to get the CARx-index and CAR data.
+
```
Gozala, line 28: I’m failing to understand what information is advertised exactly. Is that a cid for the index ? Is it a claim itself ? Something else ?

rfc/ipni-w3c.md
```diff
@@ -0,0 +1,346 @@
...
+### Motivation
+
+- Align IPNI advertisement entries with CAR block sets, so that these CIDs can be queried to get the location of the CARx-index and CAR data.
+  - This exposes our block-to-car indexes. Anyone could use IPNI to find which CAR a block is in.
+  -   The context id bytes provide the CAR CID for any block look up. The CAR CID can then be used to find the CAR index via our content-claims API.
+  - We could delete the IPNI records by CAR CID if the CAR is deleted.
+- Make IPNI advertising an explicit UCAN capability that clients can invoke rather than a side-effect of bucket events
+  - With this we are free to write CARs anywhere. The user's agent invokes a `ipni/offer` capability to ask us to publish an IPNI advertisement for the blocks in their CAR.
+  - This empowers the user to opt-in or out as they need, and allows us to bill for the (small) cost of running that service.
```
Gozala, line 46: In the other thread we kind of converging on the idea that when user asks for the content state in their space they will get a location claim issued by service to the space did. Which is effectively our signed commitment to serve that content from that url through a ucan validity timeframe.

User can redelegate that to other user to allow them to read it. I proposed to make advertisement request on IPNI to simply be delegation of location claim to an IPNI did. Similarly authorization to serve content form gateway would be delegation to our gateway did etc… my point is we don’t need new capabilities this way, we simply allow users to choose whom to give read access

Gozala: Plz note that if UCAN authorization chain form a loop it effectively allows anyone to access excercise  it which is why publishing it to IPNI would allow anyone to perform reads

rfc/ipni-w3c.md
Gozala: Precisely this ! If we delegate user claim signed by us, it’s up to them to decide who they share it with. They can share with everyone or just share with selected group. They can also leverage UCAN expiry and notBefore fields to reduce level of access shared.

rfc/ipni-w3c.md
gammazero: Yes, multihash is what we want here. 👍

rfc/ipni-w3c.md
gammazero: I was thinking that a user's claims bucket would be stored in user space. Does that work?

rfc/ipni-w3c.md
gammazero: > we claim that some that (multi)hash A is a byte range of content addressed by (multi)hash B

The intent of this is to record where to retrieve the whole CAR file from, or the CARx-index from. So, byte ranges should not apply here. This is just saying "you can get the CAR file from this URL, and the signer is asserting this to be true".

> not sure if raw cid is better than multihash

I would prefer to use multihashes for everything and only use CIDs as a form of user input for compatibility with other CID-based systems.

rfc/ipni-w3c.md
gammazero: In IPNI, the location of the claims bucket is stored. Any one of the set of multihashes designated by the user resolved to where to get the content claim.

Once the claims bucket is acquired, then the CARx-index location claim is read to get the location of the CARx-index and the CAR file. The CAR location claim is read to get the location of the CAR file.

rfc/ipni-w3c.md
gammazero: That was what I was saying, I think I left out a word. it should read:
"Providing a client-signed source claim allows more nuanced reputation decisions."

So the client should sign the source claim.  Also, the client can be the signer of the IPNI advertisement, signing as the provider (using the key that matches the ProviderID).

rfc/ipni-w3c.md
gammazero: The client can sign the advertisement as the provider. All that is needed is for the client to sign the advertisement using the private key that matches the ProviderID in the advertisement. If the advertisement includes a claim that the provider ID belongs to the client, then we can varify via UCAN that the client created the advertisement and we do not care who publishes it.

IPNI advertisements can be signed by one of two entities, the provider (sig must match ProviderID in ad) or the publisher (sig must match peer ID of host serving the advertisement).

rfc/ipni-w3c.md
gammazero: I can certainly do that. This document may provider some explanation:
https://github.com/ipni/storetheindex-docs/blob/main/filecoin-network-indexer/technical-walkthrough.md

rfc/ipni-w3c.md
gammazero: I think this should be done after fetching and verifying the CAR/CARx. If cannot fetch and verify, then we fail here possibly after some time of retrying.

rfc/ipni-w3c.md
Gozala: I think I was not very clear here let me try to spell out in more concrete terms

1. User asks to upload blob identified by multihash A and receives presigned PUT url.
2. User uploads blob to the pre-signed PUT url.
3. User checks blob status and get UCAN delegation for `assert/location` where subject `with` field is a provider DID (did:web:web3.storage), issuer is a provider DID and audience is space DID.
4. User decides to publish blob on IPNI. To do so they invoke delegated `assert/location` capability.
5. Invocation handler verifies UCAN chain and ensures that user indeed can make claim on behalf of provider `did:web:web3.storage` because the delegation chain therefor it publishes IPNI advertisements with ProviderID (because it corresponds to `with` field)

In other words provider delegates capability to the client to publish IPNI advertisement on it's behalf. As long as client does not escalate (e.g. commitment window) everything checks out during invocation so provider can simply sing and publish advertisement.

Gozala: Broader point I'm trying to make here is that service makes a commitment to the user that they will make content available, so it should be services reputation at stake and not the users of the service.

rfc/ipni-w3c.md
gammazero: > assumes public retrieval 

The data could also be encrypted by the client, and the symmetric decryption key could be encrypted for each allowed reader, as is done with [age-encryption](https://github.com/C2SP/C2SP/blob/main/age.md). While that should probably be left up to the client, it might be nice for the client to have a feature like that built-in.

rfc/ipni-w3c.md
gammazero: OK, This makes sense to encode into the `Addresses` field of an advertisement. This allows there to be multiple addresses that can easily be changed by updating in a more recent advertisement. That address plus the cliaims bucket CID from the metadata will give us what we need.

rfc/ipni-w3c.md
```diff
@@ -0,0 +1,346 @@
...
+
+## Authors
+
+- [gammazero], [olizilla]
+
+## Goals
+
+A user can use a CID that identifies data stored on w3c to lookup the location where the data can be retrieved and content claims describing that data.
+
+The data's location can change and content claims about the data can change, at anytime during the data's storage lifetime, and queried for the data CIDs will be able to get the most locations and claims.
```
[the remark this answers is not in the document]
gammazero: I will check to just say, "w3s can update content claims if necessary without having to republish data multihashes."

rfc/ipni-w3c.md
gammazero: OK, I understand that the service is making a commitment to the user about being able to serve content.

What does this mean in terms of data structures:
> To do so they invoke delegated assert/location capability.

Is this a [UCAN invocation](https://github.com/ucan-wg/invocation/blob/v0.2/README.md#ucan-invocation-specification-v020)? If so, does it contain the CID of the `assert/location` claim that w3s returned as a result of looking up blob status? Also, I do not see a `with` field in that spec.

Is it necessary to include CIDs for both CAR and CARx-index in invocation, or is CAR CID sufficient?

gammazero: @Gozala  When does the assert/location claim get created
> User checks blob status and get UCAN delegation for assert/location

Is the location claim created on-demand when the user checks status? Or, does it get created after the user uploads their blob? If created after upload, then what triggers the creation since we want to get away from bucket events?

rfc/ipni-w3c.md
```diff
@@ -0,0 +1,368 @@
...
+    w3infra-carpark-consumer -->|region/bucket/cid/cid.car| C[/indexer queue/]
+    indexer ---> dynamo[Dynamo\nblocks index]
+```
+
+## Proposal
+Provide a ipni/offer UCAN ability to submit the location of a CAR and a CARx-index that a user has stored with w3s, to make the them discoverable via IPFS implementations and other IPNI consumers.
+
+The ipni/offer is optional, and if not invoked, then the user's data is not indexed.
+
+If opting-in to IPNI indexing, the CAR file and the CARx-index are created by the client. The client may choose which multihashes they want indexed and specify this in the ipni/offer request.
```
vasco-santos, line 153: Like we do in w3filecoin protocol, we should consider here having a `ipni/accept` capability. Offer invocation asks the service to do X, service responds saying I got it and soon I will solve this for you, and by the way, here is the CID of the invocation I am working on. Then user can query for the receipt, which will be done when service ends.

A simplified version may exist when we UCAN1.0 + UCAN Invocation Spec, but I think this is the only way for it to land before that cc @Gozala

rfc/ipni-w3c.md
```diff
@@ -212,6 +204,27 @@ Invoke `ipni/offer` with the CID for an [inclusion claim] that associates a CAR
...
+
+```json
+{
+  "iss": "did:key:Alice",
+  "aud": "did:web:web3.storage",
+  "att": [{
+    "can": "ipni/offer",
+    "with": "did:key:space", // users space DID
+    "nb": {
+        "inclusion": CID   // inclusion claim CID
```
hannahhoward, line 221: thinking of this in terms of evolution, it seems to me we can ship this without the CAR indexing on client side for write everywhere by simply not having the inclusion claim, and then all that is generated in CARPark.

hannahhoward (approved):
It seems like there is still a small bit of tightening on alignment about location claims, but I think we're basically there. I'm LGTM at this point.

rfc/ipni-w3c.md
gammazero: If user wants to publish to IPNI, they invoke ipni/offer request:                                                                              
- Get blob status to check data is stored and get claims bundle CID
- Invoke `ipni/offer` with the claims bundle CID associated with the stored data to index
- User gets back a receipt that contains the advertisement ID
- Can be done once at anytime during stored data lifetime

rfc/ipni-w3c.md
```diff
@@ -2,53 +2,54 @@
...
+A user stores content as opaque blobs containing series of blocks addressed by multihashes. Currently, these blocks are collected into a CAR file for storage as one body of data. This will be generalized into a blob of data that does not require a CAR file.
 
-With an [inclusion claim](https://hackmd.io/@gozala/content-claims?utm_source=preview-mode&utm_medium=rec#Inclusion-Claims), the user who created the CAR asserts that CAR contains a given set of multihashes, transitively, via the CARx-index.
+The user creates an [inclusion claim] that contains the multihash of the data blob and an array of tuples of (multihash, offset, length) of any items the user intends to make individually retrievable from their data blob. This inclusion claim serves as an external index into the user's data that can be stored and retrieved separately from the data. This inclusion claim is also the user's assertion that their data blob contains the data represented by a given set of multihashes.
```
Gozala, line 20: > This inclusion claim serves as an external index into the user's data that can be stored and retrieved separately from the data

I've read this sentence several times and still confused by it. Specifically fact that it mensions **user's data** followed by **the data**. Could this be rephrased to make it bit more clear.

I think you're trying to say that indexed data can be retrieved independent from the blobs it's contained in ? Perhaps embracing term [IPLD Block](https://ipld.io/docs/intro/primer/#blocks-vs-nodes) could be of value to make things more concrete.

Gozala (approved):
Looks pretty good, I did provide bunch of comments and suggestions, but I'll let it up to you what you want to incorporate and what to ignore.

I think this is probably already very detailed for the RFC and probably good idea to merge and iterate on details in the specs (probably can copy & paste this there too).

I'm still wondering if decoupling content claims from location claims might be a good idea that would allow client querying IPNI to choose which location claim to follow independent from which partition claim to follow.

Other thought I had is that it would a good idea if claim bundle CID was not affected by who made those claims that is issuer, audience, and subject. That way multiple users indexing same data would arrive to same claims bundle as opposed to different ones. That would probably require a lot more thinking how to represent index though, but perhaps it is not a bad idea never the less.

rfc/ipni-w3c.md
gammazero: Changing to...

> The user creates an [inclusion claim] that contains the multihash of the data blob and an array of tuples of (\
multihash, offset, length) of any blocks within the blob that the user intends to make individually retrievabl\
e. This inclusion claim serves as an external index into the user's data blob that can be stored and retrieved\
 separately from the user's data blob. The inclusion claim is also the user's assertion that their data blob c\
ontains the blocks represented by the set of multihashes in the inclusion claim.

rfc/ipni-w3c.md
```diff
@@ -2,53 +2,54 @@
...
 
-After storing the CAR and CARx-index, w3s creates a [location claim](https://hackmd.io/@gozala/content-claims?utm_source=preview-mode&utm_medium=rec#Location-Claims) for the CAR file and the CARx-index. Each location claim maps an item's CID to a URL where the item can be retrieved.
+After storing the data blob and the inclusion claim, w3s creates a [location claim](https://hackmd.io/@gozala/content-claims?utm_source=preview-mode&utm_medium=rec#Location-Claims) for the blob and the inclusion claim (serialized to bytes). Each location claim maps an item's CID to a URL where the item can be retrieved.
 
-w3s creates a Content Claims Bucket. This is a container, stored on w3s network, that holds all content claims associated with the stored data. The location claims and inclusion claim are written to the claims bucket. The claims bucket is then stored.
+w3s creates a Content Claims Bundle. This is a container, stored on the w3s network, that holds all content claims associated with the stored data. The location claims and inclusion claim are written to the claims bundle. The claims bundle is then stored in the user's storage space.
```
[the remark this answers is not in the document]
gammazero: This is better, thank you. You are correct that the user creates the inclusion claim and the service creates the location claim.

I think either the service or the user could create the claims bundle. I was thinking the service would create the bundle, since otherwise it would need to verify its presence before generating an IPNI advertisement, and that would require another message from the user to tell the service that the bundle had been written.

rfc/ipni-w3c.md
```diff
@@ -2,53 +2,54 @@
...
 
-[IPNI](https://github.com/ipni/specs/blob/main/IPNI.md) will allow a client to lookup any of the multihashes, that are in the CARx-index, to get information sufficient to recover the claims bucket. The location claims from the claims bucket can then be used to get the CARx-index and CAR data.
+[IPNI](https://github.com/ipni/specs/blob/main/IPNI.md) will allow a client to lookup any of the multihashes from the blob inclusion claim, to get information sufficient to retrieve the claims bundle. The location claims and the inclusion claim can be read from the bundle to allow the client to retrieved the desired pieces of data from the stored data blob.
```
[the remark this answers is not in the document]
gammazero: How about,

> [IPNI](https://github.com/ipni/specs/blob/main/IPNI.md) will tell a client where/how to get the claims bundle when IPNI is queried using any of the multihashes from the inclusion claim. The client can fetch the claims bundle and read the location and inclusion claims from the bundle. The client then uses the information read from these claims to retrieve the desired pieces of data from the stored data blob.

rfc/ipni-w3c.md
```diff
@@ -2,53 +2,54 @@
...
 
 Information is provided to IPNI in the form of [IPNI Advertisements](https://github.com/ipni/specs/blob/main/IPNI.md#advertisements)
-that are signed by w3s and published to IPNI by w3s. Each  Advertisement 
+that are signed by w3s and published to IPNI by w3s. When a new advertisement is published, an [IPNI Announcement](https://github.com/ipni/specs/blob/main/IPNI.md#announcements) is sent to IPNI to trigger IPNI to fetch the new advertisement.
 
-When a new advertisement is published, an [IPNI Announcement](https://github.com/ipni/specs/blob/main/IPNI.md#announcements) is sent to IPNI to trigger IPNI to fetch the new advertisement.
+Each Advertisement is associated with a data blob, and allows IPNI to map all of the multihashes in the inclusion claim to information needed to retrieve the claims bundle.
```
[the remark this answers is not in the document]
gammazero: The multihash of the blob is also included. Should be stated explicitly.

rfc/ipni-w3c.md
```diff
@@ -2,53 +2,54 @@
...
 
 ## Introduction
 
-We publish ad-hoc batches of multihashes to IPNI. This proposal aims to align our usage of IPNI with [content-claims], by publishing an advert per [inclusion claim], and include the source claim in the IPNI Advertisement
+W3s publishes ad-hoc batches of multihashes to IPNI. This proposal aims to align our usage of IPNI with [content-claims], by publishing an advert per stored data blob. This is done so that any of the multihashes to chunks of data within the blob or the blob itself, can be used to look up the content claims bundle associated with the stored data, which in turn locates the data.
```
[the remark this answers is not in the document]
gammazero: You might be getting confused by terminology:

- **IPNI Index**: A mapping of a single multihash to information needed to retrieve the body of data containing the block represented by the multihash.
- **IPNI Advertisement**: Information about the location of a body of data accompanied with all the multihashes of blocks contained within that body of data. An Advertisement is the record, given by an index-provider to IPNI, from which IPNI creates its indexes.

Each advertisement is for an entire blob/claims-bundle. IPNI maps each multihash in the advertisement to the location of the claims bundle. 

This is what indexing refers to, in the sense of database indexing:
The claims bundle is a record of all the blocks within a blob (as well as the blob itself). To find the claims bundle(s) containing a particular multihash would require searching through all the bundles to look for the ones with the multihash. So, the bundle is indexed which means we create a separate index for each multihash, that maps each multihash to the claims bundle.

rfc/ipni-w3c.md
```diff
@@ -93,21 +94,19 @@ curl https://cid.contact/cid/bafybeicawc3qwtlecld6lmtvsndimoz3446xyaprgsxvhd3aap
 ]}]}
 ```
 
-web3.storage publishes the blocks it can provide by encoding a batch of multihashes as an IPLD object and writing it to a bucket as an `Advertisement`, addressed by it's multihash.
+web3.storage will publish the blocks it can provide by encoding a batch of multihashes and information needed to retrieve tha data into an IPNI `Advertisement`. The advertisement is referred to by its CID.
 
-An `Advertisement` includes `Provider` info which claims that the batch of multihashes are available via bitswap or HTTP, and are signed by the providers PeerID private key; Each advert is a claim that this peer will provide that batch of multihashes.
+An `Advertisement` includes `Provider` info which claims that the batch of multihashes are available via some protocol (e.g. bitswap or HTTP), and are signed by the provider's PeerID private key; Each advertisement advertises to IPNI that the stored data, specified by multihashes, is available from the provider.
```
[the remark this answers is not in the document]
gammazero: Need to think about this...

The w3 service is the one publishing an advertisement. I thought we were going to have all of the location claims created by the w3s service, meaning that is where the service understands the data to be located.

We can certainly include location claims created by the user. I was under the impression that we did not want to do this since the service has authority over the location, or at least over verifying the data was at the location.  @hannahhoward any opinion here?

FYI, the above mention of the signature has nothing to do with the location claim or claim bundle, it is just a mechinism to ensure the advertisement was legitimately added to the ad chain.

rfc/ipni-w3c.md
```diff
@@ -93,21 +94,19 @@ curl https://cid.contact/cid/bafybeicawc3qwtlecld6lmtvsndimoz3446xyaprgsxvhd3aap
...
 
-The advert `ContextID` allows providers to specify a custom grouping key for multiple adverts. You can update or remove multiple adverts by specifying the same `ContextID`. The value is an opaque byte array as far as IPNI is concerned, and is provided in the query response. The ContextID also serves as a key that refers metadata, and is used to update or delete that metadata. Updating metadata changes the metadata returned by IPNI lookups for all multihashes that were advertised with that context ID.
+The advert `ContextID` allows providers to specify a custom grouping key for multiple adverts. You can update or remove multiple adverts by specifying the same `ContextID`. The value is an opaque byte array as far as IPNI is concerned, and is provided in the query response. The ContextID also serves as a key that refers to metadata and is used to update or delete that metadata. Updating metadata changes the metadata returned by IPNI lookups for all multihashes that were advertised with that context ID.
```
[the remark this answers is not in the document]
gammazero: This is called out explicitly later. The above section describes IPNI in general and does not talk about specific values used for `ContextID`. 

The `ContextID` is the blob CID, not the bundle CID. The reason for this is that the bundle CID, contained in the metadata, may change if the bundle is modified, but the bundle is still associated with the same blob. The `ContextID` is a key for the metadata, so the `ContextID` remains the same if the metadata is updated.

rfc/ipni-w3c.md
```diff
@@ -93,21 +94,19 @@ curl https://cid.contact/cid/bafybeicawc3qwtlecld6lmtvsndimoz3446xyaprgsxvhd3aap
...
 
-A `Metadata` field is also available for provider specific retrieval hints, that a user should send to the provider when making a request for the block, but the mechanism here is unclear _(HTTP headers? bitswap?)_.
+A `Metadata` field is also available for provider specific retrieval information, that a user should send to the provider when making a request for data. The metadata can be updated independently of all the multihashes indexes. In our case this field will contain the identifier of the claims bundle that is associated with the blob of stored data.
```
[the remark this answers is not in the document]
gammazero: > Is there case where we imagine updating metadata without updating index

Yes. If the claims bundle is modified, thereby creating a different CID for the claims bundle, then the metadata will be updated with the new CID, but indexing remains the same.

> I imagine us wanting to rotate location claims without updating inclusion claims and ideally users would not have to recreate claim bundles only to rotate location claims.

This is all true. I did not know if rewriting the claims bundle, after only changing the location claims, was enough of a burden to justify separating these things. So, at least for now, there is just one claims bundle that gets mutated.

rfc/ipni-w3c.md
```diff
@@ -146,67 +145,93 @@ flowchart TD
...
+    w3s-)ipni: announce head (CID)
     deactivate w3s
     ipni-->>w3s: fetch advert
     activate ipni
     ipni-->>ipni: index entries
     deactivate ipni
 ```
+- User creates their data blob and inclusion claim, then uploads these to w3s to the pre-signed PUT URL.
+  - The inclusion claim listing sub-blocks within the blob is created by the client whether or not they intend to have their data indexed. This is because these blocks may still be needed to find data within the blob, and to allow indexing to be done at a later point without having the re-read the entire blob.
+  - W3s does not need to immediately verify the correctness of the inclusion claim, but can at any time and then replace the inclusion claim with its own to certify that it was verified.
```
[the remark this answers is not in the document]
gammazero: OK, that's fine. If the verified w3s claim is there, then it seemed like there was no reason to keep the user-created claim. Also, if w3s moves the blob to a new location later on then the user's claim will no longer be correct.

rfc/ipni-w3c.md
```diff
@@ -146,67 +145,93 @@ flowchart TD
...
+- w3s creates the location claim for the blob after verifying the content was uploaded. Multiple location claims are created if the data is partitioned into multiple blobs or if the blob is stored in multiple locations.
+  - The client triggers w3s to create the blob location claim(s) when the client is done uploading the data blob(s) to the URLs issued by w3s. By creating the location claims, w3s is responsible for the location of the data, and should verify the data is present before creating the claims.
```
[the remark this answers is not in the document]
gammazero: I thought we gave the user a URL to PUT the data at. If that is the case, then how will we know when the user has uploaded their data to that location? This is why I specified that the user will tell w3s.

It seems like what you are saying is that the user is going to upload the data to w3s. This means that w3s must then upload that to the storage location. That would incur a double ingress cost.

Need to make a decision here.

rfc/ipni-w3c.md
```diff
@@ -146,67 +145,93 @@ flowchart TD
...
+  - After triggering location claims creation following blob upload, the client can then issue a request to index the data. This request can be made once anytime during the data's lifetime.
```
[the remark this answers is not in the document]
gammazero: The client sends a request to w3s to have the blob indexed using IPNI. This request tells the w3s service to create an IPNI advertisement to tell IPNI what to index. IPNI reads the advertisement published by w3s, and creates an index (mapping of multihash to location) for each multihash in the advertisement.

rfc/ipni-w3c.md
```diff
@@ -146,67 +145,93 @@ flowchart TD
...
-    "location": "https://r2.cf/bag...car"
+  "cmd": "assert/inclusion",
+  "sub": "did:key:alice",
+  "args": {
+    "content" : blob-hash,
+    "parts" [
+      [some-block-mh, 128, 32],
+      [other-block-mh, 160, 30],
+      [bafy...cat, 190, 100]
+    ]
```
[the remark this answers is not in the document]
gammazero: Added note to RFC

rfc/ipni-w3c.md
```diff
@@ -146,67 +145,93 @@ flowchart TD
...
+```
+```json
+{
+  "cmd": "assert/location",
+  "sub": "did:web:web3.storage",
+  "args": {
+    "content" : blob-hash,
+    "url": https://w3s.link/blob-hash
+    "headers": {}
+    "range": [0,1280]
```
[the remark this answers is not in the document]
gammazero: I think that a location claim should claim one location, instead of bundling multiple locations of some data. Consider the situation where some of the locations are verified by someone and others are not. With single-location claims it seems easier to search for other matching claims from the verifying issuer, instead of looking in different multi-location claims to find matches... but could see this resulting in opinions either way for differen t use cases. so...

Let's do the simplest thing until there is a clear reason to do something more complex.

rfc/ipni-w3c.md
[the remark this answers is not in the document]
Gozala: That’s a lot clearer thanks

rfc/ipni-w3c.md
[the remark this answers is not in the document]
Gozala: I suppose it’s fine, even though I do find it bit counterintuitive

rfc/ipni-w3c.md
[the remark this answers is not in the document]
Gozala: That assumes that our claims are superior, but in reality it’s just more trustworthy from our own perspective. On the other hand others may trust claim issuer more than they trust us. Does that make sense ? I don’t think a our claims as replacement for user claims they are just ours so we can trust them.

> Also, if w3s moves the blob to a new location later on then the user's claim will no longer be correct.

We should not move for the duration of claim we issued or we aren’t upholding our commitment. That said, that is also why I’m somewhat biased for having location claims separate form inclusion ones, because that way later is agnostic of former

rfc/ipni-w3c.md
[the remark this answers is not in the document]
Gozala: You are correct on invocation client gets URL where it uploads the content. I’m just trying to avoid specifying details on how we discover that upload was complete. And trying to highlight that when they’re done uploading they’ll get commitment a.k.a location claim from us.

perhaps I didn’t phrased it very clearly,  feel free to change it.

rfc/ipni-w3c.md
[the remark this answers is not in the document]
Gozala: I think I was thinking of bundle as index & was confused here.

rfc/ipni-w3c.md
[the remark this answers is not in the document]
gammazero: All good. I used a slight editing of your words.

rfc/ipni-w3c.md
gammazero: Once the advertisement CID is returned, then the advertisement has been published to IPNI. I don't think there is a need to make this async. The async part is waiting for IPNI to index the advertisement. That is not something we control, so the client will just have to query IPNI if they want to see when their stuff gets indexed.

PR #18, opened 2024-03-28
RFC: networked-dot-storage

hannahhoward:
[Preview](https://github.com/web3-storage/RFC/blob/rfc/networked-dot-storage/rfc/networked-dot-storage.md)

### DIFF
--- a/rfc/networked-dot-storage.md
+++ b/rfc/networked-dot-storage.md
@@ -1,0 +1,174 @@
+# Networked Dot Storage
+
+> or what does decentralziation even mean?
+
+## Authors
+
+- [Hannah Howard], [Protocol Labs]
+
+## Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC2119](https://datatracker.ietf.org/doc/html/rfc2119).
+
+## Abstract
+
+This RFC puts forward a proposal for the broad parameters of "decentralization" in W3S. It defines both the goals for "decentralization" and identifies where resources should be invested first on the path to "decentralization".
+
+## Non-goals
+
+1. Addressing anything related to tokens
+2. W3S corporate structure or nucleation plans
+
+## Background
+
+The current state of the world for web3.storage is as follows:
+
+- Web3.storage UCAN protocols are open and well specified. 
+
+- We've now created a version of these protocols that can be run from any machine-- so anyone can spin up theoretically own version of our service.
+
+- Seperately Web3.storage backs up data to Filecoin, via open protocols.
+
+- We are actively moving towards supporting running on top of underlying storage mechanisms that lack rich event functionality (a.k.a. bucket events)
+
+## Why Go Further
+
+Today, someone who is not the web3.storage team could spin up their own version of web3.storage, using the open source software we've published and the protocols we've specified. No one has done this in practice, but we are increasingly close to the point where it's plausable. Is this "sufficient decentralization"?
+
+One proposed path forward is to not pursue "decentralization" beyond trying to make other nodes a reality, perhaps through incentivization, and leaning heavily into product improvements that differentiate our platform. We could aim for "S3 if AWS invested in open source S3 implementations and made it content addressed".
+
+This is not the path we are going to pursue. It feels important to say that up front.
+
+But before we go deeper into where we are headed, it's worth looking at the implied dichotomy between "decentralization" and product features. That this dichotomy is almost reflexively referred to likely results from failing to appropriately frame "decentralization" in terms of product features. So let's start by framing the product features we want to unlock through decentralization.
+
+## The Network We Want
+
+### No Data Loss
+
+1. Users should be able to specify how many unique copies of their data they want backed up, and the W3S network should handle insuring that many copies remain backed up as long as the user pay to store in W3S.
+
+### Perpetuity
+
+1. Users should be able to store hot data in W3S that will be available for  retrieval until they stop paying for it. They should not have to worry about any single business entity holding their data disappearing. If it does, the data should go somewhere else without their needing to intervene.
+
+### Sensible defaults with customization
+
+1. Users should have the option to store with simply "the W3S network" rather than a specific business entity or Filecoin SP. This is a hard lesson learned from Filecoin that forcing users to choose from a marketplace where information on available options is hard to come by is a terrible user experience.
+
+2. Users should still have the option to choose a provider they like, or to move between providers.
+
+### Trust
+
+In general, users should be able to "trust" W3S to handle their data well and meet service level agreements. That trust should be derived by giving users the ability to verify any actions taken on the W3S network. 
+
+1. Users should be able to gather a detailed history of what has happened with their data when they put it on the W3S Network.
+
+2. A user paying for storage should be able to verify that data is being stored.
+
+3. A user paying for storage should be able to verify it can be retrieved quickly. 
+
+4. A user paying for multiple backups should be able to verify each backup is still functioning.
+
+4. If verification demonstrates that a particular provider is failing to provide a service that the user has paid for, the network should detect the failures and move their data automatically. (or alternatively, the user should have the ability to create a dispute and move to a different provider while getting money back)
+
+### Distribution of resources
+
+6. Users should not have to worry about large service providers establishing monopolies that reduce user choice or drive up costs. To insure this, W3S should emphasize data portabiltity, fairness in data distribution, and low barrier to entry for new providers.
+
+## But who wants this?
+
+A common refrain in response to some of the positive properties that "decentralization" might enable is "none of our existing users asked for this". 
+The truth is that many of these features seem far out of reach in the current architecture of the internet. The way users have interacted with the internet, shaped by the business interests of large tech companies, has told them for three decades these features are not possible. Even so, we've already spoken to potential customers who've expressed concrete interest in this kind of product.
+
+Content addressing provides the core from which to start building toward something better. UCAN is another key component. Filecoin is another part of the puzzle. Blockchains in general are a key component (yes trustless distributed consensus matters). Our job is to fill in the remaining missing pieces.
+
+## Next steps
+
+Taken as a whole, yes, this is absurdly ambitious. It IS a more complex product (especially since it's a network, not a product). Getting every property on the list above is beyond a 1 year scope. Also, W3S IS going to focus on other product features besides "decentralization", particularly those important to market segments we can get to while we build towards our end state.
+
+But, yes, we're going to try to build this. So given that, what missing pieces SHOULD be built in the next six months? 
+
+**This is a loosely prioritized of inituitions. It's not final and it's open to discussions.**
+
+*Btw, we're not alone here -- because we're ultimately building a network, a critical one to the overal PLVerse vision, we can draw from other teams, and eventually, some of the required components may not be built by the W3S*
+
+### Other implementations 
+
+The first step on the path for W3S is indeed to build other implementations of w3up. Local.storage is our first experiment. It would be nice to produce a cloudflare only implementation. The architecture proposed in [Core Node Architecture](https://w3sat.notion.site/Filecoin-SPs-in-the-W3S-Network-16a9b66e7ef3425bbee390c6481bd752?pvs=4) is a kind of aspirational target -- an ultraportable implementation that could run in most server environments, be ultra-sandboxed, provide great update mechanisms, and support rapid iteration.
+
+We should constinue to explore these options.
+
+### Router.storage
+
+We need an implementation of W3up that simply routes to other instances of W3Up. Building this implies several tasks:
+- deciding who gets each piece of content
+- how to manage UCAN delegations across relays
+
+And eventually:
+- how to allow nodes to join and how to decide when to kick them out
+
+But first, we should probably just hardcode a list of nodes.
+
+*** But isn't this a point of centralization if everyone uses it? ***
+
+Absolutely. It's important that key coordination functions in the network like router.storage eventually operate through distributed consensus and produce verifiable results.
+
+### Read Pipeline that finds content from many write pipelines
+
+A retrieval service (i.e. Freeway) should not be tied to a specific instance of w3up. This implies some form of global content discovery. [W3 IPNI Indexing RFC](https://github.com/web3-storage/RFC/blob/main/rfc/ipni-w3c.md) is a major step in this direction. 
+
+Eventually we may need some kind of regional caching.
+
+### Repair service for both W3UP and Filecoin
+
+Repair service means a service that creates permaneant reliable storage from a series of time-limited storage deals with providers who might fail.
+
+It's a vital missing component today on Filecoin AND a missing component of W3S as soon as their is more than one instance. We've already encountered this problem indirectly as we've thought about deal renewals and the reality that deal terms on Filecoin do not coincide in time with users joining and leaving W3UP.
+
+It needs to exist for Filecoin AND it needs to exist for W3S.
+
+*** But isn't this a point of centralization if everyone uses it? ***
+
+Absolutely. It's important that key coordination functions in the network like a repair service eventually operate through distributed consensus and produce verifiable results.
+
+### Shared w3filecoin pipeline
+
+W3Filecoin is an excellent service for aggregating data on to Filecoin. At the same time, running it is NOT easy. While the basic aggregation functions are simple, making Filecoin deals is not. Multiple instances of W3UP should be able to share an instance of W3Filecoin. 
+
+*** But isn't this a point of centralization if everyone uses it? ***
+
+Absolutely. It's important that key coordination functions in the network like w3-filecoin eventually operate through distributed consensus and produce verifiable results.
+
+### Verification Mechanisms For Retrievals from W3UP Instances
+
+See [Proofs Overview](https://w3sat.notion.site/Filecoin-SPs-in-the-W3S-Network-16a9b66e7ef3425bbee390c6481bd752?pvs=4)
+
+It is important that the W3UP service level agreements are not compromised across the network by poorly performing or malicious W3Up instances.
+
+### Filecoin.Storage
+
+Produce an implementation of W3UP and W3Filecoin that can be run by a Filecoin L1 Storage Provider. In order words, a Filecoin storage provider should be able to be more than a backup on our network -- and eventually get seperate compensation -- if they can meet our service level requirements. See [Filecoin SPs in the W3S Network](https://w3sat.notion.site/Filecoin-SPs-in-the-W3S-Network-16a9b66e7ef3425bbee390c6481bd752?pvs=4)
+
+## FAQ
+
+### Is decentralization the wrong word to use?
+
+Decentralization is both a concept so overloaded it lacks any coherence, and a narrative that our product team has validated is indeed important to many market segments in web3. Actual customers have told us they won't use our product unless it's "decentralized". Likely, in some cases this is a novel kind of CYA (cover-your-ass) -- where enterprise customers want SOC-2 compliance in a data center so they can blame someone else if data gets lost, web3 customers want services they use to be "decentralized" so they can't get called out by the internet for not being "web3". Decentralization is both an annoying meme and an important narrative we have to contend with.
+
+I would say what we're really building is an incentivized permissionless cooperative distributed network for content addressed data. If anyone wants to try to market that, I encourage them to. But internally, maybe we should consider not saying decentralization as much. At minimum, let's align on what we're trying to achieve for our users.
+
+### Does decentralization mean running W3S on a network like Saturn?
+
+Saturn is an interesting experiment in crypto incentization of a retrieval network. There is a lot to learn from that experience. But there is as much to learn about what not to do as what to do. In my opinion, Saturn is over indexed on diversity of small nodes and under indexed on decentralization of coordination. Also, our revenue model with Saturn clearly needs improvement. At minimum money in needs a connection to money out.
+
+### What about quality of service?
+
+Web3.storage users love our reliability, ease of use, and performance. In the content addressed world, these are easily our strongest differentiators. As @alanshaw has pointed out several times, in general, we MUST maintain our level of service as we evolve the network. That is a north star. The world already has enough broken decentralized systems.
+
+### Does "distributed consensus and produce verifiable results" mean smart contracts on a blockchain?
+
+Most likely, yes. But I'm open to hearing about other consensus mechanisms. That said, for the specific problem of trustless consensus on limited execution of relatively small rule sets, honestly, I think blockchains are the most proven solution to date. We need to assess objectively the risks/benefits of a different approach.
+
+### I noticed you didn't put some kind of decentralized CDN on the six month plan
+
+Yes, this is correct. Freeway is an open source project that can be run locally, or you can use the instance we run in Cloudflare (that is widely available and cached due to Cloudflare's CDN). The most immediate goal is to support multiple storage instances. Eventually, we might build an incentivized freeway retrieval network. Or, someone else might.


rfc/networked-dot-storage.md
```diff
@@ -0,0 +1,174 @@
...
+
+This is not the path we are going to pursue. It feels important to say that up front.
+
+But before we go deeper into where we are headed, it's worth looking at the implied dichotomy between "decentralization" and product features. That this dichotomy is almost reflexively referred to likely results from failing to appropriately frame "decentralization" in terms of product features. So let's start by framing the product features we want to unlock through decentralization.
+
+## The Network We Want
+
+### No Data Loss
+
+1. Users should be able to specify how many unique copies of their data they want backed up, and the W3S network should handle insuring that many copies remain backed up as long as the user pay to store in W3S.
```
vasco-santos, line 48: some food for thought: more than replication, I have always liked the idea of allowing user custom region, or even selecting providers with different SLAs. For instance, one node running this service could host things in S3 EU Region, with greater SLA characteristics. Storage costs SHOULD not be tied to what web3.storage says, but what each provider offers to enable a more open market

rfc/networked-dot-storage.md
```diff
@@ -0,0 +1,174 @@
...
+
+1. Users should be able to gather a detailed history of what has happened with their data when they put it on the W3S Network.
+
+2. A user paying for storage should be able to verify that data is being stored.
+
+3. A user paying for storage should be able to verify it can be retrieved quickly. 
+
+4. A user paying for multiple backups should be able to verify each backup is still functioning.
+
+4. If verification demonstrates that a particular provider is failing to provide a service that the user has paid for, the network should detect the failures and move their data automatically. (or alternatively, the user should have the ability to create a dispute and move to a different provider while getting money back)
```
vasco-santos, line 72:
```suggestion
5. If verification demonstrates that a particular provider is failing to provide a service that the user has paid for, the network should detect the failures and move their data automatically. (or alternatively, the user should have the ability to create a dispute and move to a different provider while getting money back)
```

vasco-santos: I liked the idea @Gozala pointed out of both the service, as well as the providers should put some collateral that can be collected when they fail to meet their promises

vasco-santos (commented):
@hannahhoward thanks for putting this together!

rfc/networked-dot-storage.md
prodalex: > I liked the idea @Gozala pointed out of both the service, as well as the providers should put some collateral that can be collected when they fail to meet their promises

@vasco-santos I am not sure if you read the crypto economic requirements proposal. This contains the collateral requirement. We will need this for all the nodes (storage, retrieval, index)

vasco-santos: oh cool, no I did not know about that requirements document IIRC

rfc/networked-dot-storage.md
```diff
@@ -0,0 +1,174 @@
...
+
+- We are actively moving towards supporting running on top of underlying storage mechanisms that lack rich event functionality (a.k.a. bucket events)
+
+## Why Go Further
+
+Today, someone who is not the web3.storage team could spin up their own version of web3.storage, using the open source software we've published and the protocols we've specified. No one has done this in practice, but we are increasingly close to the point where it's plausable. Is this "sufficient decentralization"?
+
+One proposed path forward is to not pursue "decentralization" beyond trying to make other nodes a reality, perhaps through incentivization, and leaning heavily into product improvements that differentiate our platform. We could aim for "S3 if AWS invested in open source S3 implementations and made it content addressed".
+
+This is not the path we are going to pursue. It feels important to say that up front.
```
travis, line 40: > This is not the path we are going to pursue.

Well, it's funny - what's described below is a lot closer to what I meant by "S3 if AWS invested in open source S3 implementations and made it content addressed" than what I understood we were aiming to build, and the differences are things I feel pretty positive about - coordinating different w3up operators in order to make the user experience really good, exposing more of the details of what Filecoin is providing (eg, teaching the end user about the proofs that their data is replicated across the Filecoin network and exposing them in a user friendly way in our UI), data portability and low barriers to entry to getting a node running, etc

tbh after reading this line I was expecting to be disappointed by the rest of this doc but it's just the opposite - maybe just a masterful use of expectations ;-p 

anyway TLDR I'm pretty positive about this direction, thanks for putting this together! will leave more thoughts inline below...

rfc/networked-dot-storage.md
```diff
@@ -0,0 +1,174 @@
...
+
+## The Network We Want
+
+### No Data Loss
+
+1. Users should be able to specify how many unique copies of their data they want backed up, and the W3S network should handle insuring that many copies remain backed up as long as the user pay to store in W3S.
+
+### Perpetuity
+
+1. Users should be able to store hot data in W3S that will be available for  retrieval until they stop paying for it. They should not have to worry about any single business entity holding their data disappearing. If it does, the data should go somewhere else without their needing to intervene.
```
travis, line 52: Love this. At some point I'd really like to game out "what would it look like if w3s shut down suddenly" from the perspective of our users. Really love the idea of centering "our users won't be screwed if we go away" and I'm sensitive to the fact that this is in tension with the traditional startup virtue of having a "defensible business" - I think there is going to be pressure from investors to ensure that our users are dependent on us and can't leave and I think we should be ready to push back on that when it comes.

rfc/networked-dot-storage.md
```diff
@@ -0,0 +1,174 @@
...
+
+### Sensible defaults with customization
+
+1. Users should have the option to store with simply "the W3S network" rather than a specific business entity or Filecoin SP. This is a hard lesson learned from Filecoin that forcing users to choose from a marketplace where information on available options is hard to come by is a terrible user experience.
```
travis, line 56: >  Users should have the option to store with simply "the W3S network" rather than a specific business entity or Filecoin SP
ok you've got me! 🚫 ⛓️ 💊 

> forcing users to choose from a marketplace where information on available options is hard to come by is a terrible user experience.
💯 💯 💯

rfc/networked-dot-storage.md
```diff
@@ -0,0 +1,174 @@
...
+
+5. If verification demonstrates that a particular provider is failing to provide a service that the user has paid for, the network should detect the failures and move their data automatically. (or alternatively, the user should have the ability to create a dispute and move to a different provider while getting money back)
+
+### Distribution of resources
+
+6. Users should not have to worry about large service providers establishing monopolies that reduce user choice or drive up costs. To insure this, W3S should emphasize data portabiltity, fairness in data distribution, and low barrier to entry for new providers.
+
+## But who wants this?
+
+A common refrain in response to some of the positive properties that "decentralization" might enable is "none of our existing users asked for this".
```
travis, line 80: Speaking for myself, the primary thing that I don't think anyone is asking for is "I want the bits of my upload to be stored in hot storage hosted by random participants in the w3s network." 

I'm _very_ sympathetic to the idea that people don't always want to think about where their bits actually live, so I do think we should provide a "default" storage location and I don't necessarily think the healthiest way to build that is to have us (eg, w3s-maintained S3 or equivalent) be the one and only default, BUT I _do_ think that things get a lot lot lot more complicated and risky when we try to rely on an incentivized permissionless network for this, imho in a way that is out of proportion to the amount of apparent market demand for this specific property.

Data portability, verifiability of actions taken on the network, sensible defaults with the ability to customize, storage perpetuity - all of these things seem (to me) clearly valuable and desired by our customers and the market at large, but I don't think any of them require a part of our system that can distribute the actual bits to a random group of w3up operators. The bits that might utilize this - a desire to avoid becoming a monopoly (defacto or otherwise), resiliance in the face of a catastrophic failure of the w3s organization - do seem valuable to me, but I'm not convinced that "a permissionless network of hot storage nodes as the default destination for the bits" is the simplest or most elegant solution to either of these. Maybe I'm wrong - maybe this is the only way to achieve these goals! - but I would definitely like to spend more time kicking this around before we go down this road, because it feels like a very very risky path to take.

rfc/networked-dot-storage.md
```diff
@@ -0,0 +1,174 @@
...
+
+3. A user paying for storage should be able to verify it can be retrieved quickly. 
+
+4. A user paying for multiple backups should be able to verify each backup is still functioning.
+
+5. If verification demonstrates that a particular provider is failing to provide a service that the user has paid for, the network should detect the failures and move their data automatically. (or alternatively, the user should have the ability to create a dispute and move to a different provider while getting money back)
+
+### Distribution of resources
+
+6. Users should not have to worry about large service providers establishing monopolies that reduce user choice or drive up costs. To insure this, W3S should emphasize data portabiltity, fairness in data distribution, and low barrier to entry for new providers.
```
travis, line 76: > data portabiltity

Love.

> low barrier to entry for new providers

Love!

> fairness in data distribution

This is the part I don't really get. Fairness in what sense?

rfc/networked-dot-storage.md
```diff
@@ -0,0 +1,174 @@
...
+- how to manage UCAN delegations across relays
+
+And eventually:
+- how to allow nodes to join and how to decide when to kick them out
+
+But first, we should probably just hardcode a list of nodes.
+
+*** But isn't this a point of centralization if everyone uses it? ***
+
+Absolutely. It's important that key coordination functions in the network like router.storage eventually operate through distributed consensus and produce verifiable results.
```
travis, line 114: > operate through distributed consensus

In this future, isn't the consensus mechanism itself then a point of centralization and a single point of failure? If the blockchain that does this consensus is down, does that mean the network stops working? A storage network like this feels like something that should be able to splinter and fragment to adapt to the needs of sub-groups, and a single (even permissionless and distributed) consensus mechanism feels to me like it sits in tension with some of the user-facing benefits of "decentralization." Maybe IPC solves this!

rfc/networked-dot-storage.md
```diff
@@ -0,0 +1,174 @@
...
+
+Repair service means a service that creates permaneant reliable storage from a series of time-limited storage deals with providers who might fail.
+
+It's a vital missing component today on Filecoin AND a missing component of W3S as soon as their is more than one instance. We've already encountered this problem indirectly as we've thought about deal renewals and the reality that deal terms on Filecoin do not coincide in time with users joining and leaving W3UP.
+
+It needs to exist for Filecoin AND it needs to exist for W3S.
+
+*** But isn't this a point of centralization if everyone uses it? ***
+
+Absolutely. It's important that key coordination functions in the network like a repair service eventually operate through distributed consensus and produce verifiable results.
```
travis, line 132: Similar question as above re: distributed consensus - I don't think this is a point of centralization if there are other nodes on the network that can perform this service for people. Maybe the blockchain provides a coordination point so people can easily hand off to other "repair services" - some level of that makes sense to me, though I still worry about it being a single point of failure in that process!

rfc/networked-dot-storage.md
```diff
@@ -0,0 +1,174 @@
...
+
+### Shared w3filecoin pipeline
+
+W3Filecoin is an excellent service for aggregating data on to Filecoin. At the same time, running it is NOT easy. While the basic aggregation functions are simple, making Filecoin deals is not. Multiple instances of W3UP should be able to share an instance of W3Filecoin. 
+
+*** But isn't this a point of centralization if everyone uses it? ***
+
+Absolutely. It's important that key coordination functions in the network like w3-filecoin eventually operate through distributed consensus and produce verifiable results.
```
travis, line 140: ditto, same, "I used a blockchain to get your data to the blockchain" decentralized inception?! 

but tbh of all the "exit to consensus" questions this one makes the most sense to me? the filecoin pipeline is already so async that a blockchain outage slowing it down feels fine, and having all of the intermediary steps show up on chain somewhere seems pretty reasonable.

rfc/networked-dot-storage.md
```diff
@@ -0,0 +1,174 @@
...
+
+### Filecoin.Storage
+
+Produce an implementation of W3UP and W3Filecoin that can be run by a Filecoin L1 Storage Provider. In order words, a Filecoin storage provider should be able to be more than a backup on our network -- and eventually get seperate compensation -- if they can meet our service level requirements. See [Filecoin SPs in the W3S Network](https://w3sat.notion.site/Filecoin-SPs-in-the-W3S-Network-16a9b66e7ef3425bbee390c6481bd752?pvs=4)
+
+## FAQ
+
+### Is decentralization the wrong word to use?
+
+Decentralization is both a concept so overloaded it lacks any coherence, and a narrative that our product team has validated is indeed important to many market segments in web3. Actual customers have told us they won't use our product unless it's "decentralized". Likely, in some cases this is a novel kind of CYA (cover-your-ass) -- where enterprise customers want SOC-2 compliance in a data center so they can blame someone else if data gets lost, web3 customers want services they use to be "decentralized" so they can't get called out by the internet for not being "web3". Decentralization is both an annoying meme and an important narrative we have to contend with.
```
travis, line 156: > Actual customers have told us they won't use our product unless it's "decentralized".

Right totally, it's just that as you describe they all mean slightly different things, so when they say this we need to dig deeper and figure out which bits they genuinely find important. Our product team coming back from a call with a customer and saying "they won't use us unless we're decentralized" is worse than no feedback at all - it just throws a log on every little fire we each maintain to our preferred definitions of the word and makes these conversations harder and less concrete. I don't think we necessarily need to stop using the word entirely, but the phrase "we need to do X because it makes us more decentralized and that's what our customers want!" is up there with "electrolytes, it's what plants crave!" in the pantheon of misguided sentiments.

rfc/networked-dot-storage.md
```diff
@@ -0,0 +1,174 @@
...
+
+I would say what we're really building is an incentivized permissionless cooperative distributed network for content addressed data. If anyone wants to try to market that, I encourage them to. But internally, maybe we should consider not saying decentralization as much. At minimum, let's align on what we're trying to achieve for our users.
```
travis, line 158: incentivized permissionless cooperative distributed network for content addressed data

IPCDNCAD.... it's got IP, CDN and CAD in it 🤣 

it is definitely wordy, but honestly I think this is a great starting point for thinking about how we do market this - different pieces will matter to different people but hey - more markets we can address!

rfc/networked-dot-storage.md
travis: this sounds good! 

fwiw slashing mechanisms always make me nervous because they open up a big new vector for attacks on the network, ie, malicious actors creating conditions where a network participant can be slashed "unfairly" - for example by DDOSing a single network participant with the explicit goal of hurting them and the larger network

I've read compelling arguments that slashing can actually make a network more fragile, but unfortunately can't find it right now - in lieu of that here are two takes from Emin Gun Sirer (founder of Avalanche and arguably the inventor of proof-of-work) and Anatoly Yakovenko that point in that direction (apologies for linking to the hellsite):

https://twitter.com/el33th4xor/status/1294773537128034304
https://twitter.com/aeyakovenko/status/1295357700570660865

This doesn't mean that slashing is bad for our system, but imho it introduces a LOT of additional complexity for questionable value.

travis: this at least gets into some of the complexity here: https://a16zcrypto.com/posts/article/the-cryptoeconomics-of-slashing/

rfc/networked-dot-storage.md
hannahhoward: the idea of trying to market that phrase was intended as a joke :P

hannahhoward: but please, have a go.

rfc/networked-dot-storage.md
hannahhoward: I mean this service would likely run as a Filecoin L2, so really it's just a sub routine :P

rfc/networked-dot-storage.md
hannahhoward: yea perhaps repair service isn't the right thing for a blockchain, cause you generally might have some level of choice there.

here's the way I look at it:
- single service = single point of failure
- multiple services = no single point of failure, lots of choices, but no coordination, which often reduces user value
- blockchain = acts like single service, but is actually resilient no single point of failure system, and produces more trustworthy output (usually) -- there are times you want this, times you really just want to give people choices. 

Also, personally I do think some of the large L1s can be considered pretty reliable at this point. (though utility is less proven)

rfc/networked-dot-storage.md
hannahhoward: Personally I do think some of the large L1s can be considered pretty reliable at this point.

Filecoin hasn't halted since 2020
Ethereum went a year without halting (surprising though it had one this recently)
Bitcoin's last halt was apparently long enough ago that I can't find any information on it.

I will assure everyone we will not be building shit on Solana which breaks all the time. :P

hannahhoward: re: other stuff -- again, I think there's a complicated issue around whether people want choice or simply something that distributes things fairly. I just disagree there isn't value to consensus in this.

rfc/networked-dot-storage.md
hannahhoward: Essentially, for a user who doesn't want to make choices about their data, they should have assurance the network isn't betting only on large players, and is insuring they get the best service regardless of size. And, they should be able to verify that.

rfc/networked-dot-storage.md
hannahhoward: eventually, I see a heterogenous network that employs a mix of permissionless nodes as well as more trusted gaurantors of service quality. At minimum, if a user gives us data, we shouldn't hand it out randomly -- we should hand it to those who have the best reputation.

I've noticed there's a certain assumption here of "we'll build it like Saturn" which is why I wanted to specifically address that below. I absolutely agree it's silly to scale to a huge permissionless network until there is demand. I call out the lack of economic connection between supply and demand to be perhaps Saturn's biggest mistake. At the same time, I don't personally think an entirely federated system is the end state. Does it look like Saturn? Probably not. But I do want us to get to a world where anyway can be part of it, and at least take a shot at offering a better service for a lower price, whether they are large or small.

rfc/networked-dot-storage.md
hannahhoward: That's fair. 

I will say something, which I think many people are talking about similar underlying things through different language, leading to a perception of difference in goals that is outsized (even if there are real, if smaller differences underneath). My thought is that describing user goals is what can produce better conversations.

travis: > people are talking about similar underlying things through different language, leading to a perception of difference in goals that is outsized 

agree 100% - super grateful for this document as I think it does a great job of 

> describing user goals is what can produce better conversations.

rfc/networked-dot-storage.md
travis: Fair, and fwiw I don't disagree with any of this - I am mostly just harping on internal use of "decentralization" without more context on what we mean specifically when we're talking about the nuts and bolts of what we want to build and why - totally agree that we can and should be taking advantage of its value as a marketing meme! 

If anything I think we have been too shy about highlighting the parts of the current state of the project that imho _are_ "decentralized" in meaningful ways simply because our backend happens to be implemented in terms of a well known cloud service provider rather than being "hosted on the decentralized web" somehow - open specs, open software, (nearly) trivial deployment of third party w3up instances are all things we have right now that address what I think are the interesting user-facing facets of "decentralization" and I think we should lean into that.

> part of the enthusiasm for decentralization is simply an ill defined impulse for a better kind of service on the internet

hard agree, this is why I'm here for sure! and the reality of this is part of why I get so worked up by imprecise usage of this term when we talk about what we want to build - I think there's a lot of real value here that tends to get overshadowed by the marketing/vibes-y usages of it, and I _really_ want to be building things that address this ill-defined desire rather than something that just uses the term to get pageviews - I think I'm in good company on that front!

rfc/networked-dot-storage.md
travis: hahaha - it's probably a little too wordy to be a single marketing meme, but we should probably break it down and see what kinds of fun memetics we can get out of combinations of it - a cooperative network for content addressed data sounds dope as heck!

rfc/networked-dot-storage.md
travis: > disagree there isn't value to consensus in this

totally on the same page - I probably come off as a blockchain hater but inside of me is a computer scientist that thinks they are the coolest thing and desperately wants to find really cool ways to use them. I've just found that it's safer to be a blockchain skeptic by default and to try solving problems nearly any other way before committing to this route. I genuinely hold a lot of hope that we have a real usecase for them here!

rfc/networked-dot-storage.md
Gozala: This is a nuance, but I can't help but wonder if making users choose number of unique copies is the right approach, my guess is users want some resilience, perhaps geographic distribution and other forms of assurances that data availability is adequate. Trying to figure out number of replicas seems several steps disconnected from those needs and also would remove some flexibility on the implementer side.

rfc/networked-dot-storage.md
Gozala: If w3s goes away there is no longer an entity user was paying for service, which would mean user stopped paying so data can disappear, is it not ?

I also wonder if entity providing this assurance should be different from the w3s itself ? In my imagination that entity will be cooperating with w3s and other instances of the protocol deployment to ensure that data can be migrated in case one of them is out of business.

Another question that comes in my head is how can data be kept available in case of entity going out of business as there may not be other entity that offers service for the same price.

rfc/networked-dot-storage.md
```diff
@@ -0,0 +1,174 @@
...
+
+### Perpetuity
+
+1. Users should be able to store hot data in W3S that will be available for  retrieval until they stop paying for it. They should not have to worry about any single business entity holding their data disappearing. If it does, the data should go somewhere else without their needing to intervene.
+
+### Sensible defaults with customization
+
+1. Users should have the option to store with simply "the W3S network" rather than a specific business entity or Filecoin SP. This is a hard lesson learned from Filecoin that forcing users to choose from a marketplace where information on available options is hard to come by is a terrible user experience.
+
+2. Users should still have the option to choose a provider they like, or to move between providers.
```
Gozala, line 58: Do you mean filecoin storage provider here or something else ? I think interesting question in my head is do users make deals with w3s and w3s makes deals with SPs or do we want to allow users to make deals with SPs directly ? I recall requests for later on few occasions although I am not sure what the use cases were.

rfc/networked-dot-storage.md
```diff
@@ -0,0 +1,174 @@
...
+
+1. Users should be able to gather a detailed history of what has happened with their data when they put it on the W3S Network.
+
+2. A user paying for storage should be able to verify that data is being stored.
+
+3. A user paying for storage should be able to verify it can be retrieved quickly. 
+
+4. A user paying for multiple backups should be able to verify each backup is still functioning.
+
+5. If verification demonstrates that a particular provider is failing to provide a service that the user has paid for, the network should detect the failures and move their data automatically. (or alternatively, the user should have the ability to create a dispute and move to a different provider while getting money back)
```
Gozala, line 72: I think this should be **and** not **or**, if someone failed to uphold commitment user should be able to hold them accountable. In other words repair should take place but user still should be able to request compensation for service disruption if they so choose.

rfc/networked-dot-storage.md
Gozala: Here is a the hot take: Job of the router is to uphold user preferences and load balance supply and demand. Assuming that is accurate do we even need blockchain, could that simply be an IPVM task handling routing ? That way users gain a choice to pick the "router" that they find to be a right fit. In fact user with right technical skill could even fork the router to create one that fits their needs if one does not exist already.

rfc/networked-dot-storage.md
travis: Yeah honestly I think this should be a much more nuanced thing - imho different w3up providers should be able to offer different levels of "quality of service" on a bunch of different axes (latency, resilience, bandwidth, etc) as well different levels of legal/social guarantees (say, "we will comply with GDPR" or "we will never host data in Iran/China/Russia/USA/etc") - we could then craft different "default" user experiences for people that connected them with providers that met their needs, and once they are in the ecosystem they wouldn't necessarily need to maintain a relationship with us if they don't want to - I think this could go a long way toward balancing "we want to build a defensible business" and "we want to minimize our role as a central point of failure or point of centralization for our users" which imho are in dramatic tension for us...

rfc/networked-dot-storage.md
Gozala: How about alternative view of the repair service. Instead of having a single blockchain operated system, allow users to run repair service anywhere they like, IPVM, AWS, on user device in their closet or whatever else (in fact I think we could have interesting offering here). In a sense repair service become an agent hired by user which acts on user behalf.

Point I'm trying to make here is that coordination is both difficult and costly and I'm not sure it provides any real benefit to the user that needing assurance of service been operational. Also note that if repair service is just a code you can run it can probably also run as smart contract if users would want to do that.

Overall I'm for plurality of choice, as opposed to assumption that there is a universal solution that fits everyones needs.

rfc/networked-dot-storage.md
travis: what I'd like to see is:

1) we set up this network and serve as the default point of entry for people
2) we continually move to make it possible for people to decouple themselves from w3s-the-legal-entity and make this a core value proposition of our service - we will constantly need to balance this against the needs and desires of stakeholders in w3s-the-legal-entity so we'll need to be thoughtful about how we maintain this line
3) in the long run steady state, w3s-the-legal-entity strives to continually improve the "default experience" for engaging with w3s-the-network while simultaneously encouraging an ecosystem of businesses to grow around w3s-the-network - some of these businesses may be addressing market segments that we choose not to engage in, while others may directly compete with us - this is a feature and something we will ideally make a core value proposition of w3s-the-network

rfc/networked-dot-storage.md
travis: oh also I think this might even open up space for people to make decisions about storage providers based on real-world social factors, which is imho what is needed to support "private data" or PII - this is honestly the market segment I am most excited about

rfc/networked-dot-storage.md
Gozala: Can we just declare ourselves decentralized ? That way customers can tell us what they actually need or tell us how what we are doing is not decentralized enough which is effectively telling us what they need.

rfc/networked-dot-storage.md
```diff
@@ -0,0 +1,174 @@
...
+
+### What about quality of service?
+
+Web3.storage users love our reliability, ease of use, and performance. In the content addressed world, these are easily our strongest differentiators. As @alanshaw has pointed out several times, in general, we MUST maintain our level of service as we evolve the network. That is a north star. The world already has enough broken decentralized systems.
+
+### Does "distributed consensus and produce verifiable results" mean smart contracts on a blockchain?
+
+Most likely, yes. But I'm open to hearing about other consensus mechanisms. That said, for the specific problem of trustless consensus on limited execution of relatively small rule sets, honestly, I think blockchains are the most proven solution to date. We need to assess objectively the risks/benefits of a different approach.
```
Gozala, line 170: I think it might be worth teasing out what do we need consensus on ? Perhaps some things do not need consensus but could instead lean into giving user a choice ? I think it is not a stretch to describe  blockchains as extremely inefficient coordination system with some mutable state. Which makes me wonder if by consensus we mean coordination and if so perhaps what we actually need is to expose (mutable) state with some concurrency controls.

Gozala (commented):
Thank you for describing coherent picture & distilling concrete goals. I do like it, that said still think there is few misalignments which I called out in the comments and will capture in the short list below:

1. I still think one network is not a right direction and we should be thinking about plurality of networks that may overlap. I suspect multiple networks will arise regardless, however designing for that future may facilitate cooperation and interop as opposed to just competition.
2. I think it is worth considering and exploring whether we really need network level consensus. I would argue that reducing amount of consensus needed will greatly improve network resilience and adaptability. I also make a point that perhaps coordination should not be function of consensus, but rather a function of user preference.

rfc/networked-dot-storage.md
hannahhoward: I mean if only there were some sort of decentralized currency system based on a public database of transactions that wasn't controlled by any one entity that we could use.

rfc/networked-dot-storage.md
hannahhoward: re: 3

Yes, this how I think of it.

For me the end state of the w3s entity is some kind of foundation where a small part of protocol revenue pays a core team to keep developing the network.

The nice thing is once the token is liquid, if we end up with a token and a token raise, the VCs have already gotten their exit. This to me is the moment you can transition to a model that more closely resembles https://github.com/web3-storage/RFC.staging/pull/1 . This is all of course fantasy, based on everything going exactly how it's supposed to, not getting caught up in some kind of crypto bubble, but that actually is how I would like it to go.

rfc/networked-dot-storage.md
hannahhoward: Yea, I'm open to ideas here. The first task is just to build a router of any kind. If it will run in easily in a WASM environment, bonus, cause then there is a path to either IPC OR IPVM, once the right path becomes clear.

hannahhoward: So if someone wants to write the router in rust, I'm open to it. As long as it doesn't use rust async, which I think won't work well in either env.

rfc/networked-dot-storage.md
hannahhoward: Yea, I get that choice is your go to.

Personally, I feel that the vast majority of every day users are not going to want to get down and dirty on these details, which I think are pretty low level. Which is why I lean towards sensible defaults first. But I'm not opposed to choice.

To be clear though I think that:
1. Repair service is absolutely something there can be many of.
2. I really misspoke, the default repair service SHOULD simply be a smart contract.

rfc/networked-dot-storage.md
hannahhoward: I think users who just want to store with an SP directly should have that option.

I've come around to the fractal design concept -- an SP can run Filecoin.storage (or whatever we call Filecoin SP tailored w3up). If they do this, and they want to find their own users, they could just tell them to make direct data uploads to their instance of w3up, which would go in their own aggregation pipeline, and then eventually get put in their own deals for their Filecoin SP operation. and/or, they could connect their filecoin.storage instance to a web3.storage router, and get a portion of uploads going through us, which would likely just get aggregated along with any direct uploads they received.

hannahhoward: But seperately, you can also work directly with a provider running a hot storage instance with no Filecoin operation of their, who would then use a common aggregation pipeline (likely the one we run for a while).

rfc/networked-dot-storage.md
hannahhoward: for sure.

rfc/networked-dot-storage.md
alanshaw: > This is a nuance, but I can't help but wonder if making users choose number of unique copies is the right approach, my guess is users want some resilience, perhaps geographic distribution and other forms of assurances that data availability is adequate. Trying to figure out number of replicas seems several steps disconnected from those needs and also would remove some flexibility on the implementer side.

Yes, +1. In the same way that I do not care how many replicas AWS stores of the data I have in S3. It should be adequete, so that the data is not lost.

When we talk about geographical distribution, my suspicion is that we're mostly thinking about retrieving the data quickly, which is what the CDN side of the product should be oriented around, not the storage side.

PR #25, opened 2024-04-19
Implementation outline router.storage

hannahhoward:
[Preview](https://github.com/web3-storage/RFC/blob/rfc/router.storage/rfc/router-storage.md)

This attempts to specify router.storage operation based on the designs we discussed in federated MVP

I am not sure if I am getting the UCAN handoffs correct. The reinvoking of `blob/allocate` and `blob/accept` on the delegated node feels off... but also there doesn't seem to be an obvious way to not re-invoke -- because `blob/allocate` inherently involves a negotiation process. Maybe though it would be better to try to do all of `blob/allocate` synchronously though cause then `blob/allocate` and `blob/accept` could be direct calls on the chosen node? I struggling to get my delegations vs invocations correctly.

### DIFF
--- a/rfc/router-storage.md
+++ b/rfc/router-storage.md
@@ -1,0 +1,767 @@
+# Router.Storage
+
+## Authors
+
+- [hannah]
+
+## Abstract
+
+Proposes extensions to the [W3 Blob Protocol](https://github.com/web3-storage/specs/blob/main/w3-blob.md) to enable users to expess preferences about how their blobs are stored, and a suggested implementation for a W3 blob "router", a capability provider for `/space/content/add/blob` that would invoke `/service/blob/allocate` and `service/blob/accept` on different storage nodes, based on the users preferences
+
+## W3 Blob Protocol
+
+To enable users to express intent for how they would like data stored, an optional `storagePreferences` argument is added to the `space/content/add/blob` invocation:
+
+```ts
+type AddBlob = {
+  cmd: "/space/content/add/blob"
+  sub: SpaceDID
+  args: {
+    blob: Blob
+    storagePreferances?: StoragePreferances
+  }
+}
+
+// intentionally left blank for now to expand over time
+type StoragePreferences = {
+}
+
+type Blob = {
+  digest:   Multihash
+  size:     int
+}
+
+type Multihash = bytes
+type SpaceDID = string
+```
+
+Expecting that well supported preferences will evolve over time, for now we are not attempting to define specific preference properties.
+
+## Routed implementation of `space/content/add/blob`
+
+A blob router provides the `space/content/add/blob` capability by routing subsequent effects to seperate storage nodes providing `service/blob/allocate` and `service/blob/accept` capabilities.
+
+The blob router contains a state of storage nodes available.
+
+```ts
+type StorageNode = {
+  did: ProviderDID
+  w3upEndpoint: URL // a UCAN invocation endpoint providing at minimum `service/blob/allocate` and `service/blob/accept`
+  storageProperties?: StorageProperties
+}
+
+type StorageNodes = {
+  [ProviderDID]: StorageNode
+}
+
+// intentionally left blank for now to expand over time
+type StorageProperties = {
+}
+```
+
+The blob router will probably also need a state of where things have been placed at and what stage they're at:
+
+```ts
+type BlobAllocation = {
+  did: ProviderDid
+  blob: Blob
+}
+
+type BlobAllocations { [digest]: BlobAllocation }
+
+```
+
+For now, we don't define any protocols for adding and removing providers to a router. Moreover, expecting what we track about nodes will evolve over time, for now we are not attempting to define specific storage properties for now
+
+### `space/content/add/blob` execution flow
+
+#### `space/content/add/blob` invoked on space:
+
+```json
+{ // "/": "bafy..add"
+  "cmd": "/space/content/add/blob",
+  "sub": "did:key:zAlice",
+  "iss": "did:key:zAlice",
+  "aud": "did:web:web3.storage",
+  "args": {
+    "blob": {
+      // multihash of the blob as byte array
+      "digest": { "/": { "bytes": "mEi...sfKg" } },
+      // size of the blob in bytes
+      "size": 2_097_152,
+    },
+    "storagePreferences": {
+
+    }
+  }
+}
+```
+
+Legacy invocation of `blob/add`:
+
+```json
+{ // "/": "bafy..add"
+  
+  "iss": "did:key:zAlice",
+  "aud": "did:web:web3.storage",
+  "att": [
+    {
+      "cmd": "/space/content/add/blob",
+      "with": "did:key:zAlice",
+      "nb": {
+        "blob": {
+          // multihash of the blob as byte array
+          "digest": { "/": { "bytes": "mEi...sfKg" } },
+          // size of the blob in bytes
+          "size": 2_097_152,
+        },
+        "storagePreferences": {
+        }
+      }
+    }
+  ]
+}
+```
+
+Receipt:
+
+```json
+{ // "/": "bafy..work",
+  "iss": "did:web:web3.storage",
+  "aud": "did:key:zAlice",
+  "cmd": "/ucan/assert",
+  "sub": "did:web:web3.storage",
+  "args": {
+    "assert": [
+      // refers to the invocation from the example
+      { "/": "bafy..add" },
+      // refers to the receipt corresponding to the above invocation
+      {
+        "out": {
+          "ok": {
+            // result of the add is the content (location) commitment
+            // that is produced as result of "bafy..accept"
+            "site": {
+              "ucan/await": [
+                ".out.ok.site",
+                { "/": "bafy...accept" }
+              ]
+            }
+          }
+        },
+        // Previously `next` was known as `fx` instead, which is
+        // set of tasks to be scheduled.
+        "next": [
+          // 1. System attempts to allocate memory in user space for the blob.
+          { // "/": "bafy...alloc",
+            "cmd": "/service/blob/allocate",
+            "sub": "did:web:web3.storage",
+            "args": {
+              // space where memory is allocated
+              "space": "did:key:zAlice",
+              "blob": {
+                // multihash of the blob as byte array
+                "digest": { "/": { "bytes": "mEi...sfKg" } },
+                // size of the blob in bytes
+                "size": 2_097_152,
+              },
+              // task that caused this invocation
+              "cause": { "/": "bafy..add" }
+            }
+          },
+          // 2. System requests user agent (or anyone really) to upload the content
+          // corresponding to the blob
+          // via HTTP PUT to given location.
+          { // "/": "bafy...put",
+            "cmd": "/http/put",
+            "sub": "did:key:zMh...der", // <-- Ed299.. derived key from content multihash
+            "args": {
+              // pipe url from the allocation result
+              "url": {
+                  "ucan/await": [
+                    ".out.ok.address.url",
+                    { "/": "bafy...alloc" }
+                  ]
+              },
+              // pipe headers from the allocation result 
+              "headers": {
+                "ucan/await": [
+                  ".out.ok.address.headers",
+                  { "/": "bafy...alloc" }
+                ]
+              },
+              // body of the http request
+              "body": {
+                // multihash of the blob as byte array
+                "digest": { "/": { "bytes": "mEi...sfKg" } },
+                "size": 2_097_152
+              },
+            },
+            "meta": {
+              // archive of the principal keys
+              "keys": {
+                "did:key:zMh...der": { "/": "mEi...sfKg" } 
+              }
+            }
+          },
+          // 3. System will attempt to accept uploaded content that matches blob
+          // multihash and size.
+          { // "/": "bafy...accept",
+            "cmd": "/service/blob/accept",
+            "sub": "did:web:web3.storage",
+            "args": {
+              "space": "did:key:zAlice",
+              "blob": {
+                // multihash of the blob as byte array
+                "content": { "/": { "bytes": "mEi...sfKg" } },
+                "size": 2_097_152,
+              },
+              "expires": 1712903125,
+              // This task is blocked on allocation
+              "_put": { "ucan/await": [".out.ok", { "/": "bafy...put" }] }
+            }
+          }
+        ]
+      }
+    ]
+  }
+}
+```
+
+Legacy receipt for `blob/add` pre-UCAN 1.0:
+
+```json
+{ // "/": "bafy..work",
+  "iss": "did:web:web3.storage",
+  "aud": "did:key:zAlice",
+  "ran": {
+    "/": "bafy..add"
+  },
+  "out": {
+    "ok": {
+      // result of the add is the content (location) commitment
+      // that is produced as result of "bafy..accept"
+      "site": {
+        "ucan/await": [
+          ".out.ok.site",
+          { "/": "bafy...accept" }
+        ]
+      }
+    }
+  },
+        // Previously `next` was known as `fx` instead, which is
+        // set of tasks to be scheduled.
+  "fx": {
+    "fork": [
+
+      // 1. System attempts to allocate memory in user space for the blob.
+      { // "/": "bafy...alloc",
+        "iss": "did:web:web3.storage",
+        "aud": "did:web:web3.storage",
+        "att": [
+          {
+            "can": "/web3.storage/blob/allocate",
+            "with": "did:web:web3.storage",
+            "nb": {
+              // space where memory is allocated
+              "space": "did:key:zAlice",
+              "blob": {
+                // multihash of the blob as byte array
+                "digest": { "/": { "bytes": "mEi...sfKg" } },
+                // size of the blob in bytes
+                "size": 2_097_152,
+              },
+              // task that caused this invocation
+              "cause": { "/": "bafy..add" }
+            }
+          }
+        ]
+      },
+      // 2. System requests user agent (or anyone really) to upload the content
+      // corresponding to the blob
+      // via HTTP PUT to given location.
+      { // "/": "bafy...put",
+        "iss": "did:key:zAlice",
+        "aud": "did:key:zAlice",
+        "att": [
+          {
+            "can": "/http/put",
+            "with": "did:key:zMh...der", // <-- Ed299.. derived key from content multihash
+            "nb": {
+              // pipe url from the allocation result
+              "url": {
+                  "ucan/await": [
+                    ".out.ok.address.url",
+                    { "/": "bafy...alloc" }
+                  ]
+              },
+              // pipe headers from the allocation result 
+              "headers": {
+                "ucan/await": [
+                  ".out.ok.address.headers",
+                  { "/": "bafy...alloc" }
+                ]
+              },
+              // body of the http request
+              "body": {
+                // multihash of the blob as byte array
+                "digest": { "/": { "bytes": "mEi...sfKg" } },
+                "size": 2_097_152
+              },
+            },
+          }
+        ],
+        "fct": [
+          // archive of the principal keys
+          "keys": {
+            "did:key:zMh...der": { "/": "mEi...sfKg" } 
+          }
+        ]
+      },
+    ],
+  
+    // 3. System will attempt to accept uploaded content that matches blob
+    // multihash and size.
+    "join": { // "/": "bafy...accept",
+      "iss": "did:web:web3.storage",
+      "aud": "did:web:web3.storage",
+      "att": {
+        "can": "/web3.storage/blob/accept",
+        "with": "did:web:web3.storage",
+        "nb": {
+          "space": "did:key:zAlice",
+          "blob": {
+            // multihash of the blob as byte array
+            "content": { "/": { "bytes": "mEi...sfKg" } },
+            "size": 2_097_152,
+          },
+          "expires": 1712903125,
+          // This task is blocked on allocation
+          "_put": { "ucan/await": [".out.ok", { "/": "bafy...put" }] }
+        }
+      }
+    }
+  }
+}
+```
+
+Note that while this appears identical to for 1.0 Invocation receipts, a careful inspection of the legacy receipt suggests something slightly different has happened: unlike the current flow, the return for `blob/allocate` is NOT synchronous. 
+
+#### Router handling of `blob/allocate`:
+
+1. The router first checks for an existing allocation and shortcuts if it exists
+2. The router runs a choice function based on the storage preferences and list of nodes available:
+
+```ts
+type ChooseNode = (prefs: StoragePreferences, nodes: StorageNodes, alreadyFailed: StorageNode[]) => StorageNode
+```
+
+**TODO: should this be a choose one or a produce complete ordering? The `alreadyFailed` argument is intended to capture previous choices that failed, to remove them from the pool, on the theory choosing 1 is probably easier and more efficient**
+
+Such a function MUST be deterministic from the input nodes and preferences. Eventually we imagine a space having a configured CID that represents a custom implementation of such a function.
+
+Let's say the chosen storage node is:
+
+```json
+{
+  "did": "did:web:node1.storage",
+  "w3upEndpoint": "https://node1.storage"
+}
+```
+
+3. The service invokes `blob/allocate` remotely on the chosen node:
+
+```json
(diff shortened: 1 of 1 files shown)


rfc/router-storage.md
```diff
@@ -0,0 +1,767 @@
...
+
+```json
+{ // "/": "bafy..add"
+  "aud": "did:web:node1.storage",
+  "iss": "did:web:web3.storage",
+  "att": [
+    {
+      "can": "/web3.storage/blob/allocate",
+      "nb": "did:web:node1.storage",
+      "with": {
```
Gozala, line 403:
```suggestion
      "with": "did:web:node1.storage",
      "nb": {
```

rfc/router-storage.md
```diff
@@ -0,0 +1,767 @@
...
+
+Receipt:
+
+```json
+{ 
+  // "/": "bafy..work",
+  "iss": "did:web:node1.storage",
+  "aud": "did:web:web3.storage",
+  "cmd": "/ucan/assert",
+  "sub": "did:web:web3.storage",
```
Gozala, line 426: Storage node `did:web:node1.storage` will not have an authority to sign things on `did:web:web3.storage`  sub(ject) unless explicitly authorized.  I also don't think we should authorize storage nodes to issue receipts on our behalf, instead they should issue receipts on their own sub(ject) and we can decide what to do with it.


Also in 1.0 we're going to make `iss == aud` on receipts so they form a loop and effectively provides a way to distribute state. You can think of it as I can send receipt back to the sub(ject) telling it to assert state as per receipt.

```suggestion
  "iss": "did:web:node1.storage",
  "aud": "did:web:node1.storage",
  "sub": "did:web:node1.storage",
  "cmd": "/ucan/assert",
```

rfc/router-storage.md
```diff
@@ -0,0 +1,767 @@
...
+}
+```
+
+Legacy receipt:
+
+```json
+{ 
+  // "/": "bafy..work",
+  "iss": "did:web:node1.storage",
+  "aud": "did:web:web3.storage",
```
Gozala, line 458: Old receipts have no audience.

```suggestion
```

rfc/router-storage.md
```diff
@@ -0,0 +1,767 @@
...
+        "headers": {},
+        "expires": 1712903125
+      }
+    }
+  },
+  "fx": [],
+}
+```
+
+4. On success, router creates a success receipt for the original user
```
Gozala, line 476: I'm not sure we need this, it does not matter who did the allocation, we just want user to upload there and issue their own receipt when they're done. If there is reason I'm missing, it's probably best to add a segment elaborating it.

Gozala: I would also expect this state to be added into the `BlobAllocations` state.

rfc/router-storage.md
```diff
@@ -0,0 +1,767 @@
...
+    "digest": { "/": { "bytes": "mEi...sfKg" } },
+    // size of the blob in bytes
+    "size": 2_097_152,
+  },
+}
+```
+
+4. On fail, router adds the failed nodes to a list of "alreadyFailed", and returns to step 2.
+
+5. If all nodes fail, return a fail receipt to the user for `blob/allocate`
```
Gozala, line 551: I think what I'm failing to understand if we expect to get `n` allocations or if we just trying to get a single allocation.

rfc/router-storage.md
```diff
@@ -0,0 +1,767 @@
...
+
+#### Router handling of blob accept
+
+After the user puts their blob to the presigned URL, they signal to the router that `blob/accept` can proceed by sending them the UCAN receipt for `http/put`(currently this is handled differently between legacy and 1.0 UCAN)
```
Gozala, line 555: Is there a reason why do we need to route these as opposed to letting user deal directly with the storage node that allocated and let them perform their own `accept` ? I would expect that `blob/add` simply describes coordination with storage nodes by forking set of `allocation → put → accept` tasks flows and then combines their results into an output.

Gozala: I'm not going to comment on this section of the spec as I think we probably should not try to route `blob/accept` but rather combine outputs from storage nodes. It also would remove dependency on local state that otherwise requires side-effects and can be racy.

rfc/router-storage.md
```diff
@@ -0,0 +1,767 @@
...
+      "site": "bafy..locationCommitment"
+    }
+  },
+  "fx": [],
+}
+```
+
+4. Router fetches location commitment, and creates new commitment for the user.
+
+**Note: initially I thought this was a redelegation, but I think it's maybe not -- the commitment here is to a limited download URL of JUST the blob -- it's not a w3s gateway url with full download capabilities.
```
Gozala, line 665: I do think it should be re-delegation, which BTW can impose extra restrictions. I think it should be re-delegation because it would enable us to track reads.

I think gateway URLs are separate deal, specifically I think user should re-delegate location commitment to the gateway to authorize them to serve content through gateway because they will be billed for each read gateway will utilize. That also would enable user to revoke such delegation to stop gateway from serving said content at their expense.

rfc/router-storage.md
```diff
@@ -0,0 +1,767 @@
...
+
+the "did:web:web3.storage" is maybe really a did for two seperate things:
+- the web3.storage read pipeline
+- the web3.storage write pipeline
+
+As long as we're the only ones running the read pipeline, it can all go through w3s.link, but I imagine a future situation where other people want to use another read pipeline provider.
```
Gozala, line 671: Exactly, which is why I think user can delegate location commitment to say `did:web:w3s.link` authorizing it to serve this content at their expense, but they also could go and authorize other gateways or libp2p nodes for different read pipelines.

rfc/router-storage.md
```diff
@@ -0,0 +1,767 @@
...
+
+We also need to start thinking about the difference between a storage link -- i.e. a place to just download a CAR file as is with range parameters, and a gateway endpoint -- like w3s.link that serves flat files, pathing, etc -- even in an "encodingless IPFS" world, w3s.link is different that it's CDN cached and highly available **
+
+Original:
+
+```json
+{
+  "iss": "did:web:node1.storage",
+  "aud": "did:web:web3.storage",
```
Gozala, line 680: Makes me wonder if we actually want to be in the middle of the authorization chain or whether storage nodes should delegate to the space directly. I imagine we may want different choice here for different use cases. If we are in the middle we are taking on some custody but could potentially reuse stored blob (e.g. store once but charge multiple users) but if things fail we're accountable. If we aren't in the middle than it is all between user (space) and the storage node itself, no deduplication opportunity here and if things fail it's storage node who is accountable to the user (space).

rfc/router-storage.md
```diff
@@ -0,0 +1,767 @@
...
+    // multihash must match be for the blob uploaded
+    ["==", ".content", { "/": { "bytes": "mEi...sfKg" } }],
+    // must be available from this url -- but for a storage node this is probably just a raw blob, not w3s.link with full gateway support
+    ["==", ".url", "https://node1.storage/raw/bafy...7fi"],
+    // from this range
+    ["==", ".range[0]", 0],
+    ["==", ".range[1]", 2_097_152],
+  ],
+  // does not expire
+  "exp": null
```
Gozala, line 693: I think if we are in the middle we could issue non-expiring commitments and take on the task of renewing deals with storage nodes to uphold this commitment. On the other hand if we aren't in the middle then commitment really should have a expiry.

Gozala (commented):
Thanks @hannahhoward for putting it together I think this is a great first draft, although we probably need to iterate bit more in regards to selection of storage nodes.

Also for what it's worth I have been sketching rules syntax for kind of tasks that simply coordinate invocations https://observablehq.com/@gozala/invocation-rules-syntax I think it might be a good idea to describe coordination somewhat along those lines that way all of it could be compiled into `next` field and we'd only need to implement some generic operators to do candidate selection. I'll try to take stab at it and share here.

vasco-santos (commented):
This is interesting, as it goes a bit on a different direction that I have been thinking. I would like to call out how I was thinking this interactions would look like and would be great to at least compare trade offs.

I would expect `web3.storage` did to always be the audience of `blob/allocate` and `blob/accept` (this also allows the client to derive task CID of them, in order to query receipts given they would have all the information).

The core difference in my mind, would be that `web3.storage` would rely on such kind of router when running `allocate` task with a new set of capabilities, let's consider a flow:

1. client invokes `blob/add` to be executed by `did:web:web3.storage`
2. client derives `blob/allocate` and `blob/accept` task CIDs and subscribes for receipts available for them
3. `did:web:web3.storage` schedules allocate task to run on their schedule
4. once `blob/allocate` runs, it would invoke a `storage-node/offer` to a router principal. 
5. router principal would discover available hot storage nodes on the network to store content based on the provided `StorageConfiguration`
6. `blob/allocate` finishes when it can find a receipt for `storage-node/offer` with the write target address for the client to use
7. client writes the bytes
8. client issues receipt for `http/put`
9. `did:web:web3.storage` verifies bytes are there and runs `blob/accept`

As previously mentioned, I think the core advantage here is that client has all the information to derive taskCID for the pending tasks service will run. If `blob/allocate` and `blob/accept` become dynamic, that would not happen and we can still put on `blob/allocate` effects for the `storage-node/offer` + `storage-node/accept` receipts so that user can follow entire receipt chain. Moreover, this provides a much easier avenue to introduce the billing/usage capabilities, as well as metrics tracking.

On the billing side of things, it is not clear to me on this approach how would we have a billing related capability tied with allocation if allocation runs on a third party.

Gozala:
> I would expect `web3.storage` did to always be the audience of `blob/allocate` and `blob/accept` (this also allows the client to derive task CID of them, in order to query receipts given they would have all the information).

Please note that per my sketch all of those will be in the `next` field so client will not need to derive it themself, but they would be able to query by receipt.

Gozala:
> The core difference in my mind, would be that `web3.storage` would rely on such kind of router when running `allocate` task with a new set of capabilities, let's consider a flow:

I think in your version routing logic is made opaque, meaning it is implementation detail not exposed to the caller nor it provides trail of how candidate selection occured or if system had to backtrack and try different candidates.

There are different tradeoffs at play, making that opaque means router gets far more flexibility to make choices or to change behavior is it see fit. It is probably simpler to implement and faster to iterate over.

Alternative version exposes all the node decision making process and attempt to tie it with preferences on the space. It will produce a lot more complete trail of steps and decisions made and potentially provide more granular status updates along the way. 

Now that I think about it I imagine `candidate/choose` would probably be a capability user can install on their space so that it can reflect their preference. Obviously we would have a default in place kind of like `auto` region in CF.

> On the billing side of things, it is not clear to me on this approach how would we have a billing related capability tied with allocation if allocation runs on a third party.

I think billing commands would be part of the sketch workflow, specifically we would issue billing command when allocation succeeds with the duration and blob size, but I think we need to spec out new billing interface before we can do that

vasco-santos:
> There are different tradeoffs at play, making that opaque means router gets far more flexibility to make choices or to change behavior is it see fit. It is probably simpler to implement and faster to iterate over.

Yes, that is why I think this RFC should include them and we should see what makes sense. For instance, in w3filecoin makes whole sense you can follow the whole flow because it also MAY take a long time. However, here we should be optimising for all this flow to be fast, and therefore I am looking to understand what would be the advantages of opting for a full receipt chain solution. There may exist merits on it, but I am until now failing to understand them.

> Alternative version exposes all the node decision making process and attempt to tie it with preferences on the space. It will produce a lot more complete trail of steps and decisions made and potentially provide more granular status updates along the way.

> Now that I think about it I imagine candidate/choose would probably be a capability user can install on their space so that it can reflect their preference. Obviously we would have a default in place kind of like auto region in CF.

Here I think we can agree, as well as on making something with space specific characteristics.

For instance, billing capabilities will probably be opaque, unless we need some kind of on chain primitive that make it required to show a proof. Here, I think it MAY actually make sense to treat this as 2 different systems, with their own receipt chain that you can track. So `web3.storage` being a kind of a client of `router.storage` like a user is a client of `web3.storage`. So, on `blob/allocate`, I would expect a well known invocation to discover a memory to allocate to a audience called `router.storage`. `router.storage` receipt could then have a CID where it links to all the "internal" receipt chain that was made.

Gozala:
> However, here we should be optimising for all this flow to be fast, and therefore I am looking to understand what would be the advantages of opting for a full receipt chain solution. There may exist merits on it, but I am until now failing to understand them.

My main arguments in favor are

1. Transparency - All the decisions are recorded and exposed to users for transparency.
2. User Agency - I don't think one size fits all, that is different users may have different preferences in regards to who they want to store data with. By making process transparent and making decision making process programmable users gain authority to drive decisions.
   - ℹ️ I don't believe you can achieve 2nd without 1st, unless provenance of decision is captured in signed chain there is no way of knowing how did system chose specific storage node and whether it did really reflected user preferences
3. System introspection - Unless we log all the steps it's really hard to introspect what has occured. We could use conventional logging but I think signed log has advantages above and I believe overhead should be negligible.
4. Optimization opportunities - Each step is a computation on data, if inputs are the same we can reuse outputs and avoid recomputing same thing like choosing a best candidate storage node.

Gozala:
> For instance, billing capabilities will probably be opaque, unless we need some kind of on chain primitive that make it required to show a proof. 

I don't think it should be opaque, I think it would be really nice if receipt chain captured precisely when we charge user how much and why. Maybe we mean different things by what it means to make it opaque. But I would system to have something like

```js
{ 
  cmd: '/balance/add',
  sub: 'did:web:web3.storage',
  args: {
    consumer: 'did:key:zAlice',
    product: 'storage',
    amount: 128,
    cause: { '/': 'bafy..inv' },
  }
}
```

Here, I think it MAY actually make sense to treat this as 2 different systems, with their own receipt chain that you can track. So `web3.storage` being a kind of a client of `router.storage` like a user is a client of `web3.storage`. So, on `blob/allocate`, I would expect a well known invocation to discover a memory to allocate to a audience called `router.storage`. `router.storage` receipt could then have a CID where it links to all the "internal" receipt chain that was made.

Yeah so it boils down to whether things should be encapsulated or exposed. If you expose internal chains outside there is no real difference especially if progress tracking can be built into the low level library (like ucanto). If we do want to hide internal details that is different matter, but I personally do not think we should more transparency make it easier to introspect, builds trust and is likely more welcome by w3 community.

It is also worth calling out that I think you are making an assumption that there is a single allocation per blob, however if were to support with higher replication factor that may not be the case. In such scenario we would probably want to allocate memory in several storage nodes and now that becomes a lot more difficult with more opaque system as are no longer able to parallelize allocations and uploads, you'd have to allocate `n` locations and then upload to them as opposed to doing things concurrently. Or alternatively you'd have to allocate one memory location upload and then replicate it in other nodes, but now there is added costs that user could cut if they were to upload in each memory location.

rfc/router-storage.md
```diff
@@ -0,0 +1,767 @@
...
+            }
+          },
+          // 3. System will attempt to accept uploaded content that matches blob
+          // multihash and size.
+          { // "/": "bafy...accept",
+            "cmd": "/service/blob/accept",
+            "sub": "did:web:web3.storage",
+            "args": {
+              "space": "did:key:zAlice",
+              "blob": {
```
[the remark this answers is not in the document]
hannahhoward: note space is included here?

rfc/router-storage.md
```diff
@@ -0,0 +1,767 @@
...
+          "keys": {
+            "did:key:zMh...der": { "/": "mEi...sfKg" } 
+          }
+        ]
+      },
+    ],
+  
+    // 3. System will attempt to accept uploaded content that matches blob
+    // multihash and size.
+    "join": { // "/": "bafy...accept",
```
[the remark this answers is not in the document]
hannahhoward: I think the current implementation does do a join though I'm not sure -- just reading @vasco-santos code

rfc/router-storage.md
```diff
@@ -0,0 +1,767 @@
...
+            }
+          }
+        },
+        // Previously `next` was known as `fx` instead, which is
+        // set of tasks to be scheduled.
+        "next": [
+          // 1. System attempts to allocate memory in user space for the blob.
+          { // "/": "bafy...alloc",
+            "cmd": "/service/blob/allocate",
+            "sub": "did:web:web3.storage",
```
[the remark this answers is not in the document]
hannahhoward: ok, sorry, stop rewind need to check for understanding here.

The fundamental problem is that while we are delegating a /service/blob/allocate to storage node, we may have to try multiple nodes first in order to succeed. The proposed architecture treats that negotiation as an asynchronous process, with a higher level "wrapper" version of /service/blob/allocate running on the router, and this being returned in the invocation chain rather than an invocation on a specific node.

An alternate approach is to do that entire negotiation synchronously, and simply discard any failures in the invocation chain, returning the one that succeeded. I think this is much simple, but it does hide the failed invocations and negotiation from the user. More importantly, it could take a while potentially, depending on node acceptance rate and time to respond.

Given your suggestion here, I believe that you are assuming an approach closer this alternative -- can I confirm that is what you are thinking.

rfc/router-storage.md
```diff
@@ -0,0 +1,767 @@
...
+  },
+        // Previously `next` was known as `fx` instead, which is
+        // set of tasks to be scheduled.
+  "fx": {
+    "fork": [
+
+      // 1. System attempts to allocate memory in user space for the blob.
+      { // "/": "bafy...alloc",
+        "iss": "did:web:web3.storage",
+        "aud": "did:web:web3.storage",
```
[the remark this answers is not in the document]
hannahhoward: See comment above

rfc/router-storage.md
```diff
@@ -0,0 +1,767 @@
...
+        "att": [
+          {
+            "can": "/web3.storage/blob/allocate",
+            "with": "did:web:web3.storage",
```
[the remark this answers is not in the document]
hannahhoward: see lengthy comment above

rfc/router-storage.md
hannahhoward: blob allocate is included

rfc/router-storage.md
hannahhoward: this is a loop until success operation.

rfc/router-storage.md
hannahhoward: how does the router know about the resulting location commitment then?

arguably I guess instead of keeping a table of blob allocations, it could just query our indexing system.

rfc/router-storage.md
hannahhoward: makes sense

hannahhoward (commented):
Ok @Gozala 

I think the central question I'm struggling with is how much does traffic go through the router to the storage nodes vs the router simply selects nodes and let's the client communicate directly.

In my design, everything goes through the router, with the usage of nodes largely an implementation detail.

I gather from a number of your comments you feel that the client should quickly start communicate with nodes directly, with the routers central task to produce a priority list of nodes.

So like for `/blob/select/store`, if it produces a bunch of `blob/allocates` do you imagine the client trying them one by one? In the current assumption, the router communicates with them one by one till it succeeds, albeit asynchronously and transparently to the client. How would all this get represented in terms of UCAN?

I think a good next step is a discussion between you, myself, and Alan -- I'll try to find time on calendars for that. I do not want this design to spiral out too much, but also I'm pretty sure there are weaknesses in the way it's specified.

For me the most ideal design is no change to the client. For this, I think the right approach is to just make all of the logic of attempting allocations a part of the router, and also synchronous (change from current design), with the client simply receiving the same ucan chain it does currently, which it can use identical logic to process.

rfc/router-storage.md
```diff
@@ -0,0 +1,767 @@
...
+  storageProperties?: StorageProperties
+}
+
+type StorageNodes = {
+  [ProviderDID]: StorageNode
+}
+
+// intentionally left blank for now to expand over time
+type StorageProperties = {
+}
```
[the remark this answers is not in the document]
Gozala: So what we have coming in UCAN 1.0 and something I already have implemented (but not shipped) is UCAN policy system which is roughly equivalent of WHERE clause in SQL, so when I say `StorageCapibaliites` I imply something like

```ts
{ cmd: '/memory/allocate',
  with: did:...,
  where: [
    ["<=", ".blob.size", 1024]], // size of the blob should be < 1024
    ["some", ["asia", "europe"], [= ".region" "."]] // reigion should be aisa or europe
  ]
}
```


As mentioned benefit of using UCAN policy system would be that it would be well defined and will not require us to roll out custom code for figuring out which node providers satisfy desired criteria as we would be able to reuse most logic. It will also give us framework for expressing properties, which will be limiting but in my experience good guardrails to have.

rfc/router-storage.md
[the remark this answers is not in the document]
Gozala: > The fundamental problem is that while we are delegating a /service/blob/allocate to storage node

Other way round storage node delegates ability to invoke that to the router, that way router can allocate space on any registered storage node as needed.

> Given your suggestion here, I believe that you are assuming an approach closer this alternative -- can I confirm that is what you are thinking.

I am assuming async not the sync, not sure which one you were referring to as an alternative. That said I think both options should be possible if we get a delegation from the storage node to perform an allocation, whether we expose decision making details from router to user or not is something that can be decided separately.

rfc/router-storage.md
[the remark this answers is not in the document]
Gozala: I know but that is kind of what I was flagging, currently we would not have a space there as we'd need a unique accept for each attempt to add, instead we make both put and accept space agnostic that way it does not matter who provided content. If someone attempts to add content we already have we just supply them with receipts for http/put and blob/accept that we already have.

rfc/router-storage.md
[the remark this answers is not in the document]
Gozala: I'm not sure I understand what you're trying to imply, however if allocation is done by a storage node, which is my understanding, audience should be a storage node.

rfc/router-storage.md
Gozala: Sounds like a single success, correct ?

rfc/router-storage.md
Gozala: Why does it need to know of it ? If it needs to we can add a step in the pipeline to that effect. That said I don't think router needs to know or care about it, if user wants to utilize received commitment it would have to authorize gateway by delegating commitment effectively giving it permission to perform reads at user expense. If it never shared with gateway or libp2p node user does not wish to share read permission and that seems very reasonable choice IMO.

Gozala:
> Ok @Gozala
> 
> I think the central question I'm struggling with is how much does traffic go through the router to the storage nodes vs the router simply selects nodes and let's the client communicate directly.
> 
> In my design, everything goes through the router, with the usage of nodes largely an implementation detail.
> 
> I gather from a number of your comments you feel that the client should quickly start communicate with nodes directly, with the routers central task to produce a priority list of nodes.

Correct. I think we should get out of the way as much as possible as it costs less and makes role of router less significant.

> 
> So like for `/blob/select/store`, if it produces a bunch of `blob/allocates` do you imagine the client trying them one by one? In the current assumption, the router communicates with them one by one till it succeeds, albeit asynchronously and transparently to the client. How would all this get represented in terms of UCAN?
> 

What I'm suggesting is to work in tandem with a client in the process of node selection, once candidate is selected let client interact with it if fails authorize another allocation and so on until successful candidate storage node is found. I attempted to illustrate that in https://github.com/w3s-project/RFC/pull/25#issuecomment-2070641427

Note that I do not expect that user will send allocation requests to the storage nodes as they will not have an authorization to do so. Instead we just describe a flow of tasks where we try an allocation on a candidate node 
if succeeded get out of the way letting user provide content and receive the commitment, if failed run another task that attempts to choose another candidate and allocate there and recur until all fail or we succeed.

The code I've put together in the linked comment is effectively an illustration of UCAN task coordination system a.k.a promise pipelines (from invocation spec) just utilizing s-expressions. I also demo-ed it last week where I went into more details. General idea is that UCAN scheduler can take care of coordinating all those tasks given that primitive ones are provided via regular handlers.

> I think a good next step is a discussion between you, myself, and Alan -- I'll try to find time on calendars for that. I do not want this design to spiral out too much, but also I'm pretty sure there are weaknesses in the way it's specified.

👍

> 
> For me the most ideal design is no change to the client. For this, I think the right approach is to just make all of the logic of attempting allocations a part of the router, and also synchronous (change from current design), with the client simply receiving the same ucan chain it does currently, which it can use identical logic to process.

I think we're on the same page. It's more about how much we want to roll out as hand written code hiding details vs expose primitives as UCAN capabilities and describe decision making process a.k.a routing through task coordination (a.k.a promise pipelines)

PR #27, opened 2024-05-10
RFC: how we pick programming languages in W3S

hannahhoward:
[Preview](https://github.com/w3s-project/RFC/blob/feat/languages/rfc/languages-in-w3s.md)

This is a proposed process for thinking about what programming languages we use in W3S. 

Of note: there are no immediate proposed changes to the list of programming languages we use in W3S. Rather this is a framework for ongoing discussions and future language changes.

### DIFF
--- a/rfc/languages-in-w3s.md
+++ b/rfc/languages-in-w3s.md
@@ -34,15 +34,15 @@ As a project, we should add languages to W3S when we can do things with them tha
 
 - The single best reason to write Javascript is if you think that now or in the future you might run your code in a web browser. In our case, we know this is true for the w3up client.
 - Modern javascript is a highly capable language, if you unify a your team on a single coding style, linter, tool chain, package manager, approach to typing, and transpilation system as needed. Unfortunately, almost every one of the above has multiple options and the ecosystem is an incoherent mess.
-- Javascripts single threaded asynchronous programming model turns out to actually be pretty decent for writing web servers too.
-- A lot of devs are used to deploying code on centralized services with built in javascript/node support and forget that distributing binaries is still the best way to get code to machines in a truly distributed system like the one we're building. Javascript's binary story kind of sucks. And, while JS runtimes are pretty optimized, JS is still pretty unpredictable in terms of memory usage, and to a lesser extent performance. Javascript, at least in its node.js/browser form, also has a pretty big run time.
-- Every hiring manager loves javascript cause it's easy to hire javascript developers. I think this is a bit overrated, cause really there are so many kinds of javascript development and ways to write javascript that hiring a javascript developer is not a gaurantee they'll be able to write javascript the way your team uses it.
+- Javascript's single threaded asynchronous programming model turns out to actually be pretty decent for writing web servers too.
+- A lot of devs are used to deploying code on centralized services with built in javascript/node support and forget that distributing binaries is still the best way to get code to machines in a truly distributed system like the one we're building. Javascript's binary story kind of sucks. And, while JS runtime(s) are pretty optimized, JS is still pretty unpredictable in terms of memory usage, and to a lesser extent performance. Javascript, at least in its node.js/browser form, also has a pretty big size runtime(s).
+- Every hiring manager loves javascript cause it's easy to hire javascript developers. The main problem with this approach is there are so many kinds of javascript development and ways to write javascript that hiring a javascript developer is not a gaurantee they'll be able to write javascript the way your team uses it.
 
 #### Some Notes On Our Javascript
 
-- Thankfully this team has unified on a coding style, linter, tool chain, package manager, and approach to typing.
-- We have not however documented why we made the choices we did. We should do so ASAP. In particular types in JSDoc style comments and why we do it needs documentation. This includes our sporadic dropping into Typescript and why we do it. (For example: 'why is there an api.js and an api.ts and the js file just says `export {}` -- the answer is 'cause modules' but is EXTREMELY node clear)
-- Moreover, the w3up repo is huge, and honestly, even if you know javascript, there's a lot to get up to speed on. We should have a guide to the code.
+- Thankfully this team has unified on a coding style, linter, tool chain, package manager, and approach to typing. +1 on that.
+- We have not however documented why we made the choices we did. We should do so ASAP. In particular types in JSDoc style comments and why we do it needs documentation. This includes our sporadic dropping into Typescript and why we do it. (For example: 'why is there an api.js and an api.ts and the js file just says `export {}` -- the answer is 'cause modules' but is EXTREMELY not clear)
+- The w3up repo is huge, and honestly, even if you know javascript, there's a lot to get up to speed on. We should have a guide to the code.
 
 ### Go
 
@@ -50,7 +50,7 @@ As a project, we should add languages to W3S when we can do things with them tha
 - Go has an extremely powerful concurrency model with green threads
 - These two components together make Go one of the most powerful languages for writing servers of any kind. The challenge of server programming is not super fast compute, but super high concurrency execution. Go excels at this, which matters cause our project has a lot of servers involved.
 - Go's ability to do binary distributions by default is a major plus for kind of project.
-- By and large, people get stuff done quickly in Go because Go is extremely simple. It's fast to learn, easy to read, and it all looks the same. The only problem is its extremely boring. (Note from Hannah: When I first started writing Go I hated it. I thought someone took the joy out of programming. Now I love it cause I get my joy from what I'm building, not my programming language)
+- By and large, people get stuff done quickly in Go because Go is extremely simple. It's very fast to learn, easy to read, and it all looks the same. The only problem is its extremely boring. (Note from Hannah: When I first started writing Go I hated it. I thought someone took the joy out of programming. Now I love it cause I get my joy from what I'm building, not my programming language)
 - Lots and lots of Filecoin tooling and software is written in Go. In fact, W3S already has Go in it, we just don't maintain it -- it's called Spade
 - Garbage collection and green threads present challenges for using Go over an FFI, or for compiling go to WebAssembly, though TinyGo has improved this story by reducing the runtime size.
 
@@ -60,7 +60,7 @@ Caveat from Hannah: I am not a highly experienced Rust dev
 
 - Rust is a better C/C++. Like definitely much better. This is huge cause no one wrote a better C/C++ for low level programming for 20 years.
 - However, by and large, you should probably not write programs in Rust unless you'd consider writing them in C/C++. There is minimal evidence one can be as productive in Rust as higher level languages.
-- In particular, you should probably not write web servers in Rust or any other asynchronous, highly concurrent programs unless you have extra time on your hands which we don't.
+- In particular, you should probably not write web servers in Rust or any other asynchronous programs unless you have extra time on your hands which we don't.
 - BTW, if you haven't programmed in C/C++ recently, you've probably forgotten that long compile times are a thing on large projects in low level languages. 
 - Rust is essential for highly performant synchronous compute tasks. That's why people use it for smart contracts, hashing, and proofs.
 - Rust has the best WebAssembly story of any language.
--- a/rfc/languages-in-w3s.md
+++ b/rfc/languages-in-w3s.md
@@ -1,0 +1,84 @@
+# Programming languages in W3S
+
+...or wouldn't this all be easier if we rewrote it in Cobol...
+
+## Authors
+
+- [Hannah Howard]
+
+## Some basics
+
+- Currently, W3S is written almost entirely in Javascript, therefore, every dev on W3S must be able to write code in Javascript
+- Javascript is not a language that is well suited for every function in W3S, particularly as we evolve to a more distributed network running on bare metal hardware, and potentially do blockchain programming
+- Every additional language we add potentially adds team fragmentation. Therefore:
+  - We should not add languages because we "like" them, or get excited by them, or cause they seem #SoHotRightNow
+  - There must be a strong engineering driven reason to adopt a new language.
+- Not every dev will be an expert in all our languages. But once we adopt a language, every dev must be willing to do work in that language if it's required. We support each other and teach each other to make work get done faster.
+- Candidate languages worth considering adding right now are Go and Rust. In the future we will probably need to write some Solidity contracts.
+
+## Thoughts on programming languages from Hannah
+
+My baseline philosophy is no one should be super attached to the language they write code in. Moreover, I think everyone gets better as a developer the more experiences they get across programming languages. When I hire, I generally try to hire smart people who can reason through problems and trust them to learn what they need. I tend to think of the core craft of programming as being relatively static.
+
+One thing I miss about in office work (one of the only things) is that working in person gives people more opportunities to teach each other through pairing and knowledge sharing. This is a super fast way to learn languages. If you find yourself on a ticket in a language you're less familiar with, I highly recommend pairing with a language expert. Seriously, ask for people's time to get fast rather than staying slow. And in the age of ChatGPT, even if you can't find a pair you can probably get functioning if need be. 
+
+In my opinion, the actual reasons to choose a programming language almost always have nothing to do with the language syntax itself, or even the language's core philosophy about how to write code (if it has one). In fact, choosing a language for these reasons is probably almost an anti-pattern in the workplace.
+
+Instead the best reasons to use a language come from unique capabilities unlocked by the compiler and toolchain, the runtime, the standard library, or the surrounding ecosystem. IMHO the selling point of Java is the JVM, not the Java language. The selling point of Erlang is how its scheduling model and message passing architecture enable reliable execution of highly concurrent distributed real time applications (i.e. chat applications like WhatsApp). 
+
+As a project, we should add languages to W3S when we can do things with them that we either can't do or can't do very well in the languages we already write code with. That's the only reason to do so.
+
+## Language analysis from a business and engineering perspective
+
+### Javascript
+
+- The single best reason to write Javascript is if you think that now or in the future you might run your code in a web browser. In our case, we know this is true for the w3up client.
+- Modern javascript is a highly capable language, if you unify a your team on a single coding style, linter, tool chain, package manager, approach to typing, and transpilation system as needed. Unfortunately, almost every one of the above has multiple options and the ecosystem is an incoherent mess.
+- Javascripts single threaded asynchronous programming model turns out to actually be pretty decent for writing web servers too.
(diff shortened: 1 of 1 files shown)


rfc/languages-in-w3s.md
```diff
@@ -0,0 +1,84 @@
...
+- The w3up repo is huge, and honestly, even if you know javascript, there's a lot to get up to speed on. We should have a guide to the code.
+
+### Go
+
+- Go is a highly performant garbage collected language like Java, but is much simpler than Java
+- Go has an extremely powerful concurrency model with green threads
+- These two components together make Go one of the most powerful languages for writing servers of any kind. The challenge of server programming is not super fast compute, but super high concurrency execution. Go excels at this, which matters cause our project has a lot of servers involved.
+- Go's ability to do binary distributions by default is a major plus for kind of project.
+- By and large, people get stuff done quickly in Go because Go is extremely simple. It's very fast to learn, easy to read, and it all looks the same. The only problem is its extremely boring. (Note from Hannah: When I first started writing Go I hated it. I thought someone took the joy out of programming. Now I love it cause I get my joy from what I'm building, not my programming language)
```
gammazero, line 53: > only problem is its extremely boring

I have found Go to be one of the most interesting languages from my first use, and up to the present. It is interesting because it lets you do everything you want efficiently and elegantly, and _still_ stays out of your way. It has a nearly perfect balance, offering what is most needed and useful but nothing more to complicate things. I have not yet encountered another language that gets the balance so close to just right for most programming projects, particularly, those involving network and systems programming.

You do not need to choose a package manager, transpilation tools, or even a coding style - those are not interesting problems to solve. There is no requirement for your target execution environment to have the correct version of a particular runtime and dependencies. You can focus on solving concurrency problems without having to solve the system-dependent problem of parallelism (managing OS threads running across multiple cores). Go is interesting (and fun) because it avoids the annoying ceremony of how to go about doing things, and lets you just do things, right away.

... but, as previously stated, the engineering problem to solve is primarily what must determine which tools are best suited to solve it.

rfc/languages-in-w3s.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Proposing a language to adopt
+
+- The best time to propose a language to adopt is at the start of a new project
+- Present a proposal to write a project in a new language via an RFC
+   - Try to objectively present business and engineering tradeoffs of writing the project in a language we already support vs writing it in a new language
+   - Identify gaps in language support or tooling related to our software stack and present a plan to address it
+   - Ideally, make a case for why this will be useful for other projects going forward
+   - Present a plan for bringing other team members up to speed on the language you propose to use. Usually the language proposer will have to captain the process of language education.
+- Since it's a big decision, we should try to get input from all devs before making a decision at least till our team gets bigger
```
gammazero, line 83: > Since it's a big decision

Hopefully, it will not be a big decision, because the engineering requirements and design will naturally select a set of tools that best enables the creation, deployment, and maintenance of the designed solution.

alanshaw (approved):

rfc/languages-in-w3s.md
```diff
@@ -0,0 +1,84 @@
...
+- Lots and lots of Filecoin tooling and software is written in Go. In fact, W3S already has Go in it, we just don't maintain it -- it's called Spade
+- Garbage collection and green threads present challenges for using Go over an FFI, or for compiling go to WebAssembly, though TinyGo has improved this story by reducing the runtime size.
+
+### Rust
+
+Caveat from Hannah: I am not a highly experienced Rust dev
+
+- Rust is a better C/C++. Like definitely much better. This is huge cause no one wrote a better C/C++ for low level programming for 20 years.
+- However, by and large, you should probably not write programs in Rust unless you'd consider writing them in C/C++. There is minimal evidence one can be as productive in Rust as higher level languages.
+- In particular, you should probably not write web servers in Rust or any other asynchronous programs unless you have extra time on your hands which we don't.
```
Gozala, line 63: I find some of the statements about Rust inaccurate. I find myself a lot more productive reading and understanding Rust code than I do with go as it's type system forces you into specific design patterns. All the PL go code I had to deal is a nightmare in comparison, no comments terrible naming (one or two letter variable names 😅).

rfc/languages-in-w3s.md
```diff
@@ -0,0 +1,84 @@
...
+- By and large, people get stuff done quickly in Go because Go is extremely simple. It's very fast to learn, easy to read, and it all looks the same. The only problem is its extremely boring. (Note from Hannah: When I first started writing Go I hated it. I thought someone took the joy out of programming. Now I love it cause I get my joy from what I'm building, not my programming language)
+- Lots and lots of Filecoin tooling and software is written in Go. In fact, W3S already has Go in it, we just don't maintain it -- it's called Spade
+- Garbage collection and green threads present challenges for using Go over an FFI, or for compiling go to WebAssembly, though TinyGo has improved this story by reducing the runtime size.
+
+### Rust
+
+Caveat from Hannah: I am not a highly experienced Rust dev
+
+- Rust is a better C/C++. Like definitely much better. This is huge cause no one wrote a better C/C++ for low level programming for 20 years.
+- However, by and large, you should probably not write programs in Rust unless you'd consider writing them in C/C++. There is minimal evidence one can be as productive in Rust as higher level languages.
```
Gozala, line 62: I'm not sure what the basis of this assessment is ? Rust can be highly productive high level language as long as you grok resource ownership model. I would even argue that forcing you to think about ownership is a good thing and a perfect pairing with systems utilizing object capability model (like UCANs) because it's all about who owns the resource and who is given access at time.

rfc/languages-in-w3s.md
Gozala: I think this is a great article describing why discord has switched from go to rust https://discord.com/blog/why-discord-is-switching-from-go-to-rust

rfc/languages-in-w3s.md
```diff
@@ -0,0 +1,84 @@
...
+- In particular, you should probably not write web servers in Rust or any other asynchronous programs unless you have extra time on your hands which we don't.
+- BTW, if you haven't programmed in C/C++ recently, you've probably forgotten that long compile times are a thing on large projects in low level languages. 
+- Rust is essential for highly performant synchronous compute tasks. That's why people use it for smart contracts, hashing, and proofs.
+- Rust has the best WebAssembly story of any language.
```
Gozala, line 66: Rust is a great tool for writing libraries that you may ever conceive running elsewhere be it native code or web, no other language offers this. I know there are some efforts to make go better there, but GC is a sticking point. Rust has no GC which is why it's good fit and has no overhead, if you have to bring your GC with you that will be a non-trivial overhead and it would need to integrate with a host GC also. Maybe one day WASM will get a GC, there is certainly an effort to make it so, but I would not make any decision that assumes it, because chances are it's not going to be ready in time.

rfc/languages-in-w3s.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Table Stakes For Considering A Language for W3S
+
+In addition to having a compelling engineering or business reason, we have some other things to consider before we adopt a language:
+
+- UCAN libraries and tooling especially for anything with a web interface
+- IPLD/IPFS/Libp2p libraries (we may not always need libp2p going forward)
```
Gozala, line 73: I feel like this write up overlooks one important reason why we have sticked to JS:

1. All the cloud services CF, AWS (pretty much anyone else) have a best support for JS and some support for other languages. And if we don't ant to think about those services even PLware  station  is also JS.
   - We may reduce dependencies on such things in the future, but flexibility to use them in tandem is important IMO.
   - Rust is probably gaining momentum in all of those also mostly because of WASM and because those cloud services use Rust themself (as they need to scale)

I would also consider IPVM even if Fission is going away I keep hearing IPVM will carry on and so will everywhere.computer Even if those projects discontinue there is a lot of excellent code that was written for those that could be a great foundation to build from.

Gozala (commented):
Provided some feedback inline.

rfc/languages-in-w3s.md
```diff
@@ -0,0 +1,84 @@
...
+- Every hiring manager loves javascript cause it's easy to hire javascript developers. The main problem with this approach is there are so many kinds of javascript development and ways to write javascript that hiring a javascript developer is not a gaurantee they'll be able to write javascript the way your team uses it.
+
+#### Some Notes On Our Javascript
+
+- Thankfully this team has unified on a coding style, linter, tool chain, package manager, and approach to typing. +1 on that.
+- We have not however documented why we made the choices we did. We should do so ASAP. In particular types in JSDoc style comments and why we do it needs documentation. This includes our sporadic dropping into Typescript and why we do it. (For example: 'why is there an api.js and an api.ts and the js file just says `export {}` -- the answer is 'cause modules' but is EXTREMELY not clear)
```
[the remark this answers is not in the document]
hannahhoward: I actually have figured out rationales already over time. I'm saying they should be documented somewhere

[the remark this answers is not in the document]
alanshaw: 👍 yes I figured, was more for others ;)

I also added to our notion - forgot to put the link! https://www.notion.so/w3sat/Why-jsdoc-types-and-not-Typescript-eb6990cb04c94bebacdda6f2b5ce266a?pvs=4

rfc/languages-in-w3s.md
joaosa: That was an interesting read. Thanks 👍

rfc/languages-in-w3s.md
joaosa: I also agree that really depends on what you're trying to achieve. When I was first learning Rust, I rewrote a bunch of my personal organization scripts in Rust and it felt quite productive.

joaosa (commented):
Generally agree and left a few comments to other people's comments

rfc/languages-in-w3s.md
gammazero: This is great. Also, the async library has come a long way, and its progress is very promising.

gammazero: I think this is a good counter-example... but is also a lesson when porting from javascript to Go/Rust/etc. might be a bad idea. https://www.youtube.com/watch?v=_SzvJJ3_6M0

The gist is that a huge amount of time and effort was spent to arrive at an outcome with much more code that offered no significant advantages.

rfc/languages-in-w3s.md
hannahhoward: Ultimately, I'm interested not in the nice to haves but the must haves. Resource ownership is a great thing to understand. I take care to not pass around a lot of mutable data in whatever language I'm working in, and I try to have a pretty clear chain of provenance for data structures in my head. But do you need the strict resource ownership of a borrow checker for your specific program?

To me the answer is "only when absolute performance is a must". Cause there is a non-zero overhead to working without garbage collection.

The driver I want to keep coming back to is "what will get us to the end of our roadmap fastest". How fast can you write a high performance web server in Rust? Especially if you're building a new service from scratch? (many of the Rust success stories are often rewrites of an existing hot path service for performance reasons -- which makes sense when you reach the point optimization is no longer premature)

rfc/languages-in-w3s.md
hannahhoward: @Gozala 

The Discord example is an argument that when a code path becomes super hot, you can get more optimization by switching from a garbage collected language to one with manual memory management.

This feels both correct and also somewhat trivially obvious. Again this is where my "consider rust when you'd consider C/C++" statement comes from.

The more interesting question is when does a code path become hot enough to merit this sort of optimization, and what is the cost of trying to do this for all code paths, even the ones that don't need it.

rfc/languages-in-w3s.md
hannahhoward: I 100% agree with this.

If you want your code to run absolutely anyway, Rust is the go to.

rfc/languages-in-w3s.md
hannahhoward: Fundamentally I need more signal on IPVM continuing onward before I'd commit anything we do to using it.

hannahhoward: 100% agree on Javascript portability.

rfc/languages-in-w3s.md
```diff
@@ -0,0 +1,84 @@
...
+## Proposing a language to adopt
+
+- The best time to propose a language to adopt is at the start of a new project
+- Present a proposal to write a project in a new language via an RFC
+   - Try to objectively present business and engineering tradeoffs of writing the project in a language we already support vs writing it in a new language
+   - Identify gaps in language support or tooling related to our software stack and present a plan to address it
+   - Ideally, make a case for why this will be useful for other projects going forward
+   - Present a plan for bringing other team members up to speed on the language you propose to use. Usually the language proposer will have to captain the process of language education.
+- Since it's a big decision, we should try to get input from all devs before making a decision at least till our team gets bigger
+- Must have sign off from senior engineering leadership (for now Hannah and Alan)
```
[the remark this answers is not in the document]
hannahhoward: my concern with Rust has a lot to do with the speed of development for new code, as opposed to optimizing already well prototyped services.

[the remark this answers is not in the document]
hannahhoward: maybe @mroth wants to weight in here?

hannahhoward:
An interesting piece on Rust development from folks who've been doing it a while -- https://llogiq.github.io/2024/03/28/easy.html

rfc/languages-in-w3s.md
Stebalien: It depends on a bunch of factors.

- In the FVM, we had some issues because wasmtime needed a _static_ reference to some state and we didn't want to have rc/refcells all over the place. But, once we got past that issue, it wasn't too bad.
- In the builtin actors, we had a couple of issues translating existing go idioms, but wouldn't have had any issues if we had written them in rust first.
- If you dig too deep into the type system, there be Balrogs. You _will_ find things that really should compile but don't because rust just can't quite reason about your types.

The biggest drawbacks I see are:

1. Compile times and test time. The former is generally an issue if you're have a complex build pipeline with lots of deriving, macros, generics, etc. Especially if you have many build targets (e.g., the builtin actors). The latter can be an issue because, e.g., if you test with an unoptimized build (to improve compile times), tests can take a while in some cases (was also a problem in the builtin actors, fixed by compiling _specific_ crates with optimizations regardless of the build target).
2. Learning curve. This is something people get over, but it _does_ take a little longer to onboard new teammates.
3. **TEMPTATION**. The type system is _very_ expressive and it's easy to go down rabbit holes. This is one of the _biggest_ dangers I've seen in rust because there's always a good reason and the issue is only obvious in hindsight. Really, trying to find the perfect abstraction is probably the biggest rust time sink.
4. Async/await. IMO, it's still not ready (from what I've seen). It's getting better every release, but it still gets in the way enough to slow development down.

On the other hand, I absolutely love rust and write most of my personal projects in rust (those not written in Elisp). Mostly because:

1. Performance is very consistent. E.g., when we switched from the "legacy" VM to the FVM, block times, GC spikes, etc. got _much_ more consistent in lotus. A lot of this is simply because rust forces you to think about resource allocation and lifecycles, where as go requires some more discipline.
2. The type system is very expressive. We've had very few bugs in the FVM because we can often statically prevent them. But you have to be careful to not waste too much time trying to statically prevent all bugs.
3. The meta-programming and operator overloading allow for some fun patterns _when_ you have common patterns you need to abstract away (they can also be a time sink if you put too much effort into them up-front).

I actually do write high-level tools in rust if I need them to just work and it generally doesn't take that long. There are lots of really great high-level libraries for writing CLI tools. I used to write them in python, but then getting them to work reliably always took a while (need to write tests, etc.).

Rust only starts to slow you down when you get into more complex projects and start building out complex abstractions. And, TBH, a lot of the trouble is self-inflicted because rust _lets_ you build those abstractions.

hannahhoward: ^^ @Stebalien thanks for taking the time 

Folks I think this is worth reading

Gozala: > * In the builtin actors, we had a couple of issues translating existing go idioms, but wouldn't have had any issues if we had written them in rust first.

I actually was going to comment something along those lines but restrained myself. Going from TS to Rust would feel a lot more at home than going from TS to Go or from Go to Rust. In fact most of our TS actually feels more like Rust than JS because we put a lot of care into resource management to keep memory use at desired bounds.

> TEMPTATION. The type system is very expressive and it's easy to go down rabbit holes. This is one of the biggest dangers I've seen in rust because there's always a good reason and the issue is only obvious in hindsight. Really, trying to find the perfect abstraction is probably the biggest rust time sink.

I could not agree more! Most typed languages (I would not put go into this group) it's really easy to fall in the trap of trying to increase safety by capturing more in the type system. I find Rust and TS both to be particularly luring there, but then again solution we often resort to is: Dont' abstract until you have an actual use case for that abstraction, or put it differently if you have one implementation of the abstraction that abstraction is just a means of indirection and unnecessary complexity.

rfc/languages-in-w3s.md
gammazero: > Use javascript when you really don't know where your code might run
... but you can be certain about the runtime environment (in browser, node-js version, etc.)

Use go to build executables when you know nothing about where it is running, only that it is not in a browser.

rfc/languages-in-w3s.md
Stebalien: Also, don't spend a bunch of time re-writing your abstractions. Yes, I'd agree that TS is similar in that respect (and not in a good way).

rfc/languages-in-w3s.md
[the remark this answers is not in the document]
gammazero: One concern with Rust for larger projects is that it can slow down iterative development. There is more code, due to the need to precisely manage resources, among other things, and refactoring tends to be more difficult. With Go, it is relatively easy to refactor as a project evolves, allowing for rapid iteration of development cycles.

In addition to longer refactor time, rust compile times are significantly longer than with Go. The difference becomes pronounced as the amount of code and dependencies grows. This makes testing and trying out new code quicker in Go, resulting in considerable time savings over the course of development.

This echos what was already said by @hannahhoward, and is in the spirit of "consider rust when you'd consider C/C++".

Rust is an awesome language and enables one to write highly optimal and safe code. However, I think it is a mistake to fall in love with this without also understanding the value of what Go beings. That is: Go does everything it can to minimize thought and time spent getting the compiler to do exactly what you want, allowing you to focus on best expressing the logic needed to solve whatever problem you are trying to solve. There is some cost to that (like GC), but the benefits of productivity and clear maintainable code very often far outweigh that cost.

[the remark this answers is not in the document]
Stebalien: > refactoring tends to be more difficult

It depends. Sometimes a structural change is difficult due to the ownership model. On the other hand, sometimes a refactor is easy because, once something compiles, it usually works.

Bug I agree that iteration cycles in rust tend to be slower. E.g., I wouldn't prototype a project with an _unknown_ architecture in rust if I could help it. I'd start by prototyping and answering all the architectural questions in Go, then I'd rewrite it in rust.

---

One more thing for rust, however: C compat. CGO is a nightmare.

PR #28, opened 2024-05-28
feat: egress billing RFC

travis:
First writeup of the proposal for egress billing. Looking for feedback on the ideas and on the coherence of the writeup.

### DIFF
--- a/rfc/egress-billing.md
+++ b/rfc/egress-billing.md
@@ -1,0 +1,251 @@
+# w3s.link gateway egress billing 
+
+## Authors
+
+- [Travis Vachon](https://github.com/travis)
+- [Irakli Gozalishvili](https://github.com/gozala)
+
+## Goals
+
+We need to start charging for data served by our gateway ("gateway egress"). Users who upload data using `w3up` should
+be charged when we serve that data through the gateway that is currently hosted at https://w3s.link. Importantly, if 
+two users upload the same data, we need to ensure the correct party is billed when we serve that data through
+the gateway - for instance, if I upload a file that a massive NFT platform also uploaded, I should not be charged
+when that platform serves that data to millions of their customers through their website.
+
+Charging for *all* reads is a non-goal - we believe there is a tradeoff between strict accounting and read performance
+and we generally want to err on the side of performant reads.
+
+## Abstract
+
+The business has the capacity to pay for some number of reads - we'll call this R. We will
+first implement a global per-CID rate-limit that is expected to account for roughly 75-90%
+of R. This rate limit will be used for all content that has not been configured with another
+rate limit, and for content that has been configured with other rate limits that have already
+been reached. If a request is made for content that has "used up" all available rate limits, including
+its public rate limit, the response MAY be an HTTP 429 (`Too Many Requests`), but MAY be an HTTP 200 (`OK`) with the content
+if doing so allows us to optimize read performance. Requests served despite being over all rate
+limits will account for the remaining 10-25% of R. [note: I'm not sure if these percentage ranges for R are right]
+
+In order to allow our users to guarantee consistent read performance in their applications, we
+will allow them to configure rate limits using the “content commitments“ returned from blob uploads. 
+These content commitments can be delegated to the gateway and parameterized to set rate limits
+in a flexible and extensible way - essentially, a user will authorize the gateway to serve their content
+with some restrictions. At read time, if the gateway can find a valid content commitment it will serve the content,
+and if not it will fall back to the behavior described above.
+
+Finally, in order to give ourselves maximum flexibility to optimize read performance, we propose
+a high performance rate-limit tracking system based on a mature caching service - we will
+use [Redis] in this document as an example, but other services like memcached or Caddy should be considered. 
+When read requests come in to the gateway, this system will be used to determine whether the request should
+be served. After the request is served it will be passed to a backend job processing system which
+will both record the system for administrative purposes (eg, billing) AND update the rate limit caching
+system. We expect this to result in some amount of "unauthorized" read request service, for example if two requests for
+the same content arrive at nearly the same time they will both be served 200s, since the rate limits will only be
+updated some time after a request is served. This pattern is similar in some ways to the "stale while revalidate" 
+pattern that is popular on the web today - stale rate limits will be used while they are revalidated in the background.
+
+Taken together, this should allow us to keep reads extremely performant - ideally we will need only a single Redis
+request and a small amount of computation to decide whether to serve content. It will also give users an
+extremely flexible and extensible mechanism to control how and when their content is served, including support for
+`Origin`-gating and other CDN-esque tools to control their egress costs.
+
+## Introduction
+
+We currently operate the IPFS gateway at https://w3s.link as a free service. As a result, it is an unbound source of "cost"
+for our business. Covering some amount of the costs associated with this service make sense as a marketing and community
+building expense, but operating it for free makes it prohibitively expensive to give our customers acceptable read performance
+for their applications through our gateway.
+
+In order to fix this, we propose a design for billing our customers for data egress related to their applications. The suggested
+design allows us to maintain extremely fast read performance by allowing us to configure the "acceptable level" of "unpaid"
+data egress and giving our customers a way to authorize paid data egress using the "content claims" returned by the 
+[blob upload service].
+
+## Rate Limits
+
+Business realities (runway, need for marketing and community growth, etc) will determine an acceptable level of data egress
+through our gateway that can be paid for by the business. We will call this _R_, the "number of requests we're willing to pay for."
+
+We will use a tool like Redis to implement this rate limit, because it has a long history of being used for this and
+the right set of primitives to support it well: https://redis.io/glossary/rate-limiting/ - we will use Redis as an example
+in this document but recommend researching other tools like memcached and Caddy before making a final implementation decision.
+
+When the gateway receives a request for some content, it will make a request to Redis for the relevant rate limit data and decide to
+serve the request based on the response. In order to maintain acceptable read performance, we recommend avoiding any other network
+IO in making the decision to serve a piece of content. In the event that content is served as a result of the rate limit caching service
+being out of date, we count that request against _R_. 
+
+The following TypeScript pseudocode illustrates the rate limiting algorithm:
+
+ts
+```
+enum RateLimited {
+  Yes,
+  No,
+  Maybe,
+}
+
+async function handle(request){
+  const rateLimited = checkRedisToDetermineIfRateLimited(request)
+  const cid = getCidFromRequest(request)
+  switch (rateLimited) {
+    case Yes: 
+      return { status: 429, message: `${cid} is currently rate limited, please try again later` }
+    case No:
+      return { status: 200, body: readContentFromW3up(cid) }
+    case Maybe:
+      return { status: 200, body: readContentFromW3up(cid) }
+  }
+}
+```
+
+### User Configuration
+
+In order to allow our users to use uploaded content in their applications and guarantee acceptable read performance,
+we propose use the "[content commitments]" generated by the w3up [blob upload service].
+
+```js
+{ // Content commitment with CID bafy..site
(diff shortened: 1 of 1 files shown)


rfc/egress-billing.md
```diff
@@ -0,0 +1,309 @@
...
+sequenceDiagram
+  Alice->>Blob Service: Please store bafk...7fi
+  Blob Service-->>Alice: Stored - you can present bafy...site to fetch bafk...7fi
+  Alice-)Group: You can present bafy...group to fetch bafk...7fi
+  Group-)Gateway: You can present bafy..access to fetch bafk...7fi
+```
+
+Note that `bafy..access` alone will not be enough to authorize retrieval
+of the content - the accounting system will need to combine it with
+a CID-specific content commitment that has been delegated to `Group`.
```
alanshaw, line 253: I'm not following how this works, I think you need to show the "CID-specific content commitment that has been delegated to `Group`"

rfc/egress-billing.md
```diff
@@ -0,0 +1,309 @@
...
+Charging for *all* reads is a non-goal - we believe there is a tradeoff between strict accounting and read performance
+and we generally want to err on the side of performant reads.
+
+## Abstract
+
+The business has the capacity to pay for some number of reads - we'll call this _R_. We will
+first implement a global per-CID rate-limit that is expected to account for N%
+of _R_. This rate limit will be used for all content that has not been configured with another
+rate limit, and for content that has been configured with other rate limits that have already
+been reached.
```
alanshaw, line 25: It would be nice to model this as "just another user that set a rate limt", where "another user" is us. Eventually, some entity will have to pay the egress from the nodes storing the content, so we do need to be able to allocate responsibility for that egress to some entity.

i.e. there's no magic "global" limit, it's just some user (us) that has set a rate limit that happens to apply to ALL content served by the gateway.

rfc/egress-billing.md
```diff
@@ -0,0 +1,309 @@
...
+is acceptable since Redis will eventually be updated and the amount of reads that fall into this category are
+expected to be within the bounds of the read capacity the business is willing to serve.
+
+On the topic of Redis, it's worth noting that it is no longer open source:
+
+https://redis.io/blog/redis-adopts-dual-source-available-licensing/
+
+We may want to consider using an open source alternative like https://github.com/nalgeon/redka if we
+are concerned about using "source available" products - my inclination is to use Redis but consider products like
+Redka as a fallback in case Redis no longer meets our needs. We may also want to consider using AWS or Cloudflare products with Redis-compatible APIs:
```
alanshaw, line 277: Cloudflare's KV is _super_ fast since it's eventually consistent, which feels like a good fit.

rfc/egress-billing.md
```diff
@@ -0,0 +1,309 @@
...
+with some restrictions. At read time, if the gateway can find a valid content commitment it will serve the content,
+and if not it will fall back to the behavior described above.
+
+Finally, in order to give ourselves maximum flexibility to optimize read performance, we propose
+a high performance rate-limit tracking system based on industry-standard tools for this purpose. After
+the request is served it will be passed to a backend job processing system which will update the rate limit caching
+system. We expect this to result in some amount of "unauthorized" read request service - for example, if two requests for
+the same content arrive at nearly the same time they will always both be served even if the second request
+goes over all established rate limits, since the rate limits will only be updated some time after a request is served. 
+This pattern is similar in some ways to the "stale while revalidate"  pattern that is popular on the web today - stale
```
alanshaw, line 40: "fail open" also comes to mind...

rfc/egress-billing.md
```diff
@@ -0,0 +1,309 @@
...
+
+## Authors
+
+- [Travis Vachon](https://github.com/travis)
+- [Irakli Gozalishvili](https://github.com/gozala)
+
+## Goals
+
+We need to start charging for data served by our gateway ("gateway egress"). Users who upload data using `w3up` should
+be charged when we serve that data through the gateway that is currently hosted at https://w3s.link. Importantly, if
```
vasco-santos, line 11: are we strictly tackling gateway, or do we also plan to bitswap? I think some of the ideas here could be applied there as well, even though with extra layer of indirection to figure out the blob CID where a block CID belongs to

travis: ah I was thinking of the "gateway" as something that is also handling bitswap - imho the gateway and bitswap provider should both work off the same rate limits

rfc/egress-billing.md
```diff
@@ -0,0 +1,309 @@
...
+
+```ts
+enum RateLimited {
+  Yes,
+  No,
+  Maybe,
+}
+
+async function handle(request){
+  const rateLimited = checkRedisToDetermineIfRateLimited(request)
```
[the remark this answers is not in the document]
travis: ah I hear you but I am trying to convey that _only_ "Redis" should consulted here - I've come to think of "Redis" as an API not a product (because that seems to be the reality of it in 2024) - I do mention earlier that I'm using Redis as an example and I think this is consistent with that, so I'd like to keep it

rfc/egress-billing.md
```diff
@@ -0,0 +1,309 @@
...
+  const cid = getCidFromRequest(request)
+  switch (rateLimited) {
+    case Yes: 
+      return { status: 429, message: `${cid} is currently rate limited, please try again later` }
+    case No:
+      return { status: 200, body: readContentFromW3up(cid) }
+    case Maybe:
+      return { status: 200, body: readContentFromW3up(cid) }
```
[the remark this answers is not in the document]
travis: ah I was worried that people who aren't JS-brained would misread that as "if No, do nothing" - I could definitely be convinced but imho the way I've done it makes it a little clearer that something should be done in each case, and since they are right next to eachother it's pretty easy to verify that they're the same thing...

rfc/egress-billing.md
```diff
@@ -0,0 +1,309 @@
...
+  // Indefinitely
+  "exp": null,
+  "cmd": "/assert/location",
+  // Subject is storage node because they own that resource
+  "sub": "did:web:asia.web3.storage",
+  "pol": [
+    // multihash must match an uploaded blob
+    ["==", ".content", { "/": { "bytes": "mEi...sfKg" } }],
+    // must be available from this url
+    ["==", ".url", "https://asia.w3s.link/ipfs/bafk...7fi"],
```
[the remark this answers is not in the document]
travis: that makes sense - I was cargo culting off of @Gozala so redirect this one over to him for confirmation...

rfc/egress-billing.md
```diff
@@ -0,0 +1,309 @@
...
+{ // bafy..auth
+  "iss": "did:key:zAlice",
+  "aud": "did:web:w3s.link",
+  "exp": 1716235987 // restrict to a month
+  "cmd": "/assert/location",
+  // Subject is storage node because they own that resource
+  "sub": "did:web:asia.web3.storage",
+  "pol": [
+    // Request origin header must be example.com
+    ["==", ".headers['origin']", "example.com"],
```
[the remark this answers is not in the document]
travis: hmm, nothing that I know of - again just copying @Gozala, maybe he was trying to illustrate that either style can be used? I'm happy to use `.origin`!

rfc/egress-billing.md
```diff
@@ -0,0 +1,309 @@
...
+  Blob Service-->>Alice: Stored - you can present bafy...site to fetch bafk...7fi
+  Alice-)Gateway: You can present bafy...auth to fetch bafk...7fi
+```
+
+#### Rate Limit Accounting
+
+When the gateway receives a request, it will query the [content claims service] for claims about the requested
+CID [TODO: and also other information like origin and token?]. If multiple claims establishing different rate limits
+are found, one is chosen randomly and its request count is incremented, resulting in billing to the 
+party that authorized the rate limit. If no claims are found, the public rate limit request count is incremented.
```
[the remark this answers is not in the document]
travis: the rate limit delegation essentially _is_ a content commitment so I think the second part won't be necessary? this is an area where my understanding is a bit murky tho, something I'm hoping to fix by prototyping a bit, so this may be correct...

rfc/egress-billing.md
```diff
@@ -0,0 +1,309 @@
...
+  "sub": "did:web:asia.web3.storage",
+  "pol": [
+    // Request origin header must be example.com
+    ["==", ".headers['origin']", "example.com"],
+    ["==", ".query.token", "zrptvx"]
+  ]
+}
+```
+
+For any given CID, the gateway will be able to look up the CID-specific content commitment as well as the
```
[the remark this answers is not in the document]
travis: oh that's a good point, I actually don't know how this will handle CIDs that aren't DAG roots - do those even get content commitments back from the blob upload service? @Gozala any thoughts on this? I'm not sure we accounted for these CIDs in our chats...

rfc/egress-billing.md
travis: yea completely agree!

rfc/egress-billing.md
travis: ah cool will add it to the list!

I will say that one of the core requirements here is the ability to implement efficient rate limits - you can see an example of what that could mean here: https://redis.io/glossary/rate-limiting/

I think (but am not sure) that one of the key things is support for the `INCR` operation, which I don't think KV implements, and expiring entries, which may be better supported? 

The good news is that lots of stuff seems to implement the Redis query interface these days, including Cloudflare's Upstash and no less than two different AWS products - I think we'll have options!

rfc/egress-billing.md
travis: yea love this!

rfc/egress-billing.md
vasco-santos: probably I would make this extra specific and add it to spec then :) There are things like `origin` that are actually only specific for HTTP and as far as I know won't be available in bitswap.

Gozala: > There are things like origin that are actually only specific for HTTP and as far as I know won't be available in bitswap.

I would expect that libp2p node will be delegated capability explicitly and constraints there may be different from the ones imposed on gateway (described here). I'm not sure what those constraints could be perhaps something like peer id that can perform a request.

rfc/egress-billing.md
Gozala: 🎯

rfc/egress-billing.md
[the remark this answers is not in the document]
Gozala: It could be a range yeah, but I think we made it optional if it's not a segment but the whole thing

rfc/egress-billing.md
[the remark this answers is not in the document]
hannahhoward: ^^^ important point. I think there's an intersection where we need to understand how this relates to the indexing RFCs. Fine to just handle DAG root to start but having w3s.link ONLY work for DAG root is very frustrating and I see the indexing work heading in a direction where we're not limited in that way.

rfc/egress-billing.md
```diff
@@ -0,0 +1,309 @@
...
+  Alice->>Blob Service: Please store bafk...7fi
+  Blob Service-->>Alice: Stored - you can present bafy...site to fetch bafk...7fi
+  Alice-)Gateway: You can present bafy...auth to fetch bafk...7fi
+```
+
+#### Rate Limit Accounting
+
+When the gateway receives a request, it will query the [content claims service] for claims about the requested
+CID [TODO: and also other information like origin and token?]. If multiple claims establishing different rate limits
+are found, one is chosen randomly and its request count is incremented, resulting in billing to the
```
hannahhoward, line 178: There's an important clarification I hope I'm reading right.

When are multiple claims found?

If you have to match token and origin, why would you have multiple claims? 

Or maybe you have the matching unlimited claim and the general open access rate limited claim? In which case I'd lean towards using the unlimited claim rather than picking at random.

hannahhoward (approved):
I think I'm approving generally, but if we're going to work on Egress billing and indexing at once, we really need to sync on these two designs. I think we still have some confusion and I wonder if when we move beyond prototype we do the indexing first cause it's really essential to all this.

PR #29, opened 2024-05-29
Indexing design for w3s

gammazero:
Functional design for w3 indexing.

Replaces the previous RFC [W3 IPNI Indexing RFC](https://github.com/w3s-project/RFC/blob/main/rfc/ipni-w3c.md)

### DIFF
--- a/rfc/indexing-design.md
+++ b/rfc/indexing-design.md
@@ -1,0 +1,228 @@
+# W3 Indexing Design
+> Replaces [W3 IPNI Indexing RFC](https://github.com/w3s-project/RFC/blob/main/rfc/ipni-w3c.md)
+
+## Authors
+
+- [gammazero]
+- [gozala]
+
+## Goals
+
+A user can use a multihash, that identifies data stored on w3s, to lookup the all information necessary to retrieve and delegate access to that data.
+
+Location commitments that describe where content is stored can change over time as w3s moves data to a more favorable location, re-aggregates data storage, or replicates stored data. Location commitments are updated without having to index all multihashes.
+
+A query API will allow a user to specify indexed content to lookup and how to filter results. The API retrieves previously cached responses or composes new IPNI queries to get new responses to filter and cache.
+
+## Overview
+
+A user stores content as opaque blobs containing series of blocks addressed by multihashes. These blocks are collected into one or more CAR file (shards) and the CAR file is stored at a location designated for a particular user's data.
+
+The user creates a blob-index ([sharded-dag-index](https://github.com/w3s-project/specs/blob/main/w3-index.md#sharded-dag-index-example)) that contains the multihash of each shard of blob data, and an array of tuples of (multihash, offset, length) for all the blocks in each shard. The user may choose which blocks to include in the blob-index if they do not want all blocks to be indexed. For example, they may want only the DAG root to be indexed.
+
+The blob-index is stored in the user's space along with the CAR file(s) containing the blob data. This is done in the [w3 add-blob operation](https://github.com/w3s-project/specs/blob/main/w3-blob.md#add-blob). The w3s service creates [location commitments](https://github.com/w3s-project/specs/blob/main/w3-blob.md#location-commitment) for the blob shards and for the blob-index. 
+
+After the add-blob process completes, and the user receives an [accept-blob receipt](https://github.com/w3s-project/specs/blob/main/w3-blob.md#accept-blob-receipt), the user may then choose to make their data publicly queryable by publishing it on [IPNI](https://github.com/ipni/specs/blob/main/IPNI.md). The user optionally invokes an [index-add](https://github.com/w3s-project/specs/blob/main/w3-index.md#index-add) capability to publish the blob-index multihashes to IPNI so that they can be used to look up location commitments and retrieve blob-index information and blob data. See [W3 Index](https://github.com/w3s-project/specs/blob/main/w3-index.md#w3-index) for more.
+
+After publishing to IPNI, any user looking for content can query IPNI for the blob-index of that content. The user read the blob-index to determine what data shard needs to be retrieved, and then asks IPNI for the location of that shard.
+
+Finally the user uses the shard location, with range information from the blob index, to retrieve a specific piece of data from w3s.
+
+## Index Query
+
+A w3s client can query indexed information by specific attributes, and can filter retrieved query results.
+
+The queryable attributes are as follows.
+- Site: the location(s) where all blob data is available
+- Shard: the location(s) where a specific shard of data is available
+- Slice: the location(s) within a shard where a block of data is available
+
+Query results can be filtered to select only results that the user has access to, or for data supplied by specific data providers.
+
+### Query Specification
+
+```
+match = {
+  subject=[myID, myOrgID],
+  location=spBaseURL,
+}
+results = Query(multihash="bafy...", match)
+```
+
+### Query Operation Diagram
+
+![w3-index](https://www.tldraw.com/ro/jBOq7Ce4ILWloWfAODrlj?v=-183,481,2733,1758&p=page)
+
+## Indexing Components
+
+The indexing components are intended for deployment on the w3up gateway. They may also be deployed to the w3up client, where they will not be shared with other clients.
+
+### W3up Index
+
+The w3up index layer is responsible for reading a query specification, and then
+1. Creating a set of IPNI queries needed to get the requested index content
+2. Sending the queries to the w3up IPNI cache
+3. Filtering returned results
+4. Presenting results to the query client.
+
+### W3up IPNI Cache
+
+The w3up IPNI Cache is a cache that holds IPNI query results. The cache is:
+- Limited-size LRU. Discards the least recently used data when at storage capacity.
+- Temporary. Evicts items that have been cached for longer than the cache expiration time, regardless of last access. This allows changes to query results to be seen.
+- Negative (empty) result-aware. Caching empty results prevents more expensive IPNI queries for data that is not indexed.
+
+This is a read-through cache, meaning if it does not hold the request query results, then it forwards the query on to IPNI, and caches the results. This includes caching an empty response.
+
+There are two types of queries send to the cache. Each takes a data multihash, but returns a different result:
+1. shard/slice query. Returns one or more: shard CID, slice CID, offset, length (in shard).
+2. location query. Returns: location commitment of each shard. 
+
+By using both these query results, a specific portion of the blob data can be retrieved. The shard(s) are read from the locations in the location commitments, and data ranges within each shard are requested (when not reading an entire shard) by range requests using the offsets and lengths from the shard/slice data.
+
+If either shard or location results are not cached, then a query for the multihash is passed to the IPNI client.
+
+#### Expiration Times
+The shard/slice data does not change over the lifetime of the stored data, so this data can have a much longer expiration time. It could also have no expiration time and be explicitly removed from cache when the data is no longer stored, but this means that there need to be a communication path from data removal to all cache locations. So, better to just have an expiration time, making cache more independent.
+
+### IPNI
+
+#### Get w3up results
+
+IPNI maps a multihash to location and metadata information. When queried with a multihash, IPNI will respond with all known locations/metadatas of that multihash. The w3up IPNI query client must filter these results to only select results that are under the control of w3up.
+
+Example IPNI query and response:
+```bash
+curl https://cid.contact/cid/bafybeicawc3qwtlecld6lmtvsndimoz3446xyaprgsxvhd3aapwa2twnc4 -sS | jq
+```
+```json
+{
+  "MultihashResults": [
+    {
+      "Multihash": "EiBAsLcLTWQSx+WydZNGhjs75z18AfE0r1OPYAPsDU7NFw==",
+      "ProviderResults": [
+        {
+          "ContextID": "YmFndXFlZXJheTJ2ZWJsZGNhY2JjM3Z0em94bXBvM2NiYmFsNzV3d3R0aHRyamhuaDdvN2o2c2J0d2xmcQ==",
+          "Metadata": "gBI=",
+          "Provider": {
+            "ID": "QmQzqxhK82kAmKvARFZSkUVS6fo9sySaiogAnx5EnZ6ZmC",
+            "Addrs": [
+              "/dns4/elastic.dag.house/tcp/443/wss"
+            ]
+          }
+        },
+        {
+          "ContextID": "YmFndXFlZXJheTJ2ZWJsZGNhY2JjM3Z0em94bXBvM2NiYmFsNzV3d3R0aHRyamhuaDdvN2o2c2J0d2xmcQ==",
+          "Metadata": "oBIA",
+          "Provider": {
+            "ID": "QmUA9D3H7HeCYsirB3KmPSvZh3dNXMZas6Lwgr4fv1HTTp",
+            "Addrs": [
+              "/dns4/dag.w3s.link/tcp/443/https"
+            ]
+          }
+        }
+]}]}
+```
+
+#### Get sharded-dag-index
+
+After selecting only w3up results, these must then be processed to generate responses for shard/slice and location query attributes. To do this, first the blob-index result is read from the IPNI result metadata. This gets the CID of the sharded-dag-index (aka blob-index) and the location portion of the result gives the base URL to retrieve the blob-index from.
+
+The blob-index is retrieved and read. The blob-index data is searched to find the multihash, and to get the shard and slice (if the multihash is a slice within a shard).
+
+#### Get locations of shards
+
+After getting the shard and slice data, then location commitment data needs to be collected.
+
+If the multihash is that of a whole shard, then the location commitment results are already in the IPNI results. If the multihash is a slice, then another IPNI query needs to be made to retrieve the location(s) of the shard, using the parent shard CID from the blob-index.
+
+The location commitments are read from IPNI metadata and are returned to the cache.
+
+## IPNI Advertisement
+Information is provided to IPNI in the form of [IPNI Advertisements](https://github.com/ipni/specs/blob/main/IPNI.md#advertisements) that are signed by w3s and published to IPNI by w3s. When a new advertisement is published, an [IPNI Announcement](https://github.com/ipni/specs/blob/main/IPNI.md#announcements) is sent to IPNI to trigger IPNI to fetch the new advertisement.
+
+A separate advertisement chain is used for location commitments and blob-index data.
+
+The reason for this is that location information is subject to change, but blob index data will not change over the lifetime of the stored data.
+
+### Advertisement Structure
+
+An IPNI Advertisement has the IPLD Schema:
+
+```ipldsch
+type Advertisement struct {
+    PreviousID optional Link
+    Provider String
+    Addresses [String]
+    Signature Bytes
+    Entries Link
+    ContextID Bytes
+    Metadata Bytes
+    IsRm Bool
+    ExtendedProvider optional ExtendedProvider
+}
+
+type EntryChunk struct {
+    Entries [Bytes]
+    Next optional Link
+}
+```
+
+### IPNI Advertisement for sharded-dag-index
+
+Map all multihashes to blob-index CID/location
+ContextID = hash(userKey + blob-index CID)
+
+- `ProviderID` libp2p peer ID of w3s blob-index publisher.
+- `Addresses` the most recent value of this is returned with lookups for any CID in any advertisement. It will contain the multiaddr form of the blob-index publisher URL. This is where the blob-index is retrieved from when presented with the blob-index CID.
+- `Signature` Signature over all advertisement data. Created using private key matching `ProviderID`.
+- `ContextID` byte encoded userKey + blob-index CID.
+- `Entries` must be the CID of an `EntryChunk` for a subset (or all) of the multihashes in the blob-index.
+- `Metadata` encodes the CID of the blob-index + version info is any.
+- `IsRm` is used when removing all advertisement(s) that have that context ID and deleting all associated multihash indexes from IPNI.
+
+The set of multihashes within one blob must be encoded as 1 Advertisement. If a blob is partitioned into multiple sub-blobs this still generates one advertisement as there is still one blob index.
+
+Each advertisement encodes the multihashes into 1 or more EntryChunk blocks. EntryChunk blocks are an array of multihashes with a link to a next block, if there is another block. The EntryChunk serves a similar purpose as HTTP chunking. 
+
+All multihashes must be sorted in ascending order before being split into chunks. This is done to have deterministic output when creating entry chinks - the same set of multihashes should generate the same set of entry chunks.
+
+### IPNI Advertisement for location commitment
+
+Map shard-multihash to shard CID/location
+ContextID = hash(userKey + shard CID)
+
+- `ProviderID` libp2p peer ID of w3s location publisher.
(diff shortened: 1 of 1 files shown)


rfc/indexing-design.md
```diff
@@ -0,0 +1,228 @@
...
+match = {
+  subject=[myID, myOrgID],
+  location=spBaseURL,
+}
+results = Query(multihash="bafy...", match)
+```
+
+### Query Operation Diagram
+
+![w3-index](https://www.tldraw.com/ro/jBOq7Ce4ILWloWfAODrlj?v=-183,481,2733,1758&p=page)
```
vasco-santos, line 54: could you inline here the diagram tldr code in a different file and a svg? Github does not preview it this way

vasco-santos: @gammazero can you please also add the tldr code? see as example other PRs in this repo https://github.com/w3s-project/RFC/pull/13/files that do so. It makes it much easier for other folks on the team to iterate on something

rfc/indexing-design.md
```diff
@@ -0,0 +1,228 @@
...
+results = Query(multihash="bafy...", match)
+```
+
+### Query Operation Diagram
+
+![w3up-query-diagram](w3up-query-diagram.png)
+
+## Indexing Components
+
+The indexing components are intended for deployment on the w3up gateway. They may also be deployed to the w3up client, where they will not be shared with other clients.
```
Gozala, line 58: I think it's not super clear what the indexing component refers to here. By name I would expect component that derives the index, but I don't think it's what you mean.

rfc/indexing-design.md
```diff
@@ -0,0 +1,228 @@
...
+
+### W3up Index
+
+The w3up index layer is responsible for reading a query specification, and then
+1. Creating a set of IPNI queries needed to get the requested index content
+2. Sending the queries to the w3up IPNI cache
+3. Filtering returned results
+4. Presenting results to the query client.
```
Gozala, line 66: Not sure if this improves it, but I think something about aggregating results is worth capturing

```suggestion
4. Packaging results for the query client.
```

rfc/indexing-design.md
```diff
@@ -0,0 +1,228 @@
...
+The w3up IPNI Cache is a cache that holds IPNI query results. The cache is:
+- Limited-size LRU. Discards the least recently used data when at storage capacity.
+- Temporary. Evicts items that have been cached for longer than the cache expiration time, regardless of last access. This allows changes to query results to be seen.
+- Negative (empty) result-aware. Caching empty results prevents more expensive IPNI queries for data that is not indexed.
+
+This is a read-through cache, meaning if it does not hold the request query results, then it forwards the query on to IPNI, and caches the results. This includes caching an empty response.
+
+There are two types of queries send to the cache. Each takes a data multihash, but returns a different result:
+1. shard/slice query. Returns one or more: shard CID, slice CID, offset, length (in shard).
+2. location query. Returns: location commitment of each shard.
```
Gozala, line 79: I'm not sure while shard and slice queries are grouped into one type and the location into the other, I suspect there is some reason which is probably worth outlining

Gozala: Separately I think it would be nice if this layer was generic - agnostic to kinds of query types. That is to say it should not care what attribute `blob/shard`, `blob/slice` or `commitment/location` is queried by it simply can derive the cache key and on cache miss miss translate query into an IPNI query without worrying about the semantics.

Maybe that is the intention, but reading this I left with an impression that it's not generic over the attributes.

rfc/indexing-design.md
```diff
@@ -0,0 +1,228 @@
...
+4. Presenting results to the query client.
+
+### W3up IPNI Cache
+
+The w3up IPNI Cache is a cache that holds IPNI query results. The cache is:
+- Limited-size LRU. Discards the least recently used data when at storage capacity.
+- Temporary. Evicts items that have been cached for longer than the cache expiration time, regardless of last access. This allows changes to query results to be seen.
+- Negative (empty) result-aware. Caching empty results prevents more expensive IPNI queries for data that is not indexed.
+
+This is a read-through cache, meaning if it does not hold the request query results, then it forwards the query on to IPNI, and caches the results. This includes caching an empty response.
```
Gozala, line 75: I think caching an empty response can be problematic as we may not have a record on first query which would get cached and prevent us from discovering records on subsequent queries. Perhaps short TTL on empty result could be a way to address this or some form of exponential backoff.

rfc/indexing-design.md
```diff
@@ -0,0 +1,228 @@
...
+There are two types of queries send to the cache. Each takes a data multihash, but returns a different result:
+1. shard/slice query. Returns one or more: shard CID, slice CID, offset, length (in shard).
+2. location query. Returns: location commitment of each shard. 
+
+By using both these query results, a specific portion of the blob data can be retrieved. The shard(s) are read from the locations in the location commitments, and data ranges within each shard are requested (when not reading an entire shard) by range requests using the offsets and lengths from the shard/slice data.
+
+If either shard or location results are not cached, then a query for the multihash is passed to the IPNI client.
+
+#### Expiration Times
+The shard/slice data does not change over the lifetime of the stored data, so this data can have a much longer expiration time. It could also have no expiration time and be explicitly removed from cache when the data is no longer stored, but this means that there need to be a communication path from data removal to all cache locations. So, better to just have an expiration time, making cache more independent.
```
Gozala, line 86: I'm not sure we need to evict information about data even if we don't store that data. It could be that location of the data is not in our system (e.g. shared privately) while index of it is shared publicly. That is to say I don't think we need to worry about removing shard/slice info and instead let the user perform manual eviction when they want to do so. Specifically I imagine user would delete CAR holding an index which could trigger eviction

rfc/indexing-design.md
```diff
@@ -0,0 +1,228 @@
...
+          }
+        }
+]}]}
+```
+
+#### Get sharded-dag-index
+
+After selecting only w3up results, these must then be processed to generate responses for shard/slice and location query attributes. To do this, first the blob-index result is read from the IPNI result metadata. This gets the CID of the sharded-dag-index (aka blob-index) and the location portion of the result gives the base URL to retrieve the blob-index from.
+
+The blob-index is retrieved and read. The blob-index data is searched to find the multihash, and to get the shard and slice (if the multihash is a slice within a shard).
```
Gozala, line 131: This goes in a different direction than what I was imagining. Specifically I thought that IPNI publisher would derive and publish advertisements from the sharded-dag-index so that client querying IPNI would not have to fetch the index. Specifically idea was that multihash would be combined with an attribute / relation being queried allowing us to W3Up IPNI cache to paralellize queries as opposed to having to block on index retrieval before it is able to pull out details.

Part of this was also motivated by desire to make things generic so new kinds of indexes could be introduced by end users or even us without affecting any of the layers of this system.

Gozala: Perhaps a hybrid model could offer best of both worlds ? E.g. we could use index identifier as a namespace and expect that all attributes under that namespace would be aggregated in the same index record. It would offer a little less flexibility than I was aiming for, but it could reduce amount of records published to IPNI. But then again I'm not sure if optimizing for reduced number of IPNI records at expense of the query flexibility is the right tradeoff.

Gozala (commented):
Provided some feedback inline

rfc/indexing-design.md
```diff
@@ -0,0 +1,228 @@
...
+
+## Overview
+
+A user stores content as opaque blobs containing series of blocks addressed by multihashes. These blocks are collected into one or more CAR file (shards) and the CAR file is stored at a location designated for a particular user's data.
+
+The user creates a blob-index ([sharded-dag-index](https://github.com/w3s-project/specs/blob/main/w3-index.md#sharded-dag-index-example)) that contains the multihash of each shard of blob data, and an array of tuples of (multihash, offset, length) for all the blocks in each shard. The user may choose which blocks to include in the blob-index if they do not want all blocks to be indexed. For example, they may want only the DAG root to be indexed.
+
+The blob-index is stored in the user's space along with the CAR file(s) containing the blob data. This is done in the [w3 add-blob operation](https://github.com/w3s-project/specs/blob/main/w3-blob.md#add-blob). The w3s service creates [location commitments](https://github.com/w3s-project/specs/blob/main/w3-blob.md#location-commitment) for the blob shards and for the blob-index.
```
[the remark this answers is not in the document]
gammazero: One reason to keep them separate is because the index describes all shards, and shards may be stored in separate locations. If only the 3rd shard is needed then only the index and that shard need to be read.

It may be acceptable to encode the index as part of the first shard... but then how does the client know how much of the first shard to read? Would there be a fixed-length header?

rfc/indexing-design.md
```diff
@@ -0,0 +1,228 @@
...
+
+Finally the user uses the shard location, with range information from the blob index, to retrieve a specific piece of data from w3s.
+
+## Index Query
+
+A w3s client can query indexed information by specific attributes, and can filter retrieved query results.
+
+The queryable attributes are as follows.
+- Site: the location(s) where all blob data is available
+- Shard: the location(s) where a specific shard of data is available
```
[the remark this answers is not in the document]
gammazero: `Site`: The idea is that the query can specify "where site is in [X]" so that query response will only contain results that have something stored at a site in [X].

`Shard`: The idea is that a query can specify to retrieve results for requested shards, i.e. "where shard in [S]"

I will try to clarify.

rfc/indexing-design.md
```diff
@@ -0,0 +1,228 @@
...
+
+A user can use a multihash, that identifies data stored on w3s, to lookup the all information necessary to retrieve and delegate access to that data.
+
+Location commitments that describe where content is stored can change over time as w3s moves data to a more favorable location, re-aggregates data storage, or replicates stored data. Location commitments are updated without having to index all multihashes.
+
+A query API will allow a user to specify indexed content to lookup and how to filter results. The API retrieves previously cached responses or composes new IPNI queries to get new responses to filter and cache.
+
+## Overview
+
+A user stores content as opaque blobs containing series of blocks addressed by multihashes. These blocks are collected into one or more CAR file (shards) and the CAR file is stored at a location designated for a particular user's data.
```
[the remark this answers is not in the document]
hannahhoward: agree, cars just confuse things and are an implementation detail -- leave it out so we don't have to worry about this.

rfc/indexing-design.md
```diff
@@ -0,0 +1,228 @@
...
+
+The user creates a blob-index ([sharded-dag-index](https://github.com/w3s-project/specs/blob/main/w3-index.md#sharded-dag-index-example)) that contains the multihash of each shard of blob data, and an array of tuples of (multihash, offset, length) for all the blocks in each shard. The user may choose which blocks to include in the blob-index if they do not want all blocks to be indexed. For example, they may want only the DAG root to be indexed.
```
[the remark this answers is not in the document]
hannahhoward: +1

rfc/indexing-design.md
```diff
@@ -0,0 +1,228 @@
...
+Finally the user uses the shard location, with range information from the blob index, to retrieve a specific piece of data from w3s.
+
+## Index Query
+
+A w3s client can query indexed information by specific attributes, and can filter retrieved query results.
+
+The queryable attributes are as follows.
+- Site: the location(s) where all blob data is available
+- Shard: the location(s) where a specific shard of data is available
+- Slice: the location(s) within a shard where a block of data is available
```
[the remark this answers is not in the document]
hannahhoward: yea location will naturally be overloaded, but since we use location commitment as part of this design let's try to avoid the word elsewhere.

rfc/indexing-design.md
hannahhoward: +1

rfc/indexing-design.md
gammazero: There is some need to know what kind of data is being retrieved, not for creating an IPNI query, but to know what to get out of the blob-index indicated in the results. With a shard/slice query, the resulting blob-index must be read to get the CID/offset/length of the shard or slice. With a location query, the resulting blob-index must be read to get the shard location commitments.

Can this be generic when asking for different things?  This part probably requires  more understanding of the types of answers that a client will ask for, so we can determine if there is a generic way to do this.

rfc/indexing-design.md
hannahhoward: I would at minimum rename this `Indexing System Components`

Will make a rewording suggestion in my review

rfc/indexing-design.md
hannahhoward: Hopefully addressed by read-on-write -- if we're capturing negative responses we need to write the cache when we publish IPNI.

Because otherwise @Gozala makes a very good point and we've seen this exact problem with IPNI and Cloudflare caching -- 404s getting cached for 10 minutes if you query too soon.

rfc/indexing-design.md
hannahhoward: curious about a side issue here -- what do we do with two people who upload the same blob but publish different indexes? Seems like this would be relevant. Probably addressed if we add the "account paying" query param that the egress billing design calls for -- does make me wonder if we need to filter CID results not only on "are they W3UP" but also are "are they for the right person for this query?"

rfc/indexing-design.md
gammazero: Having cache-on-write fixes the case where new data is stored subsequent to a query for that data. When the new data is stored, or location info is updated, that would remove any negative cache. Without cache-on-write then a short TTL is necessary.  With cache-on-write, is there still a reason for a short TTL?

Negative cache entries should be kept in a separate cache so that a negative entry cannot evict a positive entry due to LRU, thereby removing the more valuable positive entries and allowing a misbehaving client to empty the cache.

Will clarify in doc.

rfc/indexing-design.md
```diff
@@ -0,0 +1,228 @@
...
+
+By using both these query results, a specific portion of the blob data can be retrieved. The shard(s) are read from the locations in the location commitments, and data ranges within each shard are requested (when not reading an entire shard) by range requests using the offsets and lengths from the shard/slice data.
+
+If either shard or location results are not cached, then a query for the multihash is passed to the IPNI client.
+
+#### Expiration Times
+The shard/slice data does not change over the lifetime of the stored data, so this data can have a much longer expiration time. It could also have no expiration time and be explicitly removed from cache when the data is no longer stored, but this means that there need to be a communication path from data removal to all cache locations. So, better to just have an expiration time, making cache more independent.
+
+### IPNI
+
```
hannahhoward, line 91:
```suggestion
IPNI is the final, "cache-miss" layer of the W3up Indexing System. IPNI is a highly optimized system for caching indexing information for massive numbers of CIDs. It's available to all and optimized to run on bare metal hardware. IPNI is used to do the following:
1. Provide a reliable global index for all w3up content, that is not only publicly available but usable by an IPFS retrieval client, without the other components of the w3up indexing system
2. Provide a low cost way to index content on w3up that is infrequently requested
```

hannahhoward: Want to clarify the role of IPNI here, and why I think it makes sense to even have it be part of this system. Gotten some questions from @Gozala and I want to clarify what I see as the unique add that IPNI gives us.

In short, the key values are:
1. Participation in the public network, where we are sharing data that can be read by regular IPFS retrieval clients (Lassie, ipfs.io, etc) and encouraging anyone who runs an IPNI instance to backup our content.
2. Low cost storage for infrequently accessed indexing data. A true "cache miss" (and to be clear, if we go to IPNI, we've already given up on fast path most likely, especially if IPNI shuts off its Cloudflare caching)

rfc/indexing-design.md
hannahhoward: I think this is not how IPNI is supposed to be used -- you can't throw the whole sharded dag index in the metadata and the metadata is constant across all the CIDS in the index. But I think it presents a question about whether the "IPNI cache" is just an IPNI cache or if we should actually store cached blob-indexes as well. I think we should and I THINK that's what the design doc is saying.

hannahhoward: I am not worried about the extra round trip on "cache miss" -- if we cache miss, the hot path is already over. (and to be clear, this is like a double cache miss -- we cache missed the whole retrieval being served from cloudflare cache, then when we setup the index query we cache missed the read of index data. Again, this should be infrequent content, if we outgrow the cache at all (there's a scenario where the cache at least for the IPNI query part as opposed to the blob-index part could be never-evict)

rfc/indexing-design.md
```diff
@@ -0,0 +1,228 @@
...
+The w3up index layer is responsible for reading a query specification, and then
+1. Creating a set of IPNI queries needed to get the requested index content
+2. Sending the queries to the w3up IPNI cache
+3. Filtering returned results
+4. Presenting results to the query client.
+
+### W3up IPNI Cache
+
+The w3up IPNI Cache is a cache that holds IPNI query results. The cache is:
```
hannahhoward, line 70: So this is my most substantive comment:

Is this an IPNI cache? Or is it a IPNI + sharded-dag-index cache?

As I understand it, IPNI results will ONLY contain a location to retrieve a Sharded Dag Index.

You'll need to read that sharded-dag-index / blob-index seperately (sidebar: I'm finding the distinction very difficult to understand -- can we have a terms section)

So are we caching that? I think the answer is yes, cause otherwise every query contains an extra round trip. I imagine these sharded-dag-indexes are evicted more often than the IPNI queries, since they're going to be larger. But I think to support a shard/slice query without blocking on an additional round trip, we should be caching these.

And if we're caching indexes, I think we should be explicit and not call this an IPNI cache.

hannahhoward (changes requested):
Overall, excellent. This is a really good design and I think we're close.

I have comments about queries I'll put in the query PR.

However, the big thing that needs clarification is the caching behavior for blob-index/sharded-dag-indexes -- to me, I think we need to cache these for super high value content, so that the fast path for a completely cached index query never has a read to an external storage provider (remember that eventually our storage nodes could have higher latency from the gateway). I'm open to debate here but it seems implied in various places.

I think this is basically ready to start implementing, but lets keep going a little more.

hannahhoward:
OK now I read @Gozala's PR and I get his confusion and how it relates to my general question about caching the blob-indexes, which would not be needed if we implement what he suggests, but it too me a couple reads to understand what I believe he's suggesting, which may actually be a good idea. We'll see!

I setup a quick sync to get us on the same page.

rfc/indexing-design.md
```diff
@@ -0,0 +1,228 @@
...
+Query results can be filtered to select only results that the user has access to, or for data supplied by specific data providers.
+
+### Query Specification
+
+```
+match = {
+  subject=[myID, myOrgID],
+  location=spBaseURL,
+}
+results = Query(multihash="bafy...", match)
```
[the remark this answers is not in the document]
gammazero: These are just subject identifiers (me, my organization). The query selects results that these subjects are authorized to read.

Location allows the client to select only results stored at some location, say s3.us-west-2... So, if there are any shards stored there, results are returned. If nothing is stored there, then no results.

hannahhoward (approved):

hannahhoward:
@alanshaw can you take a look at this when you have a chance?

@gammazero after Alan's review I'd like to promote this to the system architecture repo. We can continue to evolve as we go.

I think it would be good to capture the various outcomes of todays discussions as well. (written summary below)

---

1. The cache will translate both IPNI requests and claims bundle (i.e. sharded dag index) into a format optimized for fast querying of lots of data and supporting arbitrary relationships between multihashs. Likely some kind of EAV store
   1. There’ll be a ticket to design this data model, and maybe another to start protoyping.
1. The claims bundle / sharded dag index effectively captures 2 types of relationships:
   1. blob/shard (how is the blob broken up into multiple shards)
   1. blob/slice (what are the offsets of individual blocks or groups of blocks in the the shard)
   1. in the future, it’s possible there are other types of relationships we will want to store. we should be able to add additional relationships and potentially publish them to IPNI without redesigning the system
1. For IPNI:
   1. You can query by a shard multihash to get the location commitments for that shard. Each location commitment for each shard will have a contextID as they update independently.
   1. You can query by any multihash in DAG to get the address of the claims bundle for the DAG. These will all updated under a single context ID. Importantly, each relationship will have a separate metadata protocol. There are a number of ways to do this on a single context ID:
      1. Actually, rereading spec and https://github.com/ipni/go-libipni/blob/main/metadata/metadata.go, a single “Metadata” as specified by IPNI is MULTIPLE metadata protocols — so Andrew I believe this can be accomplished by just including multiple metadata protocols in the metadata for the main advertisement for the context ID. Metadata can also be updated in a future advertisement, so you could add future relations this way.
      1. That said we could also use extended providers here and maybe it would be useful to do so. Mostly if different bundles start ending up at different publishers
      1. Anyway, there’s going to a ticket for laser focus on the metadata and IPNI ad structures so we can figure it out then.
   1. Worth noting we still need to keep our existing advertisement chain going, cause this is what ipfs.io and other IPFS clients know how to use for now.
1. For the Index query:
   1. I’m ok with a query interface instead of a rest api for the long term. I get the rationale. I honestly think the first version should be a REST Api for the one query join we actually want. We can certainly build the query interface right away, but I would not build a generalized query parser and implementation. Instead I would just pattern match the queries against the specific joins we want to do for the moment and error on everything other than the thing we need. Again, for the moment.
   1. There’ll be a prototype ticket for this service.

rfc/indexing-design.md
```diff
@@ -0,0 +1,256 @@
...
+3. Filtering returned results
+4. Packaging results for the query client.
+
+### W3up IPNI and Index Cache
+
+The w3up IPNI Cache is a cache that holds IPNI query results and sharded-dag-index data. The cache is:
+
+- Limited-size LRU. Discards the least recently used data when at storage capacity.
+- Temporary. Evicts items that have been cached for longer than the cache expiration time, regardless of last access. This allows changes to query results to be seen.
+- Negative (empty) result-aware. Caching empty results prevents more expensive IPNI queries for data that is not indexed. Negative cache entries are kept in a separate cache so that negative entries cannot evict a positive entrries due to LRU.
```
alanshaw, line 74:
```suggestion
- Negative (empty) result-aware. Caching empty results prevents more expensive IPNI queries for data that is not indexed. Negative cache entries are kept in a separate cache so that negative entries cannot evict positive entries due to LRU.
```

I'd also give them a significantly smaller cache time.

rfc/indexing-design.md
```diff
@@ -0,0 +1,256 @@
...
+- Populated on write: When new or updated data is publisher to IPNI, the cache is populated. This replaces previous any previous cache entries, including negative ones.
```
alanshaw, line 75:
```suggestion
- Populated on write: When new or updated data is published to IPNI, the cache is populated. This replaces any previous cache entries, including negative ones.
```

alanshaw: I would invalidate. Send the data to where you want it and invalidate known caches (delete) so the latest can be fetched from source. It'll be quicker to publish.

rfc/indexing-design.md
```diff
@@ -0,0 +1,256 @@
...
+
+Similar to negative cache entries, cached sharded-dag-indexes are kept in a separate LUR cache because their size will cause them to be evicted more frequently than IPNI results.
```
alanshaw, line 77:
```suggestion
Similar to negative cache entries, cached sharded-dag-indexes are kept in a separate LRU cache because their size will cause them to be evicted more frequently than IPNI results.
```

This one does not need to be time based.

rfc/indexing-design.md
alanshaw: I'm not sure how smart this query/cache layer should be. I'd be more inclined to have it only allow querying, doing the joins and caching the result.

alanshaw (changes requested):

rfc/indexing-design.md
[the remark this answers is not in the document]
gammazero: Removing location.

PR #30, opened 2024-05-29
feat: context free infra

Gozala:
Outlining idea of reducing complexity caused by vast context currently used by a service.

### DIFF
--- a/rfc/context-free-infra.md
+++ b/rfc/context-free-infra.md
@@ -1,0 +1,33 @@
+# Context Free Infra
+
+## Background
+
+Every time I have to do deal with [w3infra] I get overwhelemed by immense context passed across capability providers. There is a significant amount of bootstrap code and a lot of configuration settings. While it configuration offers flexibility, it also adds a lot of complexity which can be a maintanance burden.
+
+## Proposal
+
+Desipite large context, list of primitives used is pretty short:
+
+1. Object Store (s3 / r2)
+2. Data Index (Dynamo DB / s3 / r2)
+3. Queues (SQS, Kenisis)
+4. Email
+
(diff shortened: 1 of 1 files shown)


travis:
Am I correct in thinking that this proposes essentially wrapping dynamo/s3/etc in UCAN invocation APIs? 

I'm curious if we could accomplish what you want by just creating normal TypeScript interfaces rather than UCAN Invocations and taking a little more care to hide the dependencies of the TypeScript interfaces from calling code... I can see advantages to having a nice set of UCAN invocation interfaces, but not entirely sure it's worth the complexity of trying to get UCAN Invocations in the middle of APIs that tend to do a lot of streaming (something that I haven't seen UCAN Invocations do well)...

I like the high level idea and I think the core problem is something we should work on, just a little nervous about UCAN Invocations as the right way to define this abstraction!

rfc/context-free-infra.md
```diff
@@ -0,0 +1,33 @@
...
+## Proposal
+
+Desipite large context, list of primitives used is pretty short:
+
+1. Object Store (s3 / r2)
+2. Data Index (Dynamo DB / s3 / r2)
+3. Queues (SQS, Kenisis)
+4. Email
+
+We could greatly simplify our infrastructure if we simply exposed APIs given access to the listed primitives in terms of privileged ucan capabilities, that is capabilities that only server authorized entity can invoke. It is worth calling out that all of these primitives provide scoping mechanims object stores have buckets, indexes have tables, queues have groups or named streams.
```
vasco-santos, line 16: In practise permissions are defined at the AWS IAM level, so while we can have a second layer of auth it does not solve auth all together.

vasco-santos (commented):
I highly agree with Travis. Maybe a more concrete example would be easier to understand, but so far I would actually feel more complexity and design work for a fairly simple thing. 

Also worth noting that this is not a w3infra thing, but the `upload-api` server context, you can see exactly same context needing to be created on the tests there.

Also as @travis pointed out, I would see an opportunity to simplify things at the Typescript definitions level, or perhaps rethink how we construct the context. But having a UCAN protocol just to communicate with these resources seems a lot of overhead for me, both in terms of more code to maintain and more IO operations to do

PR #31, opened 2024-06-03
rfc: bulk downloader

vasco-santos:


bulk-downloader.md
```diff
@@ -0,0 +1,79 @@
...
+
+This project targets a completely different goal than w3up migration tool. Here the goal would be to actually download the entire content previously uploaded by an account. This solution is technically challenging, as well as expensive for the service operator. In addition, it comes at a cost of losing these users. 
+
+Considering the initial plan to provide a bulk download URL, there are a set of drawbacks and concerns that should be taken into account, specially when users have hundreds of GBs on their accounts:
+
+- Not fault tolerant to errors
+    - network errors no recoverable
+    - no more space on disk
+- Not verifiable
+    - we would not be served content addressed data that can be verified, but a compacted file with all the data
```
gammazero, line 40: Could offset, range, hash data be provided to allow content verification? Would there be a sufficient use-case for that?

PR #33, opened 2024-07-20
Update github migrations for nucleation

hannahhoward:
# Goals

Given that web3-storage github org will no longer be owned by Storacha team, it seemed like there were a lot more repos that needed migrating, so I migrated them. This tracks all of those migrations. I'd like a review -- we can always transfer back, but this reflects current state. Is anything missing? Anything that should be moved back?

### DIFF
--- a/rfc/github-org-migration.md
+++ b/rfc/github-org-migration.md
@@ -79,7 +79,7 @@ The [Web3 Storage](https://github.com/web3-storage) SHOULD be downgraded to the 
 
 - [ ] [ipfs-car](https://github.com/web3-storage/ipfs-car)
     - CAR client for UnixFS dags. It is heavilty used.
-    - It SHOULD be migrated.
+    - It SHALL be migrated.
 - [ ] [web3.storage](https://github.com/web3-storage/web3.storage)
     - Monorepo for the old website. It deploys to Cloudflare Pages.
     - It SHOULD NOT be migrated.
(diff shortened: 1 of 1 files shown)


rfc/github-org-migration.md
```diff
@@ -95,10 +95,10 @@ The [Web3 Storage](https://github.com/web3-storage) SHOULD be downgraded to the
     - It SHOULD NOT be migrated.
 - [ ] [add-to-web3](https://github.com/web3-storage/add-to-web3)
     - Github action to put websites on web3.storage.It has probably been updated.
-    - It SHOULD be migrated.
+    - It SHALL be migrated.
 - [ ] [files-from-path](https://github.com/web3-storage/files-from-path)
     - A lib for globbing a bunch of files in a dir and returning that as an array of file objects.
-    - It SHOULD be migrated.
+    - It SHALL be migrated.
```
alanshaw, line 101: Still seems to be in the web3-storage org

alanshaw (changes requested):
I went over the remaining repos in the org and saw these that also need migrating:

* `learnyouw3up` (a nodeschool style workshopper) we might want to update at some point
* `content-claims-gql` (GraphQL interface to content-claims) we might want to update
* `linkdex` (npm module to determine if a graph is complete) we use this currently
* `w3link` (w3s.link gateway worker) I thought this moved to `reads` but perhaps not - probably best to migrate
* `ipns-publisher` (w3name firehose consumer to DHT publisher via Kubo) is currently used by w3name

hannahhoward:
updated to include repos you mention

alanshaw (approved):

PR #34, opened 2024-08-20
Proposed curio api

hannahhoward:
Proposes an HTTP API to curio that could be used by either Storacha or another hot storage market provider for PDP proofs

[Preview](https://github.com/storacha-network/RFC/blob/feat/curio-suggested-api/rfc/curio-api-rfc.md)

### DIFF
--- a/rfc/curio-api-rfc.md
+++ b/rfc/curio-api-rfc.md
@@ -1,0 +1,133 @@
+# Curio API RFC
+
+These are some suggested endpoints for collaboration between Curio and hot storage market software such as Storacha.
+
+## Piece Storage & Retrieval
+
+These endpoints are simply used for putting arbitrary blobs of bytes that has to a PieceCID of any valid size (i.e. 127/128 * 2^n where n is a positive whole number)
+
+### GET /piece/{piece cid v2}
+### HEAD /piece/{piece cid v2 }
+
+These can just follow FRC-066: https://github.com/filecoin-project/FIPs/blob/master/FRCs/frc-0066.md (please include the Range parameter!)
+
+With the additional stipulation that they should probably ONLY accept v2 Piece CID: https://github.com/filecoin-project/FIPs/blob/master/FRCs/frc-0069.md
+
+### PUT /piece/{piece cid v2}
+
+This is a new API for storing pieces of arbitrary size into Curio's storage. The request body should contain the bytes for the piece. These should be checked internally to verify they match the piece cid specified on put.
+
+TBD: Should a header or query param specify that you intend to use this for PDP proofs later? (instructing curio to cache a proof tree)
+
+Returns:
+200/204 When piece is already there
+201 When a new piece is created
+400 when piece data does not hash to piece cid v2
+
+## PDP Proving
+
+### Approach #1 - Minimal, just prove within a piece
+
+In this approach, we leave things very simple on curio's side, leave most lotus interaction to the market software and curio simply provides proofs within a piece.
+
+This means that market software would need to:
+1. Manage proof sets on chain (via lotus)
+2. Monitor proving epochs
+3. Communicate with DRAND
+4. Produces challenges for a given seed
+5. Translate challenge offsets within an overall proof set as a logical data array to challenge offsets within a single piece
+6. Reassemble challenges against roots and submit the overall proof to the chain
+
+### POST /prove
+
+Request body:
+```json
+{
+  "pieceCid": "~piece~cid~v2~previously~posted~",
+  "offset": "~offset~in~piece~for~this~challenge~"
+}
+```
+
+Response:
+```json
+{
+  "merkleProof": { 
+    // not sure what format this takes
+  }
+}
+```
+
+### Approach #2 -- Curio is a full PDP provider
+
+In this scenario, almost all PDP work is managed by curio itself (which would then need alternate software when running without curio). Essentially curio would provide APIs that mirrored the Create/Add/Remove/Delete functions that will exist on chain for PDP proof sets, and then manage the submission of proofs on a schedule (I hear it's good at scheduling)
+
+### POST /proof-sets
+
+Request Body:
+```json
+{
+  // should this be passed to curio? or is owner address built in to curio already?
+  // should curio check it has signing capabilities for the address?
+   "ownerAddress": "f3...",
+   "challengePeriod": 15
+}
+```
+
+Response:
+Code: 201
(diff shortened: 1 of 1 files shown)


rfc/curio-api-rfc.md
```diff
@@ -0,0 +1,133 @@
...
+These endpoints are simply used for putting arbitrary blobs of bytes that has to a PieceCID of any valid size (i.e. 127/128 * 2^n where n is a positive whole number)
+
+### GET /piece/{piece cid v2}
+### HEAD /piece/{piece cid v2 }
+
+These can just follow FRC-066: https://github.com/filecoin-project/FIPs/blob/master/FRCs/frc-0066.md (please include the Range parameter!)
+
+With the additional stipulation that they should probably ONLY accept v2 Piece CID: https://github.com/filecoin-project/FIPs/blob/master/FRCs/frc-0069.md
+
+### PUT /piece/{piece cid v2}
```
magik6k, line 16: Might require a UCAN header for auth

magik6k: Also for scalability it would be a good idea to have the main endpoint tell you where to go with the data.

Unfortunately redirecting PUT/POST with 3xx is really wonky (pretty much broken with the Go http.Client), so it might just require some two-endpoint setup.

rfc/curio-api-rfc.md
```diff
@@ -0,0 +1,133 @@
...
+}
+```
+
+Response:
+Code: 201
+Location header: "/proof-set/{set-id}"
+
+*TODO: do we need an interim response given this is a chain transaction with a place to fetch the set-id later?*
+
+### GET /proof-sets/{set-id}
```
magik6k, line 82: Maybe link to some spec which lays out how those proof-sets look like / what they are

hannahhoward: Truth be told I'm just betting based off the PDP service contract doc

rfc/curio-api-rfc.md
```diff
@@ -0,0 +1,157 @@
...
+1. Endpoints for manipulating proofs sets
+2. An endpoint for storing pieces + flow for storing pieces
+3. An endpoint for retrieving pieces
+
+There are addition considerations we should consider:
+1. Authorization -- In storacha's network, it's important that the original end user maintain control of authorization for any action performed (including retrieval). We accomplish this through UCANs. We should discuss how we can maintain this without forcing curio to implement a full UCAN authorization process.
+2. Aggregation - storacha's data is at times extremely small (<1mb in certain cases). Our understanding is that economically, it makes more sense to do some light aggregation of data before adding it to the proof set. The proposal below outlines a facility for doing this. While storacha would store pieces as it receives them, we would add them to the proof set in a seperate step, with a root that could optionally be an aggregate of several pieces.
+3. IPNI announcements -- we plan to use IPNI announcements in a specific way with our pieces. Our understanding is the curio IPNI flow is in flux. We can try to integrate your IPNI api or just do it ourselves.
```
magik6k, line 13: IPNI is pretty much implemented now, with the following schema that coordinates it on the curio side:

```
-- Table for storing IPNI ads
CREATE TABLE ipni (
    order_number BIGSERIAL PRIMARY KEY, -- Unique increasing order number
    ad_cid TEXT NOT NULL,
    context_id BYTEA NOT NULL, -- abi.PieceInfo in Curio
    -- metadata column in not required as Curio only supports one type of metadata(HTTP)
    is_rm BOOLEAN NOT NULL,

    previous TEXT, -- previous ad will only be null for first ad in chain

    provider TEXT NOT NULL, -- peerID from libp2p, this is main identifier on IPNI side
    addresses TEXT NOT NULL, -- HTTP retrieval server addresses

    signature BYTEA NOT NULL,
    entries TEXT NOT NULL, -- CID of first link in entry chain

    unique (ad_cid)
);

CREATE TABLE ipni_head (
    provider TEXT NOT NULL PRIMARY KEY, -- PeerID from libp2p, this is the main identifier
    head TEXT NOT NULL, -- ad_cid from the ipni table, representing the head of the ad chain

    FOREIGN KEY (head) REFERENCES ipni(ad_cid) ON DELETE RESTRICT -- Prevents deletion if it's referenced
);

-- This table stores metadata for ipni ad entry chunks. This metadata is used to reconstruct the original ad entry from
-- on-disk .car block headers or from data in the piece index database.
CREATE TABLE ipni_chunks (
    cid TEXT PRIMARY KEY, -- CID of the chunk
    piece_cid TEXT NOT NULL, -- Related Piece CID
    chunk_num INTEGER NOT NULL, -- Chunk number within the piece. Chunk 0 has no "next" link.
    first_cid TEXT, -- In case of db-based chunks, the CID of the first cid in the chunk
    start_offset BIGINT, -- In case of .car-based chunks, the offset in the .car file where the chunk starts
    num_blocks BIGINT NOT NULL, -- Number of blocks in the chunk
    from_car BOOLEAN NOT NULL, -- Whether the chunk is from a .car file or from the database
    CHECK (
        (from_car = FALSE AND first_cid IS NOT NULL AND start_offset IS NULL) OR
        (from_car = TRUE AND first_cid IS NULL AND start_offset IS NOT NULL)
    ),

    UNIQUE (piece_cid, chunk_num)
);
```

Now, IPNI likes larger ads, so ideall storacha would create aggregate ads for multiple pieces; we can extend `ipni_chunk` to support reading from storacha-stored pieces (though really technically just the piececid works fine there)

rfc/curio-api-rfc.md
```diff
@@ -0,0 +1,157 @@
...
+
+### DEL /proof-sets/{set id}
+
+Remove the specified proof set entirely
+
+
+## Piece Storage
+
+
+### POST /piece
```
magik6k, line 133: We definitely need to define how authorization works on this endpoint. This can't just be entirely open.

Also should define the lifecycle of the uploaded data somehow:
* How long is it expected to stick around in storage after upload before being included in a proof-set? When should the data be removed if not added to a proof set?
  * Signalling for expected indexing with IPNI / ipfs-type (trustless gateway/bitswap) retrievals, and who can retrieve the piece?
* What is the contract for retrieval - is it retrievable atomically when the notify hook is called? After inclusion in a proof set?

magik6k: Also what are the size bounds for pieces that you expect curio to support? We can support even very large pieces (100G+), but I don't think a client-push model is a good idea above 1GB, where managing short-term buffers becomes a real concern, and download retry becomes non-optional

PR #35, opened 2024-08-29
rfc: big data from the gateways

alanshaw:
This RFC describes a change we could make to existing infrastructure that should allow big files to be served from the `w3s.link` gateway.

### DIFF
--- a/rfc/big-data-from-the-gateway.md
+++ b/rfc/big-data-from-the-gateway.md
@@ -1,0 +1,20 @@
+# Big Data from the Gateway
+
+## Authors
+
+- [Alan Shaw](https://github.com/alanshaw), [Storacha Network](https://storacha.network/)
+
+## Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC2119](https://datatracker.ietf.org/doc/html/rfc2119).
+
+## Goal
+
+Enable serving big UnixFS encoded files from the gateway.
+
+## Background
+
+The IPFS Gateway `w3s.link` (referred to as "the gateway" subsequently) is currently run by the Storacha Network team. It receives requests for a CID (plus an optional path), assembles UnixFS encoded DAGs into files and sends the resulting bytes back to the client. It also has other functionalities but these are not relevant to this RFC.
+
+The gateway is unique in that it does not run a full IPFS node (e.g. Kubo) and does not operate over bitswap. It discovers the location (a URL and other metadata) of blocks via content claims and reads blocks by making HTTP GET requests using byte range headers to extract the exact bytes required.
+
--- a/rfc/big-data-from-the-gateway.md
+++ b/rfc/big-data-from-the-gateway.md
@@ -14,7 +14,40 @@ Enable serving big UnixFS encoded files from the gateway.
 
 ## Background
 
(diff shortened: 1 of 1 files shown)


rfc/big-data-from-the-gateway.md
```diff
@@ -0,0 +1,53 @@
...
+## Proposal
+
+To support larger file transfers we can make a small architecture change and add a worker in front of the current setup. This worker has the following responsibilities:
+
+1. Resolve the CID+path.
+2. Decode and inspect the root block to ascertain the size of the file.
+3. Make byte ranged sub-requests to the existing infrastructure to extract data required.
+4. Combine sub-request streams into a single response stream.
+
+Note: The existing infrastructure MUST be augmented to allow byte range requests to UnixFS files as this is not currently supported.
```
hannahhoward, line 44: Ouch I didn't realize we never implemented byte ranges

rfc/big-data-from-the-gateway.md
```diff
@@ -0,0 +1,53 @@
...
+
+The thesis is that the worker orchestrating the file transfer will use significantly less CPU time than workers assembling the DAG and hence should be able to transfer much larger files.
```
hannahhoward, line 46: how is CPU measured here -- is it raw time? CPU speed? Would it be constrained potentially on network transfer speed?

i.e. transfering say 10s of GBs would definitely take potentially minutes -- if we're effectively consuming CPU time the whole way, then I'm not sure this would work.

hannahhoward: my recommendation is we should talk to cloudflare first

hannahhoward (approved):
So I had assumed we'd implement roughly this approach but instead of another worker we'd just write a client that does this. Honestly, it's probably effectively almost the same software.

I would recommend a staged approach where we do the client first and then see if it can get working in a Worker. Of particular note is that I'm not convinced the central thesis will apply -- that we'll be able to keep CPU time below 30s. I put some rationale for this, but honestly, this strikes me as something we just need direct conversations with cloudflare about. it's worth noting that CF specifically says transferring large media from workers is not supported and is a license violation in their professional plans, so I'd also like to understand what conversations we've had with them about an exception on this for our enterprise plan (the internet seems to think they make them for enterprise plans). If that is the case, is it possible they'd just let us extend CPU time as well?

Fundamentally though, this is the approach I imagined -- byte range requests across a large file -- but good to know we need to add stuff to DAGULA for that (I thought we supported)

rfc/big-data-from-the-gateway.md
alanshaw: It's implemented in the JS unixfs exporter and should be easy enough to hook up. We're never had significant customer demand so it didn't become a priority.

rfc/big-data-from-the-gateway.md
alanshaw: AFAIK there is little/no CPU time expended proxying data and this is what the RFC hinges on.

alanshaw:
> So I had assumed we'd implement roughly this approach but instead of another worker we'd just write a client that does this. Honestly, it's probably effectively almost the same software.

Yes, probably!

> I would recommend a staged approach where we do the client first and then see if it can get working in a Worker. 

Sure - I think implementing byte range requests on the gateway is valuable anyway.

Like you mentioned the code to do this client side will be almost identical to what we end up doing server side.

> If that is the case, is it possible they'd just let us extend CPU time as well?

We can ask! That would be easy but flagging that it may not be possible, I imagine comes with a cost, and turnaround might be a while.

I can probably prove the thesis in a few hours...which may be faster.

> Fundamentally though, this is the approach I imagined -- byte range requests across a large file -- but good to know we need to add stuff to DAGULA for that (I thought we supported)

We support byte ranges for CAR responses (with merkle proof) but not regular files. Like I said though the JS unixfs exporter does support it and it should just be a case of wiring the params.

PR #36, opened 2024-10-07
Egress, authorized by UCAN

Peeja:
[📚 Preview](https://github.com/storacha/RFC/blob/egress-with-ucan/rfc/egress-with-ucan.md)

A proposal for managing billable egress using UCAN, distilled from a conversation between @hannahhoward, @travis, and myself (@Peeja) on 10/04/24. Builds on and evolves prior egress proposals.

Open questions and missing bits are currently in the document, at the bottom.

### DIFF
--- a/rfc/egress-with-ucan.md
+++ b/rfc/egress-with-ucan.md
@@ -22,54 +22,69 @@ To make content available as on a traditional CDN, an Agent acts on authority of
 
 This process does *not* include making and tracking Location Commitments. A Location Commitment is an attestation by a Storage Node that it holds a particular piece of content on behalf of a particular Space, and that it can provide it. In this proposal, we assume such a system already exists.
 
-## Amend command *(ability)*: Get Blob
+## New command *(ability)*: `/space/content/serve` *(`space/content/serve`)*
 
-We will modify the [Get Blob](https://github.com/storacha/specs/blob/main/w3-blob.md#get-blob) UCAN command *(or in UCAN 0.9, ability)*.
+We add a new command *(UCAN 0.9: ability)* to serve content, named `/space/content/serve` *(`space/content/serve`)*. This command is used to validate the authority to serve content. Notably, it will not (yet) actually *serve* the content, though it may in the future. That is, a successful response will not contain the means to get the content. The Invoker of this command will (for now, at least) always be an entity which *can* get any content in the network (such as the Gateway), but would like to determine whether it *should*.
 
-First, we update the original spec to reality. Where the spec names the command `/space/content/get/blob/0/1`, we have implemented it in UCAN 0.9 as `space/blob/get/0/1`. We will continue to use that name, and use that name in this document, as it provides a better authorization hierarchy—specifically, the Space can grant a Principal `space/blob/*`.
+In the future, this could be invoked from external clients to actually get content, in essence porting the [Gateway spec](https://specs.ipfs.tech/http-gateways/path-gateway/) to UCAN.
 
-Second, we will add an additional argument, `token`. This represents the bearer token when invoked through the Gateway, or something similar. It can be `null` or missing, and will be in existing uses of the command.
-
-Note that the command has a `/0/1` suffix. This appears to be included to mark it as unstable. This proposal does not include removing the suffix, though that will likely make sense to do soon.
+### Invocation
 
 ```ts
-type GetBlob = {
-  /** "Retrieve details about a Blob from a Space." */
-  cmd: "/space/blob/get/0/1" // (0.9: "space/blob/get/0/1")
+type ServeContent = {
+  /**
+   * "Serve content owned by the subject Space."
+   * 
+   * A Principal who may `/space/content/serve` is permitted to serve any
+   * content owned by the Space, in the manner of an [IPFS Gateway]. The
+   * content may be a Blob stored by a Storage Node, or indexed content stored
+   * within such Blobs (ie, Shards).
+   * 
+   * Note that the args do not currently specify *what* content should be
+   * served. Invoking this command does not currently *serve* the content in
+   * any way, but merely validates the authority to do so. Currently, the
+   * entirety of a Space must use the same authorization, thus the content does
+   * not need to be identified. In the future, this command may refer directly
+   * to a piece of content by CID.
+   * 
+   * [IPFS Gateway]: https://specs.ipfs.tech/http-gateways/path-gateway/
+   */
+  cmd: "/space/content/serve" // (0.9: "space/content/serve")
 
   /**
-   * The Space which contains the Blob. This Space will be charged egress fees
-   * if the Blob data is actually retrieved by way of this invocation.
+   * The Space which contains the content. This Space will be charged egress
+   * fees if content is actually retrieved by way of this invocation.
    */
   sub: SpaceDID
 
   args: {
-    /** The multihash digest of the Blob to look up. */
-    digest: Uint8Array
-
-    /** The token, if any, used for this request. */
+    /** The authorization token, if any, used for this request. */
     token?: string | null
   }
 }
 ```
 
-The receipt type remains [unchanged](https://github.com/storacha/specs/blob/main/w3-blob.md#get-blob-receipt).
+### Result
 
-### Existing uses
+```ts
+type ServeContentResult = Result<{}, ServeContentError>
 
-The existing use of this command is to read metadata about the Blob, specifically the size. For instance the Console uses this to display the size of a file Shard in the UI. This use remains unchanged. The `token` should be left missing, or set to `null`.
+type ServeContentError = {
+  message: string
+}
+```
 
 ## Making content retrievable with a token
 
-The Gateway itself will be authorized to `/space/blob/get/0/1` by delegation. This signals that the Gateway may serve the data. Typically, this delegation is performed by an Agent within a Storacha Client, who has received the authority to `/space/blob/get/0/1` from the Space by way of logging into an Account—thus, the delegation chain flows Space → Account → Agent → Gateway.
+The Gateway itself will be authorized to `/space/content/serve` by delegation. This signals that the Gateway may serve the data. Typically, this delegation is performed by an Agent within a Storacha Client, who has received the authority to `/space/content/serve` from the Space by way of logging into an Account—thus, the delegation chain flows Space → Account → Agent → Gateway.
 
 ```json5
 // UCAN 1.0
 {
   "iss": "did:key:zAgent",
-  "aud": "did:web:ipfs.w3s.link",
+  "aud": "did:web:w3s.link",
   "sub": "did:key:zSpace",
-  "cmd": "/space/blob/get/0/1",
+  "cmd": "/space/content/serve",
   "pol": [
     ["==", ".token", "abc123def456"],
   ],
@@ -82,11 +97,11 @@ The Gateway itself will be authorized to `/space/blob/get/0/1` by delegation. Th
 // UCAN 0.9
 {
   "iss": "did:key:zAgent",
-  "aud": "did:web:ipfs.w3s.link",
+  "aud": "did:web:w3s.link",
   "att": [
     {
       "with": "did:key:zSpace",
-      "can": "space/blob/get/0/1",
+      "can": "space/content/serve",
       "nb": {
         "token": "abc123def456"
       },
@@ -98,10 +113,6 @@ The Gateway itself will be authorized to `/space/blob/get/0/1` by delegation. Th
 }
 ```
 
-Such delegations SHOULD NOT specify a `digest`, as this will be the digest of a Shard Blob, not the file which it contains part or all of. We are not (currently) supporting a use case in which the files in a Space have different access permissions from one another: all permissions should apply to an entire Space.
-
-The UCAN 1.0 version of these delegations SHOULD specify a top-level policy of the form `["==", ".token", <string>]`. This prevents customers from shooting themselves in the foot by getting too fancy with their tokens, which is not a supported use case, or one we see value in. Such use cases may be considered for support in the future.
-
 ## Making content retrievable *without* a token
 
 The customer may also authorize the Gateway to serve requests with no token. Separately, the Gateway will treat these requests differently: it will rate-limit them and not charge for egress. However, that has no bearing on the UCAN semantics of the delegation.
@@ -110,9 +121,9 @@ The customer may also authorize the Gateway to serve requests with no token. Sep
 // UCAN 1.0
 {
   "iss": "did:key:zAgent",
-  "aud": "did:web:ipfs.w3s.link",
+  "aud": "did:web:w3s.link",
   "sub": "did:key:zSpace",
-  "cmd": "/space/blob/get/0/1",
+  "cmd": "/space/content/serve",
   "pol": [
     ["==", ".token", null],
   ],
@@ -125,11 +136,11 @@ The customer may also authorize the Gateway to serve requests with no token. Sep
 // UCAN 0.9
 {
   "iss": "did:key:zAgent",
-  "aud": "did:web:ipfs.w3s.link",
+  "aud": "did:web:w3s.link",
   "att": [
     {
       "with": "did:key:zSpace",
-      "can": "space/blob/get/0/1",
+      "can": "space/content/serve",
       "nb": {
         "token": null
       },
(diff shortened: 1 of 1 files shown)


rfc/egress-with-ucan.md
```diff
@@ -0,0 +1,159 @@
...
+
+1. Look up the Location Commitment for the given CID. If not found, respond with 404 Not Found.
+2. If there is no token given, serve the request through the standard rate-limiter. (This behavior is likely to change in the future.)
+3. If there is a token given (eg, `abcde12345`), build its corresponding DID (eg, `did:bearer:abcde12345`).
+4. Look up all delegations with the token DID as audience.
+5. Attempt to prove the ability to invoke `/space/content/retrieve` on the Space listed in Location Commitment, with the given CID. If more than one Location Commitment is found, attempt each in turn: a CID may be stored in multiple Spaces, and the token may be able to retrieve the content through one and not another.
+6. If no such proof chain is available, respond with 401 Unauthorized. [or 404 Not Found?]
+7. If a proof chain is found, execute the invocation on the Executor.
+8. Using the information in the receipt, fetch the content and proxy it as the response.
```
fforbeck, line 136: question: How will the customer ID be identified to log the egress event in the Accounting Service? At which step in this process will the customer ID be retrieved and associated with the request for accurate billing?

Peeja: The billing is tied to the Space, so once we have the Location Commitment (which specifies the Space), we can bill the right customer. That probably deserves a callout in here.

fforbeck: Got it. Thanks. Yes, it would be great to have that stated in the RFC as well.

rfc/egress-with-ucan.md
```diff
@@ -0,0 +1,159 @@
...
+
+## [To Come]
+
+* Rather than serve non-token content rate-limited by default, require a delegation of `/space/content/retrieve` to some DID representing "anyone".
+* Bitswap should execute a `/space/content/retrieve` as well to respond to requests. Bitswap should be authorized by delegating to some DID representing Bitswap/Hoverboard.
+
+## Open questions
+
+* What does the `/space/content/retrieve` receipt look like?
+* Can you enumerate the contents of a space? Is that in scope here?
```
travis, line 157: **thought**: we do have [`upload/list`](https://github.com/storacha/w3up/blob/main/packages/capabilities/src/upload.js#L138) and [`space/blob/list`](https://github.com/storacha/w3up/blob/main/packages/capabilities/src/blob.js#L109) for this - I don't think we need to worry about it here as the gateway doesn't expose any way to do this

rfc/egress-with-ucan.md
```diff
@@ -0,0 +1,159 @@
...
+* What does a Gateway URLs with a token look like?
```
travis, line 158: **suggestion**: I'd probably go with something like `https://<cid>.ipfs.w3s.link?auth=<token>` - note that the [current implementation](https://github.com/storacha/freeway/blob/9444729e959f4a52e991bddfdda14479885c42e7/src/handlers/rate-limiter.js#L83) actually looks in the `Authorization` header and will need to be extended to look in the query as well - imho both are important for different use-cases but the query is probably most important for our current efforts

rfc/egress-with-ucan.md
```diff
@@ -0,0 +1,159 @@
...
+* We eventually want to allow Storage Nodes to offer their own direct egress, bypassing our Gateway, and to charge the customer according to an agreement with them.
+
+## Abstract
+
+A proposed means for offering content egress from the Storacha Network which can be selectively authorized by UCAN, and which can track egress to bill appropriate customers. As usual within the Storacha Network, we begin from self-sovereignty: a Space definitionally has the authority to do anything with its content.
+
+To make content available as on a traditional CDN, an Agent acts on authority of a Space to make that Space's contents, or some specific CID in it, available using a bearer token, an opaque, unguessable string. The Gateway then responds to requests which contain a token by validating the proof chain, finding the content, charging for egress, and proxying it to the requester.
+
+This process does *not* include making and tracking Location Commitments. A Location Commitment is an attestation by a Storage Node that it holds a particular piece of content on behalf of a particular Space, and that it can provide it. In this proposal, we assume such a system already exists.
```
alanshaw, line 16: One interesting thing to consider is that if the location commitment were to be cited in a `space/content/retrieve` delegation, in the happy path (where the content has not been moved), the gateway would already know where to fetch the content from, simply by reading the delegation that authorizes retrieval.

Probably not worthwhile but just thought I'd write it down :)

rfc/egress-with-ucan.md
```diff
@@ -0,0 +1,159 @@
...
+
+* Rather than serve non-token content rate-limited by default, require a delegation of `/space/content/retrieve` to some DID representing "anyone".
+* Bitswap should execute a `/space/content/retrieve` as well to respond to requests. Bitswap should be authorized by delegating to some DID representing Bitswap/Hoverboard.
+
+## Open questions
+
+* What does the `/space/content/retrieve` receipt look like?
+* Can you enumerate the contents of a space? Is that in scope here?
+* What does a Gateway URLs with a token look like?
+* What can be cached? This seems relatively cacheable, but we should be explicit in the design to make sure we're on a suitable path. This process needs to be *fast*, at least once the cache is warm.
```
travis, line 159: **thought**: once we've validated a token I think we can store something in [KV](https://developers.cloudflare.com/kv/) or a similarly performant caching layer, probably with an expiry, that says it's valid. we might need a mechanism to invalidate this cache, but that can probably wait until after the initial implementation?

travis (approved):
I'm a HUGE fan of this design, really nice, very excited to get this into prod!

rfc/egress-with-ucan.md
```diff
@@ -0,0 +1,159 @@
...
+To execute the invocation, the Executor will perform the following steps:
+
+1. Validate the proof chain.
+2. [Look up the Location Commitment again? :/ Should it become an argument?]
+3. [Charge the Space for the egress.]
+4. [Produce a suitable receipt.]
+
+## [To Come]
+
+* Rather than serve non-token content rate-limited by default, require a delegation of `/space/content/retrieve` to some DID representing "anyone".
```
alanshaw, line 151: Right, I guess `did:*`/`null` audience?

Alternatively, with a `space/content/serve` delegation you might delegate to the gateway and specify an empty bearer token.

rfc/egress-with-ucan.md
```diff
@@ -0,0 +1,159 @@
...
+// UCAN 0.9
+{
+  "iss": "did:key:zAgent",
+  "aud": "did:bearer:abcde12345",
+  "att": [
+    {
+      "with": "did:key:zSpace",
+      "can": "space/content/retrieve",
+      "nb": {
+        "cid": "bafy...7pcu"
```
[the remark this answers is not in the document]
Peeja: I don't think we do want that to be a link, actually. We want to request the content using the CID as an argument; this would provide the content itself (by reference) as the argument. Importantly, I believe, policies would resolve against the resolved `bafy...7pcu`, not the string `"bafy...7pcu"`. We want to match on the string.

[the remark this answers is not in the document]
alanshaw: There's no obligation to include the content in the invocation - we don't for upload/add for example.

I'm almost certain you can match a link in a policy...

IDK maybe I'm not understanding right?

Peeja:
### `did:bearer`, or not?

#### Reasons not to delegate to the token:

- Anyone can use the delegation without going through a trusted Gateway, which could allow a third party to issue invocations just to attack our customer with egress charges.
- The Gateway ends up invoking on someone else's authority, which is weird.
- It requires a new DID method, which is not to be undertaken lightly.
- It leaves an open question: how do you delegate to no-token? Is it `did:bearer:`? A whole new DID method? (Please no…)

#### Reasons to delegate to the token:

- A single delegation works for _any_ Gateway in the network. That may not actually be a good thing, though.
- We'd like to leave open the possibility of delegating to _any_ DID, such as delegating to some Agent who could then issue their own Invocations to retrieve content. Delegating to the token is parallel to that. But then, maybe an entity which can issue Invocations of their own doesn't need to use a Gateway at all. Then what happens to egress, though…?
- Our only index on the delegation store right now is by audience. That makes it really easy to look up delegations by token, when the token is the audience.

Given all that, I'm swayed by @alanshaw's argument: the most "correct" thing here is for the Gateway to be the audience of the delegation and the issuer of invocation, and to drop `did:bearer:` altogether. You'll have to delegate to each Gateway you want to use, but that's probably a good thing. We don't need a non-Gateway egress solution yet, so we can solve that later. And indexing these by something other than audience is an annoying change, but not insurmountable.

### Some other things from the feedback

- I like @travis's suggestion to include the origin.
- @alanshaw makes a good point that it's the root CID we care about, not the individual file's CID. In fact, the Gateway takes not just a (root) CID, but also a path. We should include that in the args.

### Proposed changes

- Remove the `did:bearer:` method.
- The Agent delegates `/space/content/retrieve` to the Gateway, not the token.
- The args are now:
  - Root CID [string]
  - Path [string? array of strings, so we can match on segments?]
  - Token [string]
  - Origin (or Referer? or both, as separate args?)
- The HTTP Gateway's sole job (on behalf of that role) is to translate HTTP requests into UCAN invocations, which it issues in its own name.
- The Retrieval Service (which is likely tied to the HTTP Gateway, but is a separate role) accepts and executes the invocation, and could receive invocations directly from clients.

---

### Open questions

- What is "egress" exactly? Is it Gateway → Retriever? Or is it Retrieval Service → Retriever? Or is it _Storage Node_ → Retriever? Most saliently, if a client retrieved from the Retrieval Service directly by issuing an invocation, would the Retrieval Service need to charge for egress? Presumably yes. Does the Gateway need to charge extra for egress through it? What if you go straight to the Storage Node? What are we (who?) actually charging for here?

- How do we handle path references? Consider the following Space:

  ```text
  <did:key:zSpace>
  ┖ [bafy...dir]
    ┠ file1: [bafy...file1]
    ┖ file2: [bafy...file2]
  ```

  We can access `file1` as either `bafy...dir/file1` or as `bafy...file1`. Should a single delegation be able to cover both? In other words, given a request for `https://bafy...dir.ipfs.w3s.link/file1` should the Gateway resolve the reference down to `bafy...file1` and invoke `/space/content/retrieve` to ask for _that_, or should it ask for something like `{ "root": "bafy...dir", "path": ["file1"] }` and let the Retrieval Service (the Executor of the invocation) resolve it? TBH, I'm not completely clear on where that resolution happens today, and where it will happen after this indexer work.

  - It's much more clear cut for the Gateway to simply pass along the path, as this makes it literally a translator from HTTP requests to UCAN invocations. We probably also don't want it to know enough to be able to do the resolution.
  - But two ways to reference the same content means two different possible sets of args to match to a policy. I suppose we could try to cram all the different paths to the file into the policy as `"or"`s, but that seems pretty darn messy. Or we could say that a delegation only authorizes a single reference form to be used on the Gateway, but that seems counter-intuitive for the customer.
  - Underlying all of this is the fact that most of the time the policy won't care about the content at all: it'll just apply to an entire Space. So I don't want to spend _too_ much effort on it. But it smells like something that might point to an important architectural decision.

hannahhoward (commented):
I'm going to leave a general comment before digging into specifics, that vageuly has to do with @alanshaw 's question about the DID.

So, I want to state my own thinking that there are two distinct and separate protocols for a retrieval:
1. The raw read from a source location (simple HTTP GET with range request)
2. The gateway query across potentially one or more providers to assemble a user level response, likely deserialized to a flat file. The second protocol uses the first protocol + the indexing service to accomplish its work.

Furthermore, the first protocol is close to a typical object storage interface, while the second, especially when run on a platform like Cloudflare, is much closer to a CDN (or a CDN + edge compute). For the remainder of this I'll use "Object storage layer" to refer to the first and "CDN layer" to refer to second.

There are some other important distinctions -- the object storage layer protocol in concert with the indexing service can be run without a trust relationship between the retriever and the storage node, while the CDN layer requires a trust relationship between the end user and the gateway, unless you're specifically staying in the bounds of the trustless portion of the Gateway protocol spec. (once you serialize to a flat file, you're now trusting the gateway)
 
Up till now, the object storage protocol hasn't even really existed a separate protocol cause we're just reading our own databases and our own storage devices, but going forward it gets increasingly public. With the indexing service, in combination with the PDP storage nodes, it becomes a distinct layer (the november version will not be fully there but it will get there soone). 

I apologize not for making this super clear as part of my own thinking. I see these also as different economic units of billing for of egress eventually -- the user pays `x` for a gateway retrieval, of which `y` goes to the gateway, and `z` goes to the storage node, where `y + z = x`. None of this needs to be figured out right this second cause we're running everything and our PDP nodes will probably just have an indirect accounting mechanism, but in a final product, the storage node would never get 100% of the user egress fee because we incur a sizable cost for running the cloudflare gateway that assembles the request, caches it, and services it super quick to the user.

When we talk about storage providers serving directly to end users, I generally don't see them running a gateway. Rather, I see a user with a native or server app who doesn't prioritize TTFB choosing to save money by running the freeway software directly, and only using the indexing service + object storage layer to get data.

Another use case could be someone building a custom retrieval gateway on top of the indexing service and object storage layer. Perhaps a product builder wants to build a query interface for archived blockchain data, and sell that independently. They would be storing on storacha, but charging their users to use their retrieval service that could build more complex IPLD queries against a blockchain than is available with the gateway protocol. (or perhaps just mirror the establish RPC api for querying for their chosen blockchain).

Anyway, coming back to this PR, I want to understand if `space/content/retrieve` refers the CDN layer of retrieval, or the object storage layer. A couple things that make this a bit confusing: 
1. There is no invoker for the CDN layer of the retrieval (i.e. it's not a proper UCAN request), at which the gateway is the executor.
2. There's no proper executor for the storage layer of the of the retrieval for now, until the storage nodes do UCAN auth.

The RFC simply says the gateway is the invoker and the executor and doesn't make clear the part of the retrieval we're referring to. 

I think ultimately `space/content/retrieve` makes sense as the CDN layer, and it's a weird one cause the invoker and executor are ALWAYS the gateway. That means the space delegates to the gateway in order for it to execute, and the token goes in the policy (caveats for now).

So long and short I agree with @alanshaw -- no did:bearer, because it's not a real entity. If we build a gateway that is called with a real invocations, then it makes sense that the issuer becomes a real, verifiable entity with tracable delegation to the space, the gateway is just an executor. There are certainly use cases for that, but not worth worrying about for then.

What about the storage layer retrieval in the future? So my suggestion is to actually make `space/index/query` and `space/content/retrieve/blob` for that, eventually. So when the CDN receives a request, it invokes `space/content/retrieve` with itself as the issue + executor, using delegations it has stored. And it generates a receipt that looks something like: 

```
{
    cmd: "/ucan/assert",
    args: {
       about: "bafy...cid" // space/content/retrieve invocation
       facts: {
          out: {
             ok: {
                statusCode: 200,
                bytesServed: 1,000,000
             }
          },
         run: [
           {
              cmd: "space/index/query"
           },
           {
              cmd: "space/content/retrieve/blob"
           },
           {
              cmd: "space/content/retrieve/blob"
           }
       ]
    }
  }
}
```
This wouldn't be sent back to the retriever but this provides nice instructions on how to bill the original user, that can be checked against the storage node and the indexer submitting receipts of their own for their parts of retrieval (along with the invocation that proves a retrieval/index was actually requested by the gateway). It all vaguely works for a trustless billing system :) (where even the indexing service could get paid)

That's my take.

But yea, I'm a block on not doing `did:bearer` after further consideration. Is there any reason we can't just throw it in caveats? We don't have policies but we have ucanto that can enforce this no?

hannahhoward (changes requested):
previous review ultimate requests a change to remove `did:bearer` and put it in caveats.

travis (changes requested):
ok cool - after reading through all this I'm switching to "request changes" - sounds like we're all on the same page anyway!

alanshaw:
Just for clarity, in my comments above I suggested delegating `space/content/serve` (note "serve" not "retrieve") to authorize the gateway to _serve_ data from the space with caveats appropriate to the gateway (origin, path etc.).

I propose `space/content/retrieve/*` to exist for the purpose of retrieving byte ranges. It could either be `derived` from `space/content/serve` or delegated explicitly.

[Example of capability derivation in ucanto](https://github.com/storacha/w3up/blob/61d408a2ecd61ca81a758d4e0b30f1ed0c3b77e8/packages/capabilities/src/space.js#L28-L45). 😱 I don't recall this being in the UCAN spec so perhaps explicit delegation is what we will do...

Separating the capabilities like this allows us to refer separately to the 2 protocols @hannahhoward called out above.

@Peeja on origin/path etc. I'd omit these unless we're ready to implement them.

travis:
agree re: origin! just want to make sure there's space in the interim protocol if we do decide to go that route, but I'm honestly hoping we can upgrade to UCAN 1.0 before we need that, where it will be a fairly easy and natural extension thanks to `pol`

Peeja:
@hannahhoward 💡❗

> 1. The raw read from a source location (simple HTTP GET with range request)
> 2. The gateway query across potentially one or more providers to assemble a user level response, likely deserialized to a flat file. The second protocol uses the first protocol + the indexing service to accomplish its work.

I hadn't considered that the Gateway would be composing _multiple_ pieces of content across potentially multiple providers to assemble a single response. Given that, I agree wholeheartedly with @alanshaw: that command should be `[/]space/content/serve`. That would be defined as "serving a request", which involves making the underlying requests and assembling the response.

**Question:** What level does Bitswap operate on? Does it serve requests, or fetch blobs?

**Question:** What are the proper names for these things? Is a "Blob" specifically the same as a "Shard"? (This is non-obvious to someone with less context, as typically a "blob" is simply any set of bytes, and in Git in particular it generally means roughly "a file's contents, separate from any filename that might point to it".) What is the name for the thing that the CDN Layer (Gateway) serves, and what is the name for the thing that the Object Layer serves?

rfc/egress-with-ucan.md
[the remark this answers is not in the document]
Peeja: The spec doesn't appear to say explicitly, but [it does say](https://github.com/ucan-wg/delegation/blob/0dbaca11cdd0c6a5ccb89d77db3f5fb0d2245347/README.md#selecting-on-bytes):

> ### Selecting on Bytes
>
> Bytes MAY be selected into. When doing so, they MUST be treated as a byte array (`[u8]`), and MUST NOT be treated as a Base64 string or any other representation.
>
> ```source-js
> // DAG-JSON
> { "/": { "bytes": "1qnBjPjE" } }
> 
> // Hexadecimal
> 0xd6 0xa9 0xc1 0x8c 0xf8 0xc4
> 
> // Selector
> ".[3]"
> // ⬆️  0x8c = 140
> ```

If the policy resolver understands DAG-JSON bytes, I assume it understands links as well. I'm not sure what the correct behavior would be when matching on a link, but I don't think it would be to treat it as a string, or as a literal JSON map `{ "/": "bafy...foo" }`. My assumption would be that, if anything, it would attempt to inline the link before applying the policy, and perhaps bail if it didn't have access to the content.

hannahhoward:
1. Bitswap will be `/space/content/serve` -- it's just another way to serve content, and it similarly can pull from multiple places

2. Blob is the right name and we should be careful about use of shard maybe. What is interesting is a blob is a blob of bytes, but it is also a shard of a larger dag when you are uploading large data. @alanshaw thoughts?

alanshaw:
> Blob is the right name and we should be careful about use of shard maybe. What is interesting is a blob is a blob of bytes, but it is also a shard of a larger dag when you are uploading large data. @alanshaw thoughts?

Yeah that's exactly right.

Peeja:
Version 2 is ready for review. Open questions at the end of the doc, and also copied here:

## Open questions

* When auth fails, should we return 401 Unauthorized, or 404 Not Found to mask that the data exists?
* Should `token` be `gatewayToken`? Or `authToken`? Is `token` too broad a term to be in the args of that invocation?
* What if we find a Location Commitment, but it has no Space on it (since [it's optional](https://github.com/storacha/indexing-service/pull/18))?

hannahhoward (approved):

rfc/egress-with-ucan.md
```diff
@@ -118,42 +141,67 @@ A UCAN delegation is used to allow token access to content. Typically, this is p
...
+
+We may eventually create an object-store service. This service will sit between the Gateway and the Storage Nodes. It will speak UCAN, while the Storage Nodes will continue to speak only HTTP byte-range requests. This service will be the Executor of these `/space/blob/get/0/1` invocations, rather than the Gateway. This will allow other Principals to use the same mechanisms to retrieve Blobs, including Hoverboard as well as customers storing and retrieving Blobs directly. It would return a means to fetch the actual data in the receipt for the invocation, such as an HTTP URL.
 
 ## Open questions
 
-* What does the `/space/content/retrieve` receipt look like?
-* Can you enumerate the contents of a space? Is that in scope here?
-* What does a Gateway URLs with a token look like?
-* What can be cached? This seems relatively cacheable, but we should be explicit in the design to make sure we're on a suitable path. This process needs to be *fast*, at least once the cache is warm.
+* When auth fails, should we return 401 Unauthorized, or 404 Not Found to mask that the data exists?
```
hannahhoward, line 205: I'm always a fan of 404

rfc/egress-with-ucan.md
```diff
@@ -118,42 +141,67 @@ A UCAN delegation is used to allow token access to content. Typically, this is p
...
+* Should `token` be `gatewayToken`? Or `authToken`? Is `token` too broad a term to be in the args of that invocation?
```
hannahhoward, line 206: for now I think `token` is fine.

rfc/egress-with-ucan.md
```diff
@@ -118,42 +141,67 @@ A UCAN delegation is used to allow token access to content. Typically, this is p
...
+* What if we find a Location Commitment, but it has no Space on it (since [it's optional](https://github.com/storacha/indexing-service/pull/18))?
```
hannahhoward, line 207: To be clear, once we begin using the indexing service, all new location commitments will have spaces on them.

It is optional only to account for data uploaded so far, which don't have a specific location commitment, and which we generate at query time for (i.e. to keep currently generated commitments valid) . If there is no space, we will assume limited public retrieval. (as if a null token delegation were stored)

hannahhoward (approved):
LGTM after method rename.

removing request changes to expidite the process when done

rfc/egress-with-ucan.md
```diff
@@ -0,0 +1,159 @@
...
+```json5
+// UCAN 0.9
+{
+  "iss": "did:key:zAgent",
+  "aud": "did:bearer:abcde12345",
+  "att": [
+    {
+      "with": "did:key:zSpace",
+      "can": "space/content/retrieve",
+      "nb": {
```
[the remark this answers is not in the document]
Peeja: For completeness:

Yep, definitely. And we should have no trouble adding that later on top of what's here.

rfc/egress-with-ucan.md
```diff
@@ -0,0 +1,237 @@
...
+
+### Serving a request
+
+To serve the request, the Gateway will perform the following steps:
+
+1. Look up the Location Commitment(s) for the given root CID from the Indexing Service.
+2. If none are found, respond with `404 Not Found`.
+3. Note if any Location Commitments were found with no Spaces. (If so, these are from before these changes, and mean we should fall back to the previous behavior later.)
+4. Get the set of unique Spaces from those Location Commitments. (There will usually be one Location Commitment, with one Space.)
+5. Repeat with each Space in any order, stopping if a successful response is produced:
```
travis, line 195: Is this going to potentially be slow? Is it a separate network request for each space we need to check? The performance is the only part of this spec I have concerns about, but I might be misunderstanding - has been a week or so since my head's been here so apologies if I'm missing something!

rfc/egress-with-ucan.md
```diff
@@ -0,0 +1,237 @@
...
+   1. Look up delegations in the store where the audience is `did:web:w3s.link` and the subject is the Space.
```
travis, line 196: I guess relatedly - is "the store" here the indexing service or something else?

rfc/egress-with-ucan.md
travis: I think we're missing a step here! we need to look up the "provider" registered with the space, which is a looking in a dynamo table that Freeway does not have access to - I think this needs to be handled either in w3up or w3infra

PR #37, opened 2024-10-14
rfc: egress tracking

fforbeck:
[📚 Preview ](https://github.com/storacha/RFC/blob/rfc/egress-tracking/rfc/egress-traffic.md)

Storacha Network aims to implement a scalable, automated mechanism for tracking egress traffic and updating Stripe’s Usage Record API with relevant data for accurate customer billing, this RFC outlines two approaches with distinct trade-offs.

### DIFF
--- a/rfc/egress-traffic.md
+++ b/rfc/egress-traffic.md
@@ -1,0 +1,200 @@
+# RFC: Egress Traffic Tracking
+
+## Authors
+
+- Felipe Forbeck
+
+## Status
+
+- Proposed
+
+## Context
+
+Storacha Network aims to implement a scalable and automated mechanism to track egress traffic for customer billing purposes. The objective is to update Stripe’s API with relevant egress data for each customer, ensuring accurate billing. This RFC outlines two potential approaches to achieve this goal, each with its own set of trade-offs.
+
+## Problem
+
+Currently, there is no established automated process to track and associate egress traffic per customer and report this usage to Stripe for billing. A solution is required to automate this process for seamless, scalable, and accurate billing based on egress traffic.
+
+## Proposal
+
+We propose two alternatives for tracking and updating egress billing traffic:
+
+### Alternative 1: SQS, Lambda, DynamoDB, and Stripe Integration
+
+1. Introduce an SQS queue to receive egress event data from the Cloudflare Freeway Worker.
+2. Use a Lambda function to process these events and store the egress data in a DynamoDB table, mapping the data to the respective customer.
+3. The Lambda function will then update Stripe’s API with the corresponding egress data size for the customer.
+
+```mermaid
+graph TD
+    CF[Cloudflare Freeway Worker] --> SQS[SQS Queue]
+    SQS --> Lambda[Lambda Function]
+    Lambda --> DynamoDB[DynamoDB - Egress Data Storage]
+    Lambda --> Stripe[Stripe API]
+```
+
+**Advantages:**
+
+- Decoupled, event-driven architecture using SQS and Lambda, providing clear separation of responsibilities.
+- Fine-grained control over event handling and processing.
+
+**Disadvantages:**
+
+- Requires managing additional infrastructure (SQS queue, Lambda function, DynamoDB table) solely for this purpose.
+- Cloudflare Freeway Worker must be authorized to publish events to the SQS queue, adding complexity to authentication and authorization mechanisms.
+- Increased operational overhead due to the need to monitor and maintain new infrastructure components.
+
+### Alternative 2: Extend `w3up/upload-api` with `usage/record` Capability
+
+1. Add a new capability, `usage/record`, to the `w3up/upload-api`.
+2. Implement this capability in the `w3infra` repository to directly update Stripe’s API with the egress data size for each customer.
+3. The Cloudflare Freeway Worker would invoke the `w3up/upload-api` every time a resource is served, passing the egress data size and customer details to trigger the Stripe update.
+4. Ensure the Cloudflare Freeway Worker is authorized to invoke this API via delegation.
+5. Each invocation will be logged, providing signed receipts of egress traffic.
+
+```mermaid
+graph TD
+    CF[Cloudflare Freeway Worker] --> w3up[w3up/upload API]
+    w3up --> Stripe[Stripe API]
+    w3up --> Logs[Logs - Egress Data Receipt]
+```
+
+**Advantages:**
+
+- No need to maintain additional infrastructure components such as SQS queues, Lambdas, or DynamoDB tables, as existing infrastructure already handles similar UCAN invocations.
+- Simplified architecture by relying on current logs and receipts to record traffic data.
+- Reduces operational complexity with fewer components to manage.
+
+**Disadvantages:**
+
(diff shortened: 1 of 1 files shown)


rfc/egress-traffic.md
```diff
@@ -0,0 +1,200 @@
...
+
+Storacha Network aims to implement a scalable and automated mechanism to track egress traffic for customer billing purposes. The objective is to update Stripe’s API with relevant egress data for each customer, ensuring accurate billing. This RFC outlines two potential approaches to achieve this goal, each with its own set of trade-offs.
+
+## Problem
+
+Currently, there is no established automated process to track and associate egress traffic per customer and report this usage to Stripe for billing. A solution is required to automate this process for seamless, scalable, and accurate billing based on egress traffic.
+
+## Proposal
+
+We propose two alternatives for tracking and updating egress billing traffic:
```
alanshaw, line 21: Stepping back - do you have an idea how you're going to log the right amount of egress? Count bytes? What happens when the connection is closed - can that be handled in a worker?

Do cloudflare have logs you could just crawl?

rfc/egress-traffic.md
```diff
@@ -0,0 +1,238 @@
...
+
+- **Question:** Which approach requires less effort to maintain in the long run? Can the additional infrastructure of Alternative 1 or 3 be justified given their decoupled architecture?
+- **Answer:** Alternative 2 requires less effort to maintain in the long run since it leverages existing infrastructure (`w3up/upload-api`), avoiding the need to set up new infrastructure components (SQS, Lambda, DynamoDB).
+
+### 2. **Security:**
+
+- **Question:** How should delegation and authentication mechanisms be handled, particularly in Alternative 2? What security implications arise from allowing the Cloudflare Freeway Worker to directly interact with the `w3up/upload-api`?
+- **Answer:** Delegation can be handled by granting the Cloudflare Freeway Worker permission to invoke the required capabilities. The Gateway will need to become a `w3up` client to claim the delegation and execute the invocation securely.
+- **Question:** Who will create the delegation to the Cloudflare Freeway Worker?
+- **Answer:** ???
```
travis, line 170: we can do this manually using the upload api service's private key

rfc/egress-traffic.md
```diff
@@ -0,0 +1,238 @@
...
+### 3. **Performance Impact:**
+
+- **Question:** Will the real-time nature of API updates in Alternative 2 introduce latency concerns for the Cloudflare Freeway Worker? How does this compare to the event-driven model in Alternatives 1 and 3?
+- **Answer:** By using Cloudflare’s "fire and forget" context calls, the Freeway Worker can avoid blocking while serving the resource. This minimizes latency, allowing the request to complete without waiting for the billing events, making the performance impact negligible.
+
+### 4. **Logging and Monitoring:**
+
+- **Question:** How can logs and monitoring be effectively implemented in both alternatives? Does Alternative 2’s reliance on signed receipts offer sufficient logging capabilities? Do we really need these signed receipts?
+
+- **Answer:** ???
```
travis, line 181: I do think it's at least moderately important to have those signed receipts - I suspect customers are going to want as much information as possible about what they're being charged for. We might be able to just aggregate metrics around them and store in Glacier for the sake of minimizing costs, but having signed receipts for each request feels like something that will eventually be fairly useful?

@alanshaw is right that there will be a LOT of them and they'll probably have different retention policies than most other UCAN invocation receipts, though for the sake of getting this shipped we might want to use the upload API service for now and add a new service later?

It definitely might be worth adding a new service that uses a similar invocation processing pipeline and different retention policies - ideally we'd move egress tracking receipts to Glacier pretty quickly.

rfc/egress-traffic.md
```diff
@@ -0,0 +1,200 @@
...
+**Advantages:**
+
+- Decoupled, event-driven architecture using SQS and Lambda, providing clear separation of responsibilities.
+- Fine-grained control over event handling and processing.
+
+**Disadvantages:**
+
+- Requires managing additional infrastructure (SQS queue, Lambda function, DynamoDB table) solely for this purpose.
+- Cloudflare Freeway Worker must be authorized to publish events to the SQS queue, adding complexity to authentication and authorization mechanisms.
+- Increased operational overhead due to the need to monitor and maintain new infrastructure components.
```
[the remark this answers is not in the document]
travis: Stripe does seem to be advertising the new "billing meters" that we'll need to migrate to sooner or later as something that can "scale right along with us" - we are still fairly small fish compared to their biggest users so it seems reasonable to hope this will support the kind of volume we want to send at it! 

I'd propose we do the first implementation with the queue-based approach in option 3 and worry about rolling up usage charges ourself if Stripe starts really buckling under the load:

https://stripe.com/resources/more/what-is-metered-billing-heres-how-this-adaptable-billing-model-works
https://stripe.com/blog/three-ways-our-usage-based-billing-product-is-unique

rfc/egress-traffic.md
travis: I don't think cloudflare has historical logs like this, but I could be mistaken

rfc/egress-traffic.md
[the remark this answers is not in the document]
fforbeck: From their docs:
> Increased rate limits up to 100,000 events per second

Anyways, the first iteration is using a queue and we can fine-tune that as needed.

rfc/egress-traffic.md
```diff
@@ -0,0 +1,200 @@
...
+    CF[Cloudflare Freeway Worker] --> w3up[w3up/upload API]
+    w3up --> Stripe[Stripe API]
+    w3up --> Logs[Logs - Egress Data Receipt]
+```
+
+**Advantages:**
+
+- No need to maintain additional infrastructure components such as SQS queues, Lambdas, or DynamoDB tables, as existing infrastructure already handles similar UCAN invocations.
+- Simplified architecture by relying on current logs and receipts to record traffic data.
+- Reduces operational complexity with fewer components to manage.
```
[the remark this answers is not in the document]
fforbeck: Makes sense. I've updated the RFC to tackle this in another iteration. For the first one we will keep them in the same infra as discussed with @hannahhoward.

rfc/egress-traffic.md
```diff
@@ -0,0 +1,238 @@
...
+  const upcomingInvoice = await stripe.invoices.retrieveUpcoming({
+    customer: "{{CUSTOMER_ID}}",
+  });
+  ```
+
+  Stripe processes meter events asynchronously, so upcoming invoices might not immediately reflect recently received meter events.
+
+## Recommendation
+
+**Alternative 3** offers a more streamlined architecture by leveraging the existing `w3up/upload-api` and adds minimal infrastructure components like SQS, Lambda to prevent rate limiting issues on Stripe API.
```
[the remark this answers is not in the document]
fforbeck: Cool. The first iteration is up for review: https://github.com/storacha/w3infra/pull/430

rfc/egress-traffic.md
fforbeck: Also updated the RFC to handle the closed connection issue in another iteration using the Transformation Streams as suggested by @alanshaw .

rfc/egress-traffic.md
fforbeck: That makes sense @travis .

PR #41, opened 2025-02-11
rfc: IPNS over Lit

alanshaw:
This RFC proposes a simple solution that enables multi-writer IPNS by leveraging the Lit protocol and UCAN.

[📚 Preview](https://github.com/storacha/RFC/blob/feat/ipns-over-lit/rfc/ipns-over-lit.md)

### DIFF
--- a/rfc/ipns-over-lit.md
+++ b/rfc/ipns-over-lit.md
@@ -1,0 +1,37 @@
+# RFC: IPNS over Lit
+
+## Authors
+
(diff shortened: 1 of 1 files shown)


travis:
Agree with Felipe!

One thing I'd particularly like to understand is how much this usage of the LIT network would cost - I think each operation has an associated cost paid in their token, and it would be nice to understand how much this is.

PR #45, opened 2025-03-05
rfc: replication

alanshaw:
[📚 Preview](https://github.com/storacha/RFC/blob/rfc/replication/rfc/replication.md)

### DIFF
--- a/rfc/replication.md
+++ b/rfc/replication.md
@@ -1,0 +1,119 @@
+# RFC: Replication
+
+## Authors
+
+- [Alan Shaw](https://github.com/alanshaw), [Storacha Network](https://storacha.network/)
+
+## Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC2119](https://datatracker.ietf.org/doc/html/rfc2119).
+
+## Introduction
+
+Replication ensures that if a storage node loses a blob, it remains available in the network. This RFC proposes an extension to the blob protocol that allows nodes within the network to replicate data between each other _after_ a storage node has received an initial upload.
+
+Out of scope: This RFC does not propose any solution for repairing lost data from any node.
+
+## Proposal
+
+A `space/blob/replicate` invocation instructs the upload service to replicate a blob to the specified number of storage nodes. A replication request is only valid to be issued _after_ the client has received a `blob/accept` receipt, indicating the target storage node has successfully received the blob.
+
+The caveats for `space/blob/replicate` MUST include a `replicas` parameter, an unsigned integer indicating the number of copies the network should ensure are stored _in addition_ to the original.
+
+It is RECOMMENDED that the upload service mandate a minimum _and_ a maximum value. The minimum and maximum MAY be the same.
+
+e.g. `replicas: 2` will ensure 3 copies of the data exist in the network.
+
+```json5
+{
+  "iss": "did:key:zAlice",
+  "aud": "did:web:upload.service.example.com",
+  "att": [
+    {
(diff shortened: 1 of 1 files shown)


alanshaw:
My primary concern here is why a storage node would actually replicate the data if they know where they can obtain the bytes - i.e. cna they just fetch the bytes as needed for any challenge on the data?

So I was thinking it might be better for the replication nodes to obtain a temporary delegation from the target node allowing them access - meaning the replication node would _have_ to transfer the data in order to respond to challenges in the future.

frrist:
One incentive for storage nodes to replicate content is the revenue they (will) earn from retrieval requests, not just for passing proof checks. I’m assuming the retrieval reward (will) generally exceed any proof-only reward, motivating nodes to actually store the data rather than simply fetching bytes on-demand. I think this creates a strong case for actively replicating data, particularly when combined with your point of a temporary delegation issued by the node where the content was initially uploaded.

Balancing these delegations’ expiration times might require some fine-tuning though—especially around content size or factors like available bandwidth—to prevent short-term delegations from expiring prematurely.

frrist (approved):

rfc/replication.md
```diff
@@ -0,0 +1,125 @@
...
+        "cause": { "/": "bafy..allocate" }
+      }
+    }
+  ],
+  "prf": [],
+  "sig": "..."
+}
+```
+
+When the `blob/replica/transfer` task is complete a receipt is issued. The receipt is communicated back to the upload service via a `ucan/conclude` invocation.
```
frrist, line 121: After a second look and starting the implementation of this, what value(s) are in the receipt of a `blob/replica/transfer` invocation? Would something like the `blob.AcceptOk` receipt be sufficient:
```
type AcceptOk struct {
  site Link
  PDP optional Link
}
``` 
`site` is the location claim and `PDP` is the aggregate, inclusion proof, and piece link? All of this wrapped in a `ucan/conclude`?

rfc/replication.md
```diff
@@ -0,0 +1,125 @@
...
+        "cause": { "/": "bafy..replicate" }
+      }
+    }
+  ],
+  "prf": [],
+  "sig": "..."
+}
+```
+
+The `blob/replica/allocate` task receipt includes an async task that will be performed by the storage node - `blob/replica/transfer`. The `blob/replica/transfer` task is completed when the storage node has transferred the blob from its location to the storage node.
```
frrist, line 91: Similar question about receipts here, what goes into the receipts of `blob/replica/allocate` invocations, if anything? (aside from the mentioned async task)

frrist (commented):
couple follow on questions

rfc/replication.md
alanshaw: Yes, should be the same (or similar) to `blob/accept` receipt.

rfc/replication.md
alanshaw: I just updated the blob spec to reflect the current implementation but yes the same (or similar) to blob allocate receipt https://github.com/storacha/specs/blob/main/w3-blob.md#allocate-blob-receipt

rfc/replication.md
```diff
@@ -0,0 +1,125 @@
...
+      "with": "did:key:zStorageNode",
+      "can": "blob/replica/transfer",
+      "nb": {
+        /** The blob that will be transferred. */
+        "blob": {
+          "digest": { "/": { "bytes": "..." } },
+          "size": 1234
+        },
+        /** The location the blob will be transferred from. */
+        "location": { "/": "bafy..locationCommitment" },
```
frrist, line 110: Is the intent for the node receiving the transfer invocation to "push" the data to the replicating node? Or is the intention for the replicating node to "pull" the data from the original node?

My read of both the `blob/replica/transfer` and `blob/replica/allocate` location field is that is specifies the location on the original node where the data resides. Is that right?

PR #46, opened 2025-03-12
RFC to discuss state of indexer

hannahhoward:
# Goals

Try to understand current state of indexer deployment.

# Implementation

I've got the analysis, but still working on solutions.


rfc/indexer-deployment.md
```diff
@@ -0,0 +1,64 @@
...
+
+# Meta-issue: Triple request rates from ipfs.io, all RAW format
+
+It appears ipfs.io's initial HTTP retrieval strategy is to just send RAW single block requests to our gateway (no CAR requests). This has nearly tripled our traffic. Also, our sharded DAG indexes flow is poorly suited to this access pattern.
+
+# Recommended next steps
+
+# Short Term: Put the indexer behind Cloudfront
+
+It seems like an easy short term fix is to just put the indexer behind cloudfront. It's not clear costs would be much higher than API gateway, especially if we tied the labmda to a function URL.
```
alanshaw, line 59: Yeah, I we should definitely do this.

alanshaw: We should also add cache headers to responses from the indexer as this will cause cloudflare to cache responses to indexer requests automatically.

rfc/indexer-deployment.md
```diff
@@ -0,0 +1,64 @@
...
+
+Currently, the indexer is underperforming the content claims service. We can attribute this gap to two performance issues:
+
+1. The primary reason the indexing service underperforms is that the content claims service is fronted with a cloudfront cache.
+
+For most existing uploads, which were done with the store protocol, retrieval is done block by block with lookups to the content claims or indexer service on each block. In this context, results are usually cached on cloudfront for content claims. This leads to RTTs in the neighborhood of 20-40ms.
+
+Comparitively, the indexer's RTTs from Europe can go >1s.
+
+2. Seperately, the indexer is struggling to achieve good cache hit rates for the providers cache. This cache, set to about 10GB in size initially, can hold about 2 million records. So far, what we have seen is this cache fills almost immediately, and eviction policies leave us with a hit rate around 10%. We tried increasing cache size to 20GB and while this appeared to cause a slight improvement it was minimal.
```
alanshaw, line 29: I think...we have a lot of data. The dynamodb blocks table has >48 billion items in it. 2 million records is 0.004% of the items in there. On these counts it's not surprising for it to be filled immediately.

In hindsight a single cache for the breadth of data we have here is probably not going to work, unless it's huge. This could perhaps work with multiple regional caches, but then it gets harder for a storage node to preload the indexer with location commitments.

It might be worth considering DynamoDB as a cache, since it can grow as big as required and supports TTL. It costs us like $6,000 a month (total) but is currently holding >24,000x times more data than redis is right now. I assume read times are slightly longer, but, DyanmoDB is still very fast.

---
For reference, it's 12TB in size.

rfc/indexer-deployment.md
```diff
@@ -0,0 +1,64 @@
...
+
+Of note, we don't cache anything for not found responses, so it's possible we're also paying a price for 404's, which constitute the majority of head responses.
+
+Ultimately the penalty for slow performance is dwarfed by the RTT latency penalty of not having Cloudfront
+
+## Cost Analysis
+
+![Current costs over the last week for the indexer](./images/Indexer-costs.png)
+
+The indexer is costing about $1000 per week to run. Much of this cost is due to requests going through an API Gateway service, with secondary costs to run the lambda itself, compute associated with Elasticcache, and memory associated with Elasticache.
```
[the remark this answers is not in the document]
hannahhoward: no I have better cost options.

And again, the cost is API Gateway believe it or not!

PR #47, opened 2025-03-12
RFC: Bluesky Backups

Peeja:


hannahhoward (commented):
The original tech briefing -- https://www.notion.so/storacha/BlueSky-Backup-Technical-Briefing-1595305b552480d7939ec16b811ebb54 -- made the assumption we'd just use a local db -- SQLite or something -- to track this functinality. I'd like to understand the cost/benefit analysis of going with a full pail/clock approach vs good ol fashion web2 tech.

Things to consider:
- time to ship?
- potential risks with actually using pail/clock in prod at scale?
- potential benefits of pushing these technologies to be prod ready?
- long term cost diff of pail/clock vs traditional db approach

I'd like to see this before I say "go ahead" with this approach.

Peeja:
@hannahhoward Yeah. Having written it all out now, I agree. I've flipped it around and put the DB version first, with an iteration plan towards moving everything into the Space. I've also written what I see as the value at each step, and where the milestones (well, there's only one that's relevant) line up.

We should get a better picture of risks, costs, and complexity to build for the pail/clock portion before committing. I don't think I know enough to do that fully right now. I'll confer with @alanshaw. At the same time, I'm going to let this coast a little bit, because I think the work leading up to the two milestones we have is fully unblocked at this point and I do want to focus on getting those through without getting too distracted by the future. But I do want to keep mulling on this so we're ready for the next step when we're ready to take one.

alanshaw:
I love this idea - if you end up going down this route I'll make myself super available to help. I'll also hopefully be proving out w3clock in prod with the AI work.

If it's easier/quicker to do SQLite then I'm not going to block although...I'm not sure how it works if we're hosting the service? In general I'd like to not store data for users unless it's in a space.

PR #53, opened 2025-06-02
rfc: proposal for standard delegation handling

frrist:
:blue_book: [Preview](https://github.com/storacha/RFC/blob/b74f9c86962825a807c0d253722315ef61f7e11b/rfc/ucan-delegation-handling.md)

### DIFF
--- a/rfc/ucan-delegation-handling.md
+++ b/rfc/ucan-delegation-handling.md
@@ -1,0 +1,167 @@
+# RFC: Standardizing UCAN Delegation Handling
+
+## Summary
+
+This RFC proposes standardizing how we handle UCAN delegations and proofs across our services, adopting CAR (Content-Addressed aRchive) files as the canonical format with base64 encoding for transport.
+
+## Problem Statement
+
+Our services currently accept UCAN delegations in multiple formats:
+
+- Base64 encoded strings.
+
+- Base64-encoded CIDv1 with embedded CAR data.
+
+This fragmentation creates:
+
+- **Integration complexity**: Different services expect different formats
+- **Debugging difficulty**: Multiple encoding layers obscure issues
+- **Documentation burden**: Each format requires separate examples
+- **Tooling gaps**: No standard way to validate or inspect delegations
+
+## Proposed Solution: CAR File Standard
+
+- **Storage format**: Binary (CBOR) CAR files with `.car` extension
+- **Transport encoding**: Base64 encoded `.car` content for environment variables and HTTP headers
+
+### Why CAR?
+
+1. **Ecosystem compatibility**: Leverage existing tooling.
+2. **Binary efficiency**: Compact storage for delegation data.
+3. **Standard tooling**: `ipfs-car`, `ipfs` CLI support.
+
+### Usage Examples
+
+#### File-based configuration:
+
+```bash
+# Place delegation in standard location
+cp delegation.car /etc/service/delegation.car
+
+# Inspect CAR Roots
+ipfs-car roots delegation.car
+
+# Inspect CAR Blocks
+ipfs-car blocks delegation.car
+```
+
+#### Environment Variable Configuration:
+
+```bash
+# Convert CAR to base64 for environment
+export UCAN_DELEGATION=$(base64 -w 0 delegation.car)
+
+# In Docker/Kubernetes
+env:
+  - name: UCAN_DELEGATION
+    value: "OqJlcm9vdHOB2CpYJQAB..."
+```
+
+#### HTTP header usage:
+
+```bash
+# Include delegation in request
+curl -H "Authorization: UCAN $(base64 -w 0 delegation.car)" \
+     https://api.example.com/resource
+
+# Alternative header format
+curl -H "X-UCAN-Delegation: $(base64 -w 0 delegation.car)" \
+     https://api.example.com/resource
+
+```
+
+### Implementation Guidelines
+
+```go
+func loadDelegation(config Config) ([]byte, error) {
+    // Priority 1: File path (recommended)
+    if config.DelegationFile != "" {
+        return os.ReadFile(config.DelegationFile)
+    }
+    
+    // Priority 2: Environment variable (containerized environments)
+    if encoded := os.Getenv("UCAN_DELEGATION"); encoded != "" {
+        return base64.StdEncoding.DecodeString(encoded)
+    }
+    
+    // Priority 3: Direct value (testing only)
+    if config.DelegationBase64 != "" {
+        return base64.StdEncoding.DecodeString(config.DelegationBase64)
+    }
+    
+    return nil, errors.New("no delegation provided")
+}
+```
+
+### Standard Locations
+
+- File storage: `/etc/service/delegation.car`
+- Environment variable: `UCAN_DELEGATION`
+- HTTP header: `Authorization: UCAN <base64>` or `X-UCAN-Delegation: <base64>`
+
+## Looking Forward
+
+In the future, services may accept just the root CID of a delegation CAR file instead of the full content.
+
+### Benefits
+
+1. **Minimal transport**: Pass single CIDs instead of kilobytes of data.
+2. **De-duplication**: Common delegations stored once, referenced many times.
+3. **Flexible retrieval**: Services choose optimal resolution strategy.
+
+### Usage Examples
+
+**Environment variable with CID**:
+
+```bash
+# Instead of full CAR file
+export UCAN_DELEGATION_CID="bafy..."
+```
+
+**HTTP header with CID**:
+
+```bash
+# Lightweight authorization header
+curl -H "Authorization: UCAN-CID bafy..." \
+     https://api.example.com/resource
+```
+
+**Configuration hierarchy**:
+
+```yaml
+delegation:
+  # Option 1: Local file
+  file: /etc/service/delegation.car
+  
+  # Option 2: Inline base64
+  data: "eyJyb290cyI6W3siLyI6ImJhZnlyZW..."
+  
+  # Option 3: CID reference (future)
+  cid: "bafy..."
+  resolver: "gateway.storacha.network"  # or "local-ipfs" or "s3://bucket/delegations"
+```
+
+### Resolution Strategies
+
+Services receiving CIDs would need to implement resolution:
+
+```go
+func resolveDelegationCID(cid string, config ResolverConfig) ([]byte, error) {
+    switch config.Type {
+    case "storacha-gateway":
+        return fetchFromGateway(config.Gateway, cid)
+    
+    case "local-ipfs":
+        return fetchFromLocalIPFS(cid)
+    
+    case "s3":
+        return fetchFromS3(config.Bucket, cid)
+    
+    case "delegation-service":
+        return fetchFromDelegationService(config.Endpoint, cid)
+    
+    default:
+        return nil, fmt.Errorf("unknown resolver type: %s", config.Type)
+    }
+}
+```


rfc/ucan-delegation-handling.md
```diff
@@ -0,0 +1,167 @@
...
+
+## Summary
+
+This RFC proposes standardizing how we handle UCAN delegations and proofs across our services, adopting CAR (Content-Addressed aRchive) files as the canonical format with base64 encoding for transport.
+
+## Problem Statement
+
+Our services currently accept UCAN delegations in multiple formats:
+
+- Base64 encoded strings.
```
alanshaw, line 11: These are multibase base64 encoded CAR bytes. This is the old format.

rfc/ucan-delegation-handling.md
```diff
@@ -0,0 +1,167 @@
...
+
+- Base64-encoded CIDv1 with embedded CAR data.
```
alanshaw, line 13: This is the new(er) format. In the CID, the multihash uses the identity codec and the [IPLD codec is CAR](https://github.com/multiformats/multicodec/blob/master/table.csv#L142).

This was an attempt to make the delegation self describing.

I appreciate this is somewhat onerous to create or parse.

alanshaw (commented):
🤷 I agree that creating or parsing the delegation is onerous and that it might be much easier to just plain base64 encode the CAR data. However, there are existing delegations that have been created with the old and new formats and are currently being used in the wild. Adding another format won't make the others go away, and it'll add more complexity to parsing delegations.

I'm ok to coalesce around _another new format_ and even support it as the only format accepted in the storage node if that's what we think we need, but I'd rather standardise around the newer of the _existing_ formats.

frrist:
Obligatory [xkcd](https://xkcd.com/927/) 
image

I am not sure introducing another standard here is necessary. My main motivation was to define a mechanism for persisting/reading delegations to/from files. From a UX perspective, I much prefer passing around appropriately named files, than remembering which string is which delegation. On the other hand, I know it's necessary to provide delegations as raw strings sometimes. So I tried to land on a file format that was easy to convert to a string. Essentially, I am after a standard for files that contain delegations; it sounds like we already have a standard for encoding them as strings, so maybe we should just adopt that existing newer format and focus this RFC on:

1. Standard file extensions and locations for storing delegations (e.g., `.ucan`, `.dlg`, etc. files)
2. Conventions for loading delegations from files.
3. Tooling to inspect/validate delegation files using the existing format (via improvements to go-mkdelegation which currently creates `.b64` files (bit of a hack) containing delegations)

This way we're not introducing a new encoding format, just standardizing how we handle the existing format when it's stored in files. Would that address your concerns while still improving the developer experience around delegation management?

Peeja:
I like the idea of being more consistent. I do think we have some challenges to solve for, though.

As for tooling, I would *love* to have better tooling to examine delegations, but I think it would be a shame if it only accepted them as CARs. Sometimes you have a larger chunk of data which contains delegations. I'd prefer for it to think in terms of blocks. CARs are really a form of transport themselves, making certain blocks available.

But I do think it's worth standardizing which storage and transport formats we use across our own codebases in specific classes of use case, which this RFC is largely getting at. There may be different ways to do it which are most appropriate in different cases, but in any given case, it should be clear to us which one is the one to use.

frrist:
> As for tooling, I would love to have better tooling to examine delegations

At present, https://github.com/storacha/go-mkdelegation?tab=readme-ov-file#example-output should help with parsing - it will parse delegations, and recursively parse any proofs they contain, if they are delegations. It will also generate delegations as Base64-encoded CIDv1 with embedded CAR data - which is currently the only supported format for generation and parsing.

> but I think it would be a shame if it only accepted them as CARs

:100: For practical deployment, we still need to support multiple transport encodings since stateless services (lambdas, HTTP headers, env vars) have different constraints. The goal is standardizing *which* transport format we use in each context - CAR files for disk storage, base64-encoded CARs for constrained environments.

> But I do think it's worth standardizing which storage and transport formats we use across our own codebases in specific classes of use case, which this RFC is largely getting at. There may be different ways to do it which are most appropriate in different cases, but in any given case, it should be clear to us which one is the one to use.

Exactly. Specifically:
1. Storacha Services must accept delegations as Base64-encoded CIDv1 with embedded CAR data - something we already support nearly everywhere
2. Additionally, all storacha services must accept CAR _files_ as a standard, e.g. `delegation.car`
3. Future: services should accept just the root CID from the CAR header (not CIDs with the CAR encoded into them), with the actual CAR data retrieved via gateway/IPFS/S3/etc.

Right now, the tooling gap between formats 1 and 2 is:
- I have a Base64-encoded CIDv1 with embedded CAR data and I want to transform it into a CAR file - this isn't doable from the CLI alone far as I am aware
- I have a delegation.car and I want to transform it to a Base64-encoded CIDv1 with embedded CAR data - this also isn't doable from the CLI.

Peeja:
Love it.

Something minor that occurs to me (that's been implicit in this whole thing) is that there really is a difference between CARs writ large and _delegation CARs_ specifically. Namely, a delegation CAR's root block is the root of a delegation. You can also store the blocks of a delegation anywhere in any CAR, but that's a specific standard for using CARs in that particular way.

frrist:
> there really is a difference between CARs writ large and delegation CARs specifically

Per the V1 CAR spec (https://ipld.io/specs/transport/car/carv1/#constraints), roots _should_ be present in the CAR body. A "delegation CAR" is just a regular CAR where the root happens to be a delegation.

So a "delegation CAR" would just be a standard CAR where:
- The root CID happens to point to a delegation block
- All referenced blocks (the delegation + its proofs) are included

This seems to be the normal way CARs are meant to work - am I missing something about what makes delegation CARs special? Happy to clarify the RFC if there's a specific convention I should be aware of.

alanshaw:
@frrist @Peeja something to consider here is that I'm not sure you're aware of:

When we archive data into a CAR we usually create a "variant" block as the root block (the root CID in the header) which serves to identify the type and version of the data the CAR contains. They are typically a map with the key as the version string.

The delegation variant block is:

https://github.com/storacha/go-ucanto/blob/06a2c2d09f708014bda62aba27bb1a146ebd29eb/core/delegation/datamodel/archive.ipldsch#L2

Which looks like:

```json
{
  "ucan@0.9.1": { "/": "bafy..." }
}
```

Also agent message:

https://github.com/storacha/go-ucanto/blob/06a2c2d09f708014bda62aba27bb1a146ebd29eb/core/message/datamodel/agentmessage.ipldsch#L2

Which looks like:

```json
{
  "ucanto/message@7.0.0": { "/": "bafy..." } 
}
```

See also encrypted data:

https://github.com/storacha/upload-service/blob/def9d48e93bbe5c664f497c2d0214bad4ccb3b54/packages/encrypt-upload-client/src/core/encrypted-metadata.js#L11-L23

...and telegram backup data:

https://github.com/storacha/tg-miniapp/blob/70246a16c79d7cee523912afc9b65319ca9a1ea6/app/lib/server/runner.ts#L69-L71

frrist:
> When we archive data into a CAR we usually create a "variant" block as the root block (the root CID in the header) which serves to identify the type and version of the data the CAR contains. They are typically a map with the key as the version string.

So essentially a type-system/envelope built with JSON objects? I assume this is natural in JS where everything is dynamic, but imo feels awkward in compiled languages, like Go, where compile-time type safety is preferred.

Reactivity, this feels problematic since:
- We must parse JSON to know what type we're dealing with, though I understand the intention here. 
- The format is implicit, stringly typed, and not formally defined, far as I am aware?

Would we be open to defining a specification for this variant block, maybe something like:
```
# TODO better name
type StorachaCARHeader struct {
  type String  // enum: "delegation", "message", etc
  version String
  root cid.Cid
}
```

This would maintain the same functionality while making implementation requirements explicit across languages.

Until we land on a spec for headers, I would prefer to make this a non-requirement for car files containing delegations - unless there is a strong reason to adopt the pattern as it exists?

alanshaw:
> like Go, where compile-time type safety is preferred.

Is none of the above type safe? I linked to examples for variant blocks that are implemeted in Go. The version identifiers are literally [encoded into the struct types](https://github.com/storacha/go-ucanto/blob/06a2c2d09f708014bda62aba27bb1a146ebd29eb/core/delegation/datamodel/archive.go#L35-L37) themselves, not as strings.

> We must parse JSON to know what type we're dealing with, though I understand the intention here.

It's not JSON - it's IPLD data that I've represented as dag-json in the examples above for readability.

> The format is implicit, stringly typed, and not formally defined, far as I am aware?

You mean for the type/version info? We didn't formally spec that but I think it's not difficult to understand i.e. type name then "@" then semver version. `StorachaCARHeader` appears to encode the same information but it will allow you to decode a block with ANY version string which you then have to do some further checking on the version field to see if your code supports the specified version. With the existing variant block you either can decode it or not. If you can decode it you support it. i.e. you haven't done work to then find you can't use it. Note: it's implemented as a [keyed union in IPLD](https://ipld.io/docs/schemas/features/representation-strategies/#union-keyed-representation).

> Until we land on a spec for headers, I would prefer to make this a non-requirement for car files containing delegations - unless there is a strong reason to adopt the pattern as it exists?

We did used to do that, but I believe we largely rely on [`delegation.Archive()`](https://github.com/storacha/go-ucanto/blob/06a2c2d09f708014bda62aba27bb1a146ebd29eb/core/delegation/delegation.go#L148-L155) and [`delegation.Extract()`](https://github.com/storacha/go-ucanto/blob/main/core/delegation/delegation.go#L202-L214) (and JS equivalents) for encoding/decoding delegations from CAR files which creates the aforementioned variant block on encode and and validates it on extract.

So do we just assume the CAR file contains a delegation that decodes wrt a specific UCAN version? I don't think that's better than what we have now. What is the reason, other than preference :trollface:? My guess is fewer bytes (fair) and easier to encode/decode (mitigated by the existing utilities) - but maybe I'm missing something?

I guess I'd be open to allowing the variant block to be optional, with the assumed UCAN version being 0.9.1 when it does not exist? We have code to deal with this already: https://github.com/storacha/guppy/blob/main/pkg/delegation/proof.go and https://github.com/storacha/upload-service/blob/main/packages/w3up-client/src/proof.js

frrist:
> It's not JSON - it's IPLD data that I've represented as dag-json in the examples above for readability.

:facepalm: well that's what I get for being reactive, sorry(!) :sweat_smile:, and sorts most of what I thought was problematic. My cop-out is that I was invoking Cunningham's Law....  

> What is the reason, other than preference :trollface:?

Partially preference, my mental-model, until working in this RFC, was that when I "archived" a delegation I was receiving the CBOR bytes of _only_ the delegation. Turns out that is not the case, and I am actually getting back an "envelope" containing a delegation and a "variant" block.

> My guess is fewer bytes (fair) and easier to encode/decode (mitigated by the existing utilities) - but maybe I'm missing something?

Yeah - I feel the size concern is real since we're already hitting HTTP header limits (4-8KB) in some cases, such as retrieval. I assume the variant block adds ~64 bytes (after encoding), though maybe it's not a big deal?

The "magic" in `delegation.Archive()` also tripped me up - the method name suggests serialization, which is true, but it's also:
1. Creating a variant block with type/version metadata
2. Linking to the actual delegation 
3. Packaging everything into a CAR

I think the envelope/variant pattern is valuable when you need version safety or are storing mixed content types. But when the context already tells you it's a delegation (like a header named `X-UCAN-Delegation`), the variant block feels redundant.

I suppose, going back to preference, what I'm looking for is optionality:
- `Archive()` for the full envelope with variant block (current behavior)
- `ArchiveRaw()` or similar for just the delegation blocks without envelope

This would let implementations choose based on their constraints - version safety when needed, minimal size when that matters more. Thoughts?

Peeja:
Note from the future: In UCAN 1.0, there's an officially specified envelope in the actual delegation/invocation/etc block which specifies that "variant" info. So, this question of "raw"-ness will disappear then. Granted, it does mean that there's _always_ some overhead for self-description, but personally I think it's worth it.

PR #54, opened 2025-06-03
rfc: proposal for standard private key handling

frrist:
:blue_book: [Preview](https://github.com/storacha/RFC/blob/7f2f55203adbf2d01fc197f6f88f4b3f24d19512/rfc/private-key-handling.md)

### DIFF
--- a/rfc/private-key-handling.md
+++ b/rfc/private-key-handling.md
@@ -1,0 +1,173 @@
+# RFC: Standardizing Private Key Material Handling
+
+## Summary
+
+This RFC proposes standardizing how we handle private key material across our services, adopting PEM format as the single standard for private key storage and configuration.
+
+## Problem Statement
+
+Our services currently accept private keys in multiple formats:
+
+- PEM files (with inconsistent encryption practices).
+- JSON files (non-standard, service-specific schemas).
+- Base64 encoded strings (raw key material, no metadata).
+- Multibase base64 padded strings (uncommon in standard tooling).
+
+This creates:
+
+- **Security inconsistency**: Different formats have varying security properties.
+- **Implementation complexity**: Each service implements multiple parsers.
+- **Cognitive overhead**: Developers must remember which format works where.
+- **Error-prone deployments**: Format mismatches cause configuration failures.
+
+## Proposed Solution: PEM Standard
+
+**Format**: PEM-encoded private keys using PKCS#8 format
+
+```
+-----BEGIN PRIVATE KEY-----
+MC4CAQA...
+-----END PRIVATE KEY-----
+```
+
+### Why PEM?
+
+1. **Universal tooling support**: OpenSSL, ssh-keygen, and all major crypto libraries support PEM natively.
+2. **Self-documenting format**: Clear headers (`-----BEGIN PRIVATE KEY-----`) identify content type.
+3. **Algorithm agnostic**: Supports our current Ed25519 keys and future algorithms (RSA, ECDSA, etc.).
+4. **Built-in security**: Native support for passphrase encryption.
+5. **Human readable**: Base64 encoded with clear delimiters for debugging.
+6. **Industry standard**: Developers already understand `.pem` files.
+
+### Usage Examples
+
+**Generating Ed25519 keys**:
+
+```bash
+# Generate Ed25519 private key
+openssl genpkey -algorithm ed25519 -out service-key.pem
+
+# View key details
+openssl pkey -in service-key.pem -text -noout
+
+# Extract public key
+openssl pkey -in service-key.pem -pubout -out service-pubkey.pem
+```
+
+**Securing keys**:
+
+```bash
+# Set proper permissions
+chmod 600 service-key.pem
+
+# Encrypt for storage
+openssl pkey -in service-key.pem -out service-key-encrypted.pem -aes256
+
+# Decrypt when needed
+openssl pkey -in service-key-encrypted.pem -out service-key.pem
+```
+
+**Future algorithm support** (no code changes needed):
+
+```bash
+# RSA
+openssl genpkey -algorithm rsa -pkeyopt rsa_keygen_bits:4096 -out rsa-key.pem
+
+# ECDSA
+openssl genpkey -algorithm ec -pkeyopt ec_paramgen_curve:P-256 -out ecdsa-key.pem
+```
+
+## Security Requirements
+
+1. **File permissions**: Private key files MUST have 600 permissions.
+2. **No environment variables**: Private keys MUST NOT be passed via environment variables.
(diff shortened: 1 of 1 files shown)


rfc/private-key-handling.md
```diff
@@ -0,0 +1,173 @@
...
+This RFC proposes standardizing how we handle private key material across our services, adopting PEM format as the single standard for private key storage and configuration.
+
+## Problem Statement
+
+Our services currently accept private keys in multiple formats:
+
+- PEM files (with inconsistent encryption practices).
+- JSON files (non-standard, service-specific schemas).
+- Base64 encoded strings (raw key material, no metadata).
+- Multibase base64 padded strings (uncommon in standard tooling).
```
alanshaw, line 14: Ok a few observations:

1. I believe you introduced PEM.
2. I don't know if we've ever actually used JSON to pass around private keys although I know the w3 CLI has the option to print them as JSON.
3. I don't believe we've ever passed plain base64 encoded strings around but maybe I'm incorrect.

So there was only really the 1 format prior - for which some of the reasons below don't apply.

I'm not trying to oppose the proposal here but I think if this was phrased more as "switch from multibase base64 padded private keys to PEM" then I think it would be more accurate.

rfc/private-key-handling.md
```diff
@@ -0,0 +1,173 @@
...
+
+```
+-----BEGIN PRIVATE KEY-----
+MC4CAQA...
+-----END PRIVATE KEY-----
+```
+
+### Why PEM?
+
+1. **Universal tooling support**: OpenSSL, ssh-keygen, and all major crypto libraries support PEM natively.
```
alanshaw, line 35: Yes, but typically we want to display the DID at the same time, so we have to dip back into our own tooling anyways.

rfc/private-key-handling.md
```diff
@@ -0,0 +1,173 @@
...
+2. **Self-documenting format**: Clear headers (`-----BEGIN PRIVATE KEY-----`) identify content type.
```
alanshaw, line 36: 🤷 the multibase encoded string is also self describing - the decoded bytes are also prefixed with the alogorithm. PEM has the advantage of being human readable to some extent.

rfc/private-key-handling.md
```diff
@@ -0,0 +1,173 @@
...
+3. **Algorithm agnostic**: Supports our current Ed25519 keys and future algorithms (RSA, ECDSA, etc.).
```
alanshaw, line 37: Again, same for existing format.

rfc/private-key-handling.md
```diff
@@ -0,0 +1,173 @@
...
+openssl genpkey -algorithm ec -pkeyopt ec_paramgen_curve:P-256 -out ecdsa-key.pem
+```
+
+## Security Requirements
+
+1. **File permissions**: Private key files MUST have 600 permissions.
+2. **No environment variables**: Private keys MUST NOT be passed via environment variables.
+3. **Encryption at rest**: Keys SHOULD be encrypted when stored.
+4. **Memory handling**: Keys SHOULD be loaded once at startup, not repeatedly read.
+5. **Secure deletion**: Key material SHOULD be securely overwritten when removed from memory.
```
alanshaw, line 86: We have historically used AWS SSM for deployed services and would continue to do so regardless of encoding (I understand this is not the same situation for storage nodes).

alanshaw (approved):
👍 I'm fine with this, for the reason of switching to PEM is mildy better from the standpoint of developer familiarity - I'm not convinced there's really any concrete advantages beyond that, which the existing format doesn't already support.

rfc/private-key-handling.md
frrist: You're absolutely right on all points - thank you for the clarification.

> I don't believe we've ever passed plain base64 encoded strings around but maybe I'm incorrect.

You're correct - looking at my shell history, we've been using multibase base64 padded strings, not plain base64.

> I believe you introduced PEM

Yes, I did introduce PEM to the storage node. The current state is:
- `storage id gen` returns JSON with a DID and multibase-base64-padded private key.
- I added `--type=pem` option for PEM output (needed for Curio's API authentication).
- I added `storage id parse` to convert between formats, and to extract DIDs from PEM files.

The PEM format was specifically introduced to work with Curio's API, which requires PEM-formatted public keys for authentication.

> I think if this was phrased more as "switch from multibase base64 padded private keys to PEM" then I think it would be more accurate

Yeah, the accurate framing is probably: **"Switch from multibase base64 padded private keys to PEM"**. JSON is only used as an output wrapper in some cases, not as a key format itself.

It's worth noting that at the end of this doc the need for custom tooling to extract DIDs from PEM files is outlined. Curious what your thoughts are on this.

rfc/private-key-handling.md
frrist: Yeah, mentioned that here: https://github.com/storacha/RFC/pull/54/files#diff-d7bd9ff16dbeeb26dbb100c755bc613b14dfd9603e87cbcd5a7e56a694586c9cR170

rfc/private-key-handling.md
frrist: My main concerns with multibase for private key material center around tooling, file conventions, and security. 

- Most common cryptographic tools (OpenSSL, ssh-keygen, keytool, etc.) have native support for PEM format when working with private keys. With multibase-encoded keys, we typically need custom tooling or conversion steps for common operations like key generation, inspection, or format conversion.

- PEM has well-established conventions - `.pem` files, standard permissions (600), and clear practices around key storage. With multibase strings, we'd need to define our own conventions: What file extension? Plain text files? JSON wrapper?

- PEM has built-in support for passphrase encryption. For multibase, we'd need to implement our own encryption scheme or wrap it in another format, essentially recreating what PEM already provides

At the end of the day, specifically for the storage node, I'd prefer to maintain key material in a file, rather than a string passed over the CLI or environment variable.

rfc/private-key-handling.md
```diff
@@ -0,0 +1,173 @@
...
+
+Users often need to derive a DID (Decentralized Identifier) from the private key for identity purposes. This section briefly covers how to extract the DID from a PEM-encoded private key.
+
+### What is a DID and How do we Encode ours?
+
+A DID is a decentralized identifier that corresponds to your public key. In our system, Ed25519 keys produce DIDs in the format: `did:key:...` where the suffix is a Base58BTC-encoded representation of the public key with multicodec prefixes.
+
+### Deriving DID from PEM
+
+At present, there does not exist any standard tooling to derive a DID used by our system from a PEM File containing a private key, or public key for that matter. This necessitates the creation of a custom CLI tool which allows users to derive a DID string from their public key.
```
frrist, line 172: @alanshaw assuming we align on PEM as our mechanism for handling key material, curious to hear your thoughts on these bits.

PR #55, opened 2025-06-04
rfc: DID:PLC integration for Storacha stack

fforbeck:
[Preview](https://github.com/storacha/RFC/blob/rfc/did-plc-integration/rfc/did-plc-integration.md)

This RFC is a proposal to integrate the `did:plc` method into Storacha stack for decentralized identity and Lit PKP encryption key management.

### DIFF
--- a/rfc/did-plc-integration.md
+++ b/rfc/did-plc-integration.md
@@ -1,0 +1,759 @@
+# RFC: DID:PLC Integration for Storacha Stack
+
+## Authors
+
+* [Felipe Forbeck](https://github.com/fforbeck) [Storacha Network](https://storacha.network/)
+
+## Abstract
+
+This RFC specifies the integration of did:plc (Public Ledger of Credentials) decentralized identifiers into the Storacha stack as an authorization method. The specification defines how users are authorized using cryptographically-controlled did:plc identities through public delegation retrieval, enabling decentralized identity management with a simplified, stateless approach.
+
+> **Note**: This RFC assumes that the `AccountDID` type definition has been updated to support `did:plc` identifiers alongside the existing `did:mailto` support. This foundational change is a prerequisite for all the integrations described in this document.
+
+## 1. Introduction
+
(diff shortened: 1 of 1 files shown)


rfc/did-plc-integration.md
```diff
@@ -0,0 +1,837 @@
...
+
+    User->>Client: Authenticate with existing did:plc
+    Client->>User: Private key available locally
+    Client->>Server: access/fetch (with: did:plc)
+    Server->>Storage: Find delegations for did:plc
+    Storage->>Server: Return delegations
+    Server->>Client: Return delegations
+    Client->>Client: Combine local did:plc→agent delegation with fetched did:plc delegations
+    Client->>User: Authorization complete
+```
```
Peeja, line 128: **suggestion:** I'd like to see us track the specific keys a little more clearly in this diagram. I'm still confused about how this works. How does the user/client prove their identity as `did:plc:[…]`? In other words, how do they issue and sign a delegation/invocation as that identity?

rfc/did-plc-integration.md
```diff
@@ -0,0 +1,837 @@
...
+### 2.1 New access/fetch Capability
+
+The core change is introducing a new `access/fetch` capability that allows public, unauthenticated retrieval of delegations for `did:plc` identifiers. This eliminates the need for authorization flows for did:plc accounts since delegations are not secrets - only the holder of the private key can use them.
+
+**Capability Definition**
+
+```javascript
+{
+  can: 'access/fetch',
+  with: 'did:plc:*'  // Only did:plc method supported
```
Peeja, line 143: **suggestion:** This won't work right; we'd need to have authority delegated from this DID. Instead, I think we want the DID we're looking for in the `nb`, while `with` should be the issuer of the invocation (the Agent). In other words, these should be self-signed.

Alternatively, we might avoid UCAN altogether for this. It's a little odd to me to use Ucanto as a way to do the equivalent of an unauthenticated `GET` request. Maybe it would make more sense to just expose an `HTTP` `GET` endpoint?

Peeja (changes requested):

PR #56, opened 2025-06-10
RFC: Client-Side Encryption

fforbeck:
[Preview](https://github.com/storacha/RFC/blob/rfc/client-side-encryption/client-side-encryption.md)

**Problem**: Our current encryption requires users to manage crypto wallets & has no fallback if Lit Protocol is down. This blocks mainstream adoption.

**Solution**: Add social login authentication (Google/Discord/etc) via Lit Providers + PKPs, and use our Gateway as fallback system that works when Lit is unavailable. Storacha pays all crypto costs transparently.

**Key Benefits**:
- **Zero crypto knowledge needed** - users authenticate with existing social accounts
- **Browser compatibility** - works in web apps, not just Node.js
- **Resilient architecture** - files accessible even during Lit outages  
- **Security improvements** - fixes current revocation gaps

**Impact**: Transforms encryption from "developer power-user feature" into mainstream privacy solution that competes with centralized providers on UX while maintaining security through client-side encryption.

### DIFF
--- a/client-side-encryption.md
+++ b/client-side-encryption.md
@@ -173,6 +173,17 @@ Provide multiple backup options for symmetric keys to ensure content remains acc
 - **Implementation**
   - **Storage Schema**: `{space_did}:{file_cid} → encrypted_symmetric_key`
 
+#### 2.6.2 Relationship to Lit Protocol Network Backup
+
+It's important to note that Lit Protocol itself implements [backup and recovery mechanisms](https://developer.litprotocol.com/security/backup-and-recover) at the network level using verifiable encryption, Recovery Party coordination, and Blinder protection. These network-level backups ensure the Lit Protocol itself can be restored in case of catastrophic node failures.
+
+Our proposed application-level backup strategy complements Lit Protocol's network backup by addressing different failure scenarios:
+
+- **Lit Protocol Backup**: Network-wide disaster recovery (nodes permanently offline)
+- **Storacha Backup**: Application-level availability (network unreachable, expensive, or slow)
+  
+This layered approach provides comprehensive resilience across both network infrastructure and application access patterns.
+
 ## 3. Enhanced System Flows
 
 The following sequence diagrams illustrate how the proposed enhancements integrate with the existing encryption system to provide improved user experience and reliability.
--- a/client-side-encryption.md
+++ b/client-side-encryption.md
@@ -6,9 +6,9 @@
 
 ## Abstract
 
-This RFC proposes enhancements in the client-side encryption library for Storacha uploads using Lit Protocol's Programmable Key Pairs (PKPs) which unlocks social authentication without private key management. The proposal maintains the secure per-file encryption approach while adding improved user experience and backup capabilities. Access control for symmetric key decryption continues to rely on Lit Actions that validate UCAN delegations, ensuring only authorized users can decrypt specific keys within their designated Storacha Spaces.
+This RFC proposes enhancements in the client-side encryption library for Storacha uploads using Lit Protocol's Programmable Key Pairs (PKPs), which unlock social authentication without private key management. The proposal maintains the secure per-file encryption approach while adding improved user experience and backup capabilities. Access control for symmetric key decryption continues to rely on Lit Actions that validate UCAN delegations, ensuring only authorized users can decrypt specific keys within their designated Storacha Spaces.
 
-It also proposes a fallback strategy to decrypt content in case the Lit Protocol becomes prohibitively expensive, experiences network issues or suffers service outages. The proposal covers Lit Protocol's Payment Delegation Database API for streamlined cost management.
+It also proposes a fallback strategy to decrypt content in case the Lit Protocol becomes prohibitively expensive, experiences network issues, or suffers service outages. The proposal covers Lit Protocol's Payment Delegation Database API for streamlined cost management.
 
 The goal is to establish the foundational encryption system for Storacha, prioritizing security through proven cryptographic practices while removing adoption barriers for mainstream users and enterprise customers.
 
@@ -18,7 +18,7 @@ The goal is to establish the foundational encryption system for Storacha, priori
 
 The Storacha stack currently provides client-side encryption through the encrypt-upload-client package, which uses symmetric encryption with Lit Protocol for key management. The existing system employs Lit Actions to validate UCAN (User Controlled Authorization Network) delegations for access control, ensuring users can only decrypt symmetric keys within their authorized Storacha Spaces. Currently, users must manage Ethereum wallets to authenticate with Lit Protocol and individually pay for transaction costs using [capacity credits](<https://developer.litprotocol.com/paying-for-lit/capacity-credits#:~:text=Capacity%20Credits%20are%20a%20key,period%20(e.g.%20one%20week)>).
 
-This RFC proposes extending the encryption system to provide a user-friendly approach using Lit PKPs for social authentication, while maintaining the same security model through client-side encryption and UCAN-based access control. The proposal introduces Lit Protocol's Payment Delegation Database API, enabling Storacha to act as the payer for all user Lit transactions, eliminating the need for users to manage cryptocurrency wallets or capacity credits. Additionally, the RFC introduces flexible backup strategies including Cloud KMS integration and gateway-based fallback decryption.
+This RFC proposes extending the encryption system to provide a user-friendly approach using Lit PKPs for social authentication, while maintaining the same security model through client-side encryption and UCAN-based access control. The proposal introduces Lit Protocol's Payment Delegation Database API, enabling Storacha to act as the payer for all user Lit transactions, eliminating the need for users to manage cryptocurrency wallets or capacity credits. Additionally, the RFC introduces flexible backup strategies, including Cloud KMS integration and gateway-based fallback decryption.
 
 ### 1.1 Current System Overview
 
@@ -114,7 +114,7 @@ The current access control system relies on Lit Actions to validate UCAN delegat
 
 ## 2. Proposed Enhancements
 
-This RFC addresses the identified limitations through a set of enhancements that maintain the security properties of the existing system while improving usability, reliability, and accessibility. The proposal introduces five key improvements:
+This RFC addresses the identified limitations through a set of enhancements that maintain the security properties of the existing system while improving usability, reliability, and accessibility. The proposal introduces six key improvements:
 
 ### 2.1 Lit PKP Integration with Social Authentication
 
@@ -135,7 +135,7 @@ Implement Lit Protocol's Payment Delegation Database API to enable Storacha to a
 
 #### 2.2.1 Payment Delegation Database Architecture
 
-The [Lit Protocol Payment Delegation Database](https://developer.litprotocol.com/paying-for-lit/payment-delegation-db) enables applications to pay for user operations without requiring users to hold capacity credits. The system works through an allowlist-based approach:
+The [Lit Protocol Payment Delegation Database](https://developer.litprotocol.com/paying-for-lit/payment-delegation-db) enables applications to pay for user operations without requiring users to hold capacity credits. The system works through a allowlist-based approach:
 
 - **Storacha as Payer**
 
@@ -151,8 +151,8 @@ The [Lit Protocol Payment Delegation Database](https://developer.litprotocol.com
 
 - **Operation Flow**
   1. User performs PKP-based operation (decryption, signing, etc.)
-  2. Lit Protocol checks if user's PKP address is in Storacha's allowlist
-  3. If authorized, operation proceeds and costs are charged to Storacha's capacity credits
+  2. Lit Protocol checks if the user's PKP address is in Storacha's allowlist
+  3. If authorized, operation proceeds, and costs are charged to Storacha's capacity credits
   4. User receives service without managing cryptocurrency or credits
 
 ### 2.3 Revocation System Implementation
@@ -165,7 +165,7 @@ Extend the encryption library to support browser environments by replacing Node.
 
 ### 2.5 Gateway Fallback Strategy
 
-Implement a backup decryption system through the Storacha IPFS Gateway that can validate UCAN delegations and decrypt content when Lit Protocol is unavailable, expensive, or experiencing performance issues. This reduces external network dependency while maintaining the same security model.
+Implement a backup decryption system through the Storacha IPFS Gateway that can validate UCAN delegations and decrypt content when the Lit Protocol is unavailable, expensive, or experiencing performance issues. This reduces external network dependency while maintaining the same security model.
 
 ### 2.6 Symmetric Key Backup Strategies
 
@@ -190,7 +190,7 @@ Our proposed application-level backup strategy complements Lit Protocol's networ
 
 - **Lit Protocol Backup**: Network-wide disaster recovery (nodes permanently offline)
 - **Storacha Backup**: Application-level availability (network unreachable, expensive, or slow)
-  
+
 This layered approach provides comprehensive resilience across both network infrastructure and application access patterns.
 
 ## 3. Enhanced System Flows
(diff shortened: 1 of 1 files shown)


client-side-encryption.md
```diff
@@ -0,0 +1,856 @@
...
+- **Mainstream Accessibility**: Email OTP and social authentication eliminate crypto wallet barriers when encryption is chosen
+- **Enterprise Ready**: Multiple compliance tiers capture enterprise market requirements
+- **Revenue Growth**: Encryption becomes profitable feature rather than cost center
+
+## 2. Proposed Enhancements
+
+This RFC addresses the identified limitations through a strategy that maintains the security properties of the existing system while improving usability, reliability, and enterprise sales opportunities. The proposal introduces **seven key improvements**:
+
+### 2.1 Hybrid Architecture: Lit Protocol + Google KMS
+
```
BravoNatalie, line 98: Why not give customers the option to integrate their own KMS solution?
I know it would add some complexity, but it could help reduce our role as a centralized party in the privacy setup, at least from the perspective of our direct customers. It might also lower the cost for them to use private data.

fforbeck: Great idea. I just don't know what it would require in terms of time and development effort. I believe we could start simple with one cloud provider’s KMS (e.g., Google KMS). Then, if demand grows, incrementally add Bring Your Own Key as a premium enterprise feature.

client-side-encryption.md
```diff
@@ -0,0 +1,856 @@
...
+## Abstract
+
+This RFC proposes a strategy for Storacha to transform the privacy solution into a revenue-generating feature. The strategy establishes **encryption as a paid premium feature** available to Lite and Business users, while maintaining standard public storage for free Starter users.
+
+**Hybrid Architecture**: Paid users can choose from multiple encryption tiers:
+
+- **Privacy Tier**: Lit Protocol end-to-end encryption for maximum privacy
+- **Compliance Tier**: Google KMS with enterprise certifications (SOC 2, FIPS 140-2, ISO 27001)
+- **Hybrid Tier**: Lit Protocol + KMS backup for redundancy
```
alanshaw, line 49: Suggestion: offer compliance/hybrid to business plan holders only.

client-side-encryption.md
```diff
@@ -0,0 +1,856 @@
...
+Add support for Lit Protocol's Programmable Key Pairs (PKPs) alongside existing wallet-based authentication. PKPs enable social login providers (Google, Discord, WebAuthn, Stytch, etc) for users who prefer not to manage Ethereum wallets and private keys, while maintaining backward compatibility with existing wallet-based workflows.
+
+> Programmable Key Pairs (PKPs) are ECDSA public/private key pairs created by the Lit network using Distributed Key Generation (DKG). Each Lit node holds a share of the private key, and more than two-thirds of these shares must be collected to execute a given action (i.e. signing a transaction).
+> ref: https://developer.litprotocol.com/user-wallets/pkps/overview
+
+#### 2.2.1 Email Authentication via Stytch Integration
+
+Email and SMS authentication provides users with a convenient way to verify their identity using one-time passwords (OTP) sent to their registered email address or phone number.
+
+Since all Storacha users already have email addresses, we can generate PKPs using [Stytch](https://stytch.com/docs/api/send-otp-by-sms) with email authentication.
```
alanshaw:
```suggestion
Since all Storacha users already have email addresses, we can generate PKPs using [Stytch](https://stytch.com/docs/api/send-otp-by-email) with email authentication.
```

client-side-encryption.md
```diff
@@ -0,0 +1,856 @@
...
+  - **Monitoring & Auto-Replenishment**:
+    - Implements automated monitoring of capacity credit balance
+    - Triggers alerts when balance falls below configurable thresholds (e.g., 20% remaining)
+    - Supports automatic top-up functionality to prevent service interruption
+    - Maintains audit logs of all funding operations for financial tracking
+    - :warning: This automation would be implemented in another iteration and won't be covered in detail in this RFC
+
+- **User Registration Process**
+
+  - When users create PKPs through social authentication, their PKP addresses are automatically added to Storacha's allowlist
```
alanshaw, line 229: Can you speak to how this works? How does Lit know a minted PKP should be added to our allowlist?

client-side-encryption.md
```diff
@@ -0,0 +1,856 @@
...
+The specification for this new endpoint is detailed in the following ticket https://github.com/storacha/upload-service/issues/177
+
+```mermaid
+sequenceDiagram
+    participant User
+    participant Client as "Storacha Client"
+    participant RevocationAPI as "Revocation API"
+    participant LitProtocol as "Lit Protocol"
+    participant Gateway
+    participant GoogleKMS as "Google KMS"
```
alanshaw: Unnecessary - the quotes appear in the diagram.

```suggestion
    participant Client as Storacha Client
    participant RevocationAPI as Revocation API
    participant LitProtocol as Lit Protocol
    participant Gateway
    participant GoogleKMS as Google KMS
```

client-side-encryption.md
```diff
@@ -0,0 +1,856 @@
...
+
+- **Operation Flow**
+  1. User performs PKP-based operation (decryption, signing, etc.)
+  2. Lit Protocol checks if the user's PKP address is in Storacha's allowlist
+  3. If authorized, operation proceeds, and costs are charged to Storacha's capacity credits
+  4. User receives service without managing cryptocurrency or credits
+
+### 2.4 Revocation System Implementation
+
+Implement a new endpoint for revocation checking that validates delegation status during both Lit Action execution and gateway-based decryption. This closes the security gap where revoked delegations continue to grant access.
```
alanshaw, line 241: The gateway based decryption has been pitched as a fallback if Lit is unavailable up until this point. Is gateway decryption a feature in itself? Can the gateway use Lit also? Why would the gateway not just return the decrypted DEK and let the client do the decryption as before?

So the question is whether "gateway" needs to be a part of this or if we just run a "trusted Lit fallback server" that you make the same request to, should the Lit request fail.

Separately, how does the key get in Google KMS as well as Lit? How does this fallback work?

client-side-encryption.md
```diff
@@ -0,0 +1,856 @@
...
+    alt Lit Protocol Unavailable
+        User->>Gateway: Request file with UCAN invocation
+        Gateway->>RevocationAPI: Check delegation status
+
+        alt UCAN Valid and Not Revoked
+            RevocationAPI-->>Gateway: ✅ Access authorized
+            Gateway->>GoogleKMS: Decrypt DEK with space key
+            GoogleKMS-->>Gateway: Return decrypted DEK
+            Gateway-->>Gateway: Decrypt file with DEK
+            Gateway-->>User: Stream decrypted content
```
alanshaw, line 282: This bit makes me nervous...

We're holding decryption keys for extended periods of time, and sending decrypted bytes over the network.

client-side-encryption.md
```diff
@@ -0,0 +1,856 @@
...
+
+- We can use AES-CTR with pseudo-streaming (buffering chunks before emitting) for simplicity and streaming support.
+- AES-CTR allows chunked processing without padding, making it suitable for large files and browser environments.
+- The Web Crypto API supports AES-CTR natively in all modern browsers and in Node.js 19+ as globalThis.crypto.
+- For Node.js <19, we must polyfill globalThis.crypto (e.g., with `node --experimental-global-webcrypto` or a package like @peculiar/webcrypto).
+- This allows for processing large files in chunks with no padding issues found in other libraries such as node-forge.
+
+Note: The [existing implementation](https://github.com/storacha/upload-service/blob/0d7a771d77d0433934901878684123a5e733a4f3/packages/encrypt-upload-client/src/crypto-adapters/browser-crypto-adapter.js) is currently pseudo-streaming - it buffers all encrypted/decrypted chunks before emitting them as a stream. For true streaming (lower memory usage), we need to refactor it to emit each chunk as soon as it is processed.
+
+### 2.6 Gateway Fallback Strategy
```
alanshaw, line 308: Again, I don't really understand how the fallback bit works - how does the key get in both Lit _and_ Google?

Ah I see it comes later in the document. Maybe link to it from here?

alanshaw: Can we segregate this from our public gateway? Like `bafy.ipfs.private.storacha.link`?

client-side-encryption.md
```diff
@@ -0,0 +1,856 @@
...
+    else Delegation Revoked
+        RevocationAPI-->>LitProtocol: ❌ Delegation revoked
+        LitProtocol-->>Client: Access denied
+        Client-->>User: Decryption failed
+    end
+
+    Note over User, GoogleKMS: Revocation Check - Gateway Implementation
+
+    alt Lit Protocol Unavailable
+        User->>Gateway: Request file with UCAN invocation
```
alanshaw, line 274: I guess you are looking for this to land - https://github.com/storacha/RFC/pull/60

client-side-encryption.md
```diff
@@ -0,0 +1,856 @@
...
+  "litProtocol": {
+    "identityBoundCiphertext": "pkp_encrypted_dek_ciphertext",
+    "plaintextKeyHash": "hash_of_plaintext_dek",
+    "accessControlConditions": [...],
+    "chain": "ethereum"
+  },
+  "googleKMS": {
+    "keyName": "projects/PROJECT/locations/LOCATION/keyRings/RING/cryptoKeys/SPACE_ID",
+    "encryptedDEK": "kms_encrypted_dek"
+  }
```
alanshaw, line 377: You might want to put these in an array to imply a desired ordering?

client-side-encryption.md
```diff
@@ -0,0 +1,856 @@
...
+  "algorithm": "AES-256-CTR",
+  "litProtocol": {
+    "identityBoundCiphertext": "pkp_encrypted_dek_ciphertext",
+    "plaintextKeyHash": "hash_of_plaintext_dek",
+    "accessControlConditions": [...],
+    "chain": "ethereum"
+  },
+  "googleKMS": {
+    "keyName": "projects/PROJECT/locations/LOCATION/keyRings/RING/cryptoKeys/SPACE_ID",
+    "encryptedDEK": "kms_encrypted_dek"
```
alanshaw, line 376: What is it that stops a user going direct to google to get the key?

client-side-encryption.md
```diff
@@ -0,0 +1,856 @@
...
+    Note over User, GoogleKMS: Revocation Check - Lit Implementation
+
+    User->>Client: Request file decryption
+    Client->>LitProtocol: Execute Lit Action with UCAN delegation
+
+    LitProtocol->>RevocationAPI: Check delegation status
+
+    alt Delegation Valid
+        RevocationAPI-->>LitProtocol: ✅ Active delegation
+        LitProtocol-->>Client: Return decrypted DEK
```
alanshaw, line 263: Should this be encrypted with the issuer public key?

client-side-encryption.md
```diff
@@ -0,0 +1,856 @@
...
+The following sequence diagrams illustrate how the proposed enhancements integrate with the existing encryption system to provide improved user experience, reliability, and flexible backup options.
+
+### 4.1 Enhanced Encryption Flow (Dual-Path)
+
+```mermaid
+sequenceDiagram
+    participant User
+    participant StorachaClient
+    participant CryptoAdapter
+    participant LitSDK
```
alanshaw, line 674: What is the difference between `LitSDK` and `Lit Protocol` in these diagrams?

alanshaw (commented):
Do we need to restrict gateway access to Google KMS? Can it also do the Lit flow?

fforbeck:
> Do we need to restrict gateway access to Google KMS? Can it also do the Lit flow?

Gateway probably shouldn't decrypt via Lit. The idea is exactly to have a separate, reliable path (via Google KMS) if Lit is down or pricey. Also, mixing Lit into Gateway might complicate compliance stuff for enterprise customers. I think we would need more research to confirm this.

client-side-encryption.md
fforbeck: I think it makes sense to offer the Compliance tier for Business plan holders only, but the Hybrid tier could be beneficial for Lite users as well, because they will have Lit and may want a fallback strategy that doesn't rely on Lit only. WDYT?

client-side-encryption.md
fforbeck: This is described in section `2.3 Payment Delegation System`, but every time a new user is onboarded, we need to send a request to the Payment Delegation System API to include that new PKP as a payee that will consume credits from the Storacha Payer account.

client-side-encryption.md
fforbeck: > Is gateway decryption a feature in itself?

Yes.

> Can the gateway use Lit also? 

Gateway could technically use Lit, but the whole point is to keep it as a reliable fallback separate from Lit (in case Lit becomes expensive or unstable).

> Why would the gateway not just return the decrypted DEK

Great idea 🔥 We could still do client-side decryption as before, and it would preserve end-to-end encryption. I think it would cut the costs and simplify the decryption flow, the only thing that we would need to implement would be the Access Control - to return the DEK only if that request is authorized to decrypt the content. Also, for enterprise customers, we would need to check the compliance requirements. I liked this approach, sounds better, and we could return the DEK encrypted with the issuer's public key.

client-side-encryption.md
fforbeck: Yes, I'm skeptical of this approach, but your suggestion to return the DEK makes things simpler and safer, as we can encrypt the DEK with the decryptor's public key before we transfer it.

client-side-encryption.md
fforbeck: Good call. Yes, makes sense to have it running in a separate domain.

client-side-encryption.md
fforbeck: Sweet! That's exactly what we need - I was thinking about limiting this usage to the CLI and SDK, but with this RFC in place, we can enable it for direct HTTP requests. Super cool!

client-side-encryption.md
fforbeck: `projects/PROJECT/locations/LOCATION/keyRings/RING/cryptoKeys/SPACE_ID` is just a pointer, not a credential or access token. We control the KMS, and we need to grant access via IAM credentials for our services to access it, so the users can't fetch it.

client-side-encryption.md
fforbeck: YES! Good call!

client-side-encryption.md
fforbeck: Same thing. I just missed that after several updates.

client-side-encryption.md
BravoNatalie: I think that if this is something we might want in the future, then during development we could prioritize implementing it with an interface, and use the Google KMS solution as an adapter. That way, it keeps things flexible and it shouldn’t require too much effort.

client-side-encryption.md
```diff
@@ -0,0 +1,856 @@
...
+- **Use Case**: Paid users who need specific privacy/compliance requirements without disrupting standard workflows
+- **Public Sharing**: Unencrypted files remain easily shareable and collaborative
+
+### 2.2 Lit PKP Integration
+
+Add support for Lit Protocol's Programmable Key Pairs (PKPs) alongside existing wallet-based authentication. PKPs enable social login providers (Google, Discord, WebAuthn, Stytch, etc) for users who prefer not to manage Ethereum wallets and private keys, while maintaining backward compatibility with existing wallet-based workflows.
+
+> Programmable Key Pairs (PKPs) are ECDSA public/private key pairs created by the Lit network using Distributed Key Generation (DKG). Each Lit node holds a share of the private key, and more than two-thirds of these shares must be collected to execute a given action (i.e. signing a transaction).
+> ref: https://developer.litprotocol.com/user-wallets/pkps/overview
+
```
BravoNatalie, line 124: I think this definition is a bit confusing, especially since the docs also mention that PKPs are NFTs.
So I would also add this:

> The PKP is represented as an NFT, and whoever owns this NFT controls the key pair and can set custom signing logic or access rules.
> ref: https://developer.litprotocol.com/user-wallets/pkps/quick-start#mint-a-pkp-and-add-permitted-scopes

client-side-encryption.md
```diff
@@ -0,0 +1,856 @@
...
+> Reference: https://developer.litprotocol.com/paying-for-lit/payment-delegation-db#the-payment-delegation-database
+
+The system works through a allowlist-based approach:
+
+- **Storacha as Payer**
+
+  - **Registration**: Storacha registers as a payer with its own PKP wallet and capacity credits
+  - **Key Management**: Maintains a secure master payer secret key for delegation database operations
+  - **Centralized Funding**: Funds capacity credits for all user operations through a single account, eliminating individual user payment complexity
```
BravoNatalie, line 219: I understand that we’re going to sponsor their usage, but I was wondering, beyond the plans that include access to private data, do we have a defined quota for capacity credits included in each plan? And will there be an extra charge if the user exceeds that limit?
Basically, what’s the maximum amount of capacity credits we’re allowing per plan?

PR #57, opened 2025-06-13
rfc(indexer): Do not fail query when writing to cache

alanshaw:
Proposes returning query data to the client even if cache write operation fails.

[📚 Preview](https://github.com/storacha/RFC/blob/rfc/indexer-do-not-fail-query-when-write-to-cache/rfc/indexer-do-not-fail-query-when-write-to-cache.md)


volmedo (approved):
I agree with this. Given that the indexer is essentially a cache in front of IPNI there is no point in it failing the request completely if the data has already been fetched.

For now, I think the easiest way to avoid the issue going unnoticed is reporting the error to Sentry.

hannahhoward (approved):
LGTM just get it to sentry

PR #60, opened 2025-06-16
rfc: Ucanto transport in HTTP headers

alanshaw:
This RFC attempts to define a Ucanto style transport that allows invocations and receipts to be communicated in HTTP headers, leaving the HTTP request/response body usable for alternative purposes.

The idea is to use this proposal to allow us to send a `space/content/retrieve` invocation in the HTTP headers of a HTTP GET request. The response body is subsequently used to return the requested bytes in a single round trip (for the majority of requests), while allowing egress to be accounted for via signed invocations/receipts.

[📚 Preview](https://github.com/storacha/RFC/blob/rfc/http-header-transport/rfc/http-header-transport.md)

[🎬 Implementation Demo](https://youtu.be/11np-cGTe48?si=kw88R1DAlMSq-b1T)

### DIFF
--- a/rfc/http-header-transport.md
+++ b/rfc/http-header-transport.md
@@ -1,0 +1,37 @@
+# Ucanto transport in HTTP headers
+
+This RFC is an extension of [resourceful UCAN](https://github.com/storacha/RFC/blob/rfc/resourceful-ucan/rfc/resourceful-ucan.md), attempting to further formalize the protocol.
+
+Ideally we'd be able to authorize ourselves to fetch content by simply issuing a HTTP GET request. This is preferable from a caching perspective, but also because it fits with existing content claims (location commitments) that are currenty retrieved as such.
+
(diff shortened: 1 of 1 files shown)


hannahhoward (commented):
My biggest question is the format. Given the size limit challenges, should we consider:

- gzipping the archive
- using a thinner archive format than car, like https://github.com/ucan-wg/go-ucan/blob/main/pkg/container/SPEC.md (note: compare perf here -- https://github.com/ucan-wg/go-ucan/tree/main/pkg/container)

alanshaw:
I'm pretty easy on the format. In my testing this wasn't much more than 1kb using the existing format and less when gzipped (for a typical chain of invocation plus 2x delegation chain). I'm cool with compressing, would be nice to not have to introduce another format...and we can always iterate...

hannahhoward:
@alanshaw gzip is my priority, if you want to skip the container format I'm not super attached.

frrist (approved):
:shipit:

PR #61, opened 2025-06-25
rfc: proposing switch to uber.fx DI framework

frrist:
:blue_book: [Preview](https://github.com/storacha/RFC/pull/61/files?short_path=953e4d7#diff-953e4d7ba388e7acaeef22e33507b5f5cda7f8e22ab923d3e313f35bd29d0d2c)

### DIFF
--- a/rfc/fx-dep-injection-adoption.md
+++ b/rfc/fx-dep-injection-adoption.md
@@ -1,0 +1,428 @@
+# RFC: Adopting Uber FX for Dependency Injection in Piri
+
+**Author**: Forrest (@frrist)
+
+**Date**: 2025-06-25 
+
+**Status**: Draft
+
+## Summary
+
+This RFC proposes adopting Uber's FX dependency injection framework for the Piri project to address current pain points in dependency management, improve testability, and enhance maintainability as the codebase grows.
+
+## Background
+
+The Piri project currently uses manual dependency wiring throughout its codebase. As demonstrated in our proof-of-concept (see [`demofx`](https://github.com/frrist/demofx) repository), this approach leads to several challenges:
+
+1. **Complex initialization sequences** - Dependencies must be created in exact order
+2. **Propagation of constructor changes** - Adding a new dependency requires updating all constructor calls
+3. **Manual lifecycle management** - Start/stop sequences must be carefully orchestrated
+4. **Testing complexity** - Mocking dependencies requires extensive setup
+
+## Current State Analysis
+
+### Example 1: AWS Service Construction
+
+The AWS service construction in `pkg/aws/service.go` shows extreme complexity with 434 lines of manual wiring:
+
+```go
+func Construct(cfg Config) (storage.Service, error) {
+    // Manual creation of 15+ AWS service clients and stores
+    blobStore := NewS3BlobStore(cfg.Config, cfg.BlobStoreBucket, formatKey, blobStoreOpts...)
+    allocationStore := NewDynamoAllocationStore(cfg.Config, cfg.AllocationsTableName, cfg.DynamoOptions...)
+    claimStore, err := delegationstore.NewDelegationStore(NewS3Store(...))
+    ipniStore := NewS3Store(cfg.Config, cfg.IPNIStoreBucket, cfg.IPNIStorePrefix, cfg.S3Options...)
+    chunkLinksTable := NewDynamoProviderContextTable(...)
+    metadataTable := NewDynamoProviderContextTable(...)
+    publisherStore := store.NewPublisherStore(ipniStore, chunkLinksTable, metadataTable, ...)
+    // ... and many more
+    
+    // Complex conditional logic for optional features
+    if cfg.SQSPDPPieceAggregatorURL != "" && cfg.CurioURL != "" {
+        pdp, err := NewPDP(cfg)
+        // ... error handling and more setup
+    }
+    
+    // Manual option building
+    opts := []storage.Option{
+        storage.WithIdentity(cfg.Signer),
+        storage.WithBlobstore(blobStore),
+        storage.WithAllocationStore(allocationStore),
+        // ... 20+ more options
+    }
+    
+    return storage.New(opts...)
+}
+```
+
+### Example 2: Lambda Function Complexity
+
+Lambda functions in `cmd/lambda/*/main.go` suffer from heavyweight initialization:
+
+```go
+func makeHandler(cfg aws.Config) (http.Handler, error) {
+    // This constructs EVERYTHING even if lambda only needs blob store
+    service, err := aws.Construct(cfg)
+    if err != nil {
+        return nil, err
+    }
+    
+    // Only uses 3 components out of 20+ created
+    handler := blobs.NewBlobPutHandler(
+        service.Blobs().Presigner(), 
+        service.Blobs().Allocations(), 
+        service.Blobs().Store()
+    )
+    return handler, nil
+}
+```
+
+### Example 3: UCAN Server Setup
+
+The UCAN server in `cmd/cli/serve/ucan.go` shows 329 lines of manual dependency wiring:
+
+```go
+func startServer(cmd *cobra.Command, _ []string) error {
+    // Manual creation of 6 different datastores
+    allocDs, err := leveldb.NewDatastore(allocsDir, nil)
+    claimDs, err := leveldb.NewDatastore(claimsDir, nil)
+    publisherDs, err := leveldb.NewDatastore(publisherDir, nil)
+    receiptDs, err := leveldb.NewDatastore(receiptDir, nil)
+    // ... plus conditional PDP datastore
+    
+    // Complex configuration parsing and validation
+    uploadServiceDID, err := did.Parse(cfg.UploadServiceDID)
+    uploadServiceURL, err := url.Parse(cfg.UploadServiceURL)
+    indexingServiceDID, err := did.Parse(cfg.IndexingServiceDID)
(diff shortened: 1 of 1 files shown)


Peeja (approved):
(Should this be marked Ready for Review?)

LGTM! This looks like a huge win. I probably don't have enough Go experience to be a strong approval, though.

Question, for my own curiosity, being new to Fx: It looks like Fx doesn't do _deep_ injection, is that right? Like, your top-level, `Invoke()`d thing gets dependencies, which themselves get initialized recursively on demand, but then it's up to that `Invoke()`d code to pass things deeper into the stack? I know some DI systems give you a way to reach up from deep in the stack to grab things you need.

Despite a bit of (I think appropriate) magic, I think Fx looks extremely straightforward and clean.

alanshaw (approved):
❤️ This looks awesome. Lets do it!

volmedo (commented):
thanks for this @frrist.

As I stated in the related PR, I understand the reasons behind the necessity of introducing a DI framework and the benefits it provides. The RFC reinforces and illustrates these benefits and advantages.

My main concern is not whether we should use a DI framework or not, but rather which one we should choose. I tend to dislike frameworks and libraries that do too much magic under the hood, as that usually ends up hindering the developer's ability to debug, adapt, optimize and/or change the codebase in the future, depending on what they are trying to do.

Because of that, I feel inclined to use Google Wire because of its simplicity. Instead of using reflection, Wire is based on code generation. When dependencies are specified, the tool generates plain Go code that can be read and makes sense (hopefully).

The main two drawbacks I see of using Wire are:
1. It doesn't offer support for lifecycle management. I don't think this is critical, as implementing start and shutdown functions manually is not that different from using FX's lifecycle hooks.
2. Forrest already knows Uber FX. That knowledge can be shared with the rest of the team for a smoother transition. This could compensate with Wire's learning curve being not as steep as FX's one though.

All in all, my preference falls slightly on the side of Google Wire (mainly because it adds less magic), but I don't see critical reasons to not go with Uber FX instead if other folks think it's a better option. Just wanted to raise awareness of alternatives to spark discussion and learn what others think.

alanshaw:
@volmedo just so you know you're heard - your points are totally valid and correct and I appreciate them, from my point of view though I don't have experience with any DI framework in Go yet, so can't comment on the relative merits of either (or other alternatives). I don't have bandwidth to survey the landscape but I have used other Go code that has come from Uber and been happy with it so I have a sense of the likely quality. I'm also kinda willing to trust that this will allow @frrist to go fast with a framework he knows and recommends.

volmedo:
@alanshaw your comment is appreciated.

I lack experience using any of these frameworks too and my ideas are mostly based on quick AI-assisted research.

I agree on the quality of Go libraries produced by Uber, but it would be very surprising to me that Google's framework is lacking on that front, with them being the inventors of the language. It's not quality that concerns me, but the amount of magic each of them add. Solving issues that may arise is harder when one doesn't really understand what's going on under the hood.

And yeah, of course it will allow Forrest to go fast since he already knows the framework. My main concern in regard to this is the rest of us and potential new devs. From my little research, it looks like the learning curve in FX is steeper, and I wouldn't want that to become an obstacle to future contributors.

In any case, I'm totally open to disagree and commit on this. It is possible we never face any issues, and my main concerns are related to scenarios that may never occur. While using FX optimizes for pressing issues we have now, such as getting a production-ready version of Piri out as soon as possible.

So all good from my side. Again, I just wanted to make sure we are intentional about the decision and its tradeoffs.

frrist:
> It looks like Fx doesn't do deep injection, is that right? Like, your top-level, Invoke()d thing gets dependencies, which themselves get initialized recursively on demand, but then it's up to that Invoke()d code to pass things deeper into the stack?

@Peeja yup, that's right - FX automatically injects dependencies into the functions you register with it (`Provide` and `Invoke`), but it doesn't reach deep into your application code. Once you're in your regular application logic, you pass dependencies down explicitly through normal Go parameters; FX wires things together by matching types between what _providers_ return and what _constructors_ need:

Here's a short example:
```go
app := fx.New(
    // Step 1: Provide a config
    fx.Provide(func() *Config {
        return &Config{BucketName: "my-bucket"}
    }),
    
    // Step 2: Provide S3 client (needs config)
    fx.Provide(func(cfg *Config) *S3Client {
        // FX sees this function needs *Config, finds it above
        return s3.NewClient(cfg.Region)
    }),
    
    // Step 3: Provide blob store (needs both client and config)
    fx.Provide(func(client *S3Client, cfg *Config) *BlobStore {
        // FX finds both dependencies automatically
        return NewBlobStore(client, cfg.BucketName)
    }),
    
    // Step 4: FX injects here at the top level only
    fx.Invoke(func(lc fx.Lifecycle, store *BlobStore) {
        // FX built the whole chain: Config → S3Client → BlobStore
        handler := blobs.NewBlobPutHandler(store)
        
        // From here down, you pass dependencies explicitly
        server := &http.Server{Handler: handler, Addr: ":8080"}
        
        lc.Append(fx.Hook{
            OnStart: func(ctx context.Context) error {
                return server.ListenAndServe()
            },
            OnStop: func(ctx context.Context) error {
                return server.Shutdown(ctx)
            },
        })
    }),
)

// Dependencies are built and lifecycle hooks registered, but nothing running yet
app.Start(context.Background())  // OnStart hooks run here
app.Stop(context.Background())   // OnStop hooks run here
```

But `Invoke` isn't required here, alternatively we could do:
```go
app := fx.New(
    // Same providers as before
    fx.Provide(func() *Config {
        return &Config{BucketName: "my-bucket"}
    }),
    
    fx.Provide(func(cfg *Config) *S3Client {
        return s3.NewClient(cfg.Region)
    }),
    
    fx.Provide(func(client *S3Client, cfg *Config) *BlobStore {
        return NewBlobStore(client, cfg.BucketName)
    }),
    
    // Replace fx.Invoke with fx.Provide that returns the server
    fx.Provide(func(lc fx.Lifecycle, store *BlobStore) *http.Server {
        handler := blobs.NewBlobPutHandler(store)
        server := &http.Server{Handler: handler, Addr: ":8080"}
        
        lc.Append(fx.Hook{
            OnStart: func(ctx context.Context) error {
                return server.ListenAndServe()
            },
            OnStop: func(ctx context.Context) error {
                return server.Shutdown(ctx)
            },
        })
        
        return server
    }),
    
    // No fx.Invoke needed - the Provide above handles everything
)

app.Start(context.Background())  // Still works the same way
app.Stop(context.Background())   // same same
``` 

The key difference is that with `Provide`, you must return something (the `*http.Server`), while `Invoke` is purely for side effects. But the lifecycle registration works identically in both cases.
Most of the time we'll use `Invoke` for application startup since we usually wouldn't need the server object in our dependency graph, but both approaches work.

frrist:
@volmedo great points - 

On the magic front, I had similar concerns/reaction when I first encountered FX. But I find its approach is actually quite transparent. When something goes wrong, it gives you clear error messages like 'missing dependency: *BlobStore for *Handler' and you can visualize the dependency graph - it has dot graph support, too for debugging. So the 'magic' is fairly bounded and debuggable imho.

I agree FX has a bit of a learning curve initially. But I'd also argue the current manual wiring in our setup code is already complex and challenging to navigate. FX replaces that complexity with a more structured, documented pattern. 

Preference for Wire's code generation approach makes total sense - you can literally see the generated initialization code, step through it in a debugger, and understand exactly what's happening. Though here's an interesting counterpoint: Wire arguably has more lock-in than FX because it generates code that becomes part of our codebase. We'll need the Wire tool in our build pipeline, and our injector functions would use Wire-specific syntax like [`wire.Build()`](https://go.dev/blog/wire#how-does-it-work). If we wanted to remove Wire later, we'd need to either keep the generated code (defeating the purpose) or rewrite all the injection logic.

With FX, the provider functions are just normal Go functions - few framework-specific annotations (we'll need _some_ fx specific struct tags for sure), and no generated code. The only FX-specific parts are the struct tags, module definitions and app setup. So if we later decided the "magic" was too much, we could swap frameworks while keeping all our provider functions intact.

In a sense, FX essentially becomes a forcing function for good architecture. The primary value isn't FX or DI itself - it's breaking apart our monolithic constructors into focused, single-purpose provider functions. Instead of some 200+ line constructor, we'd have clean provider functions.

This modular approach would work just as well with Wire if we ever wanted to switch. According to the [official Go blog post about Wire](https://go.dev/blog/wire), Wire's dependency graph definition states:
- _Providers are ordinary Go functions that “provide” values given their dependencies, which are described simply as parameters to the function_
- _Injectors are generated functions that call providers in dependency order. You write the injector’s signature, including any needed inputs as arguments, and insert a call to wire.Build with the list of providers or provider sets that are needed to construct the end result_

So, Wire actively encourages the same single-purpose provider pattern we'd use with FX. And since both frameworks encourage the same architectural pattern, the refactoring work benefits either approach.

volmedo:
thank you very much @frrist, interesting arguments.

Thanks for letting me know about FX's debuggability, it definitely helps making me a bit more comfortable and less concerned on that front.

I think your point on vendor lock-in is reasonable. I'd argue, though, that I feel it's highly unlikely this can become an issue. At least in the traditional sense of the "vendor" starting to do nasty things that cannot be avoided due to the high cost of getting rid of the lock-in. I think assuming that we'll never have the need to remove this code or swap the framework is a reasonable assumption.

In any case, this discussion is good enough of a proof for me to confirm that different approaches have been considered, thank you all for sharing your different inputs and perspectives.

volmedo (approved):

PR #64, opened 2025-08-14
rfc: new usage capability proposal

BravoNatalie:
[Preview](https://github.com/storacha/RFC/pull/64/files#diff-431973756864813137effe3cb6fac091cdc228d3c0f1a4934dc375f12ce314f0)

### DIFF
--- a/rfc/account-usage-get-capability.md
+++ b/rfc/account-usage-get-capability.md
@@ -50,7 +50,7 @@ type AccountUsageGetNB struct {
 type Period struct {
   from Int # inclusive
   to Int # inclusive
-} representation tuple
+}
 ```
 
 > example: getting the total usage
@@ -85,7 +85,10 @@ type Period struct {
       "can": "account-usage/get",
       "nb": {
           "spaces": ["did:key:z6MkuxVKbEvYzXw89c9ESd3xoZ988MFrCgqT5JF5wtBvuYWe"],
-          "period": [1740357624, 1740357624]
+          "period": {
+            "to": 1758111728,
+            "from": 1754006400
+          },
       }
     }
   ],
--- a/rfc/account-usage-get-capability.md
+++ b/rfc/account-usage-get-capability.md
@@ -133,13 +133,13 @@ type AccountUsageGetFailure {
 }
 
 type AccountUsageGetSuccess {
-  total        Int
-  providers    Record<ProviderDID, ProviderUsage>   # keys MUST be sorted
+  total       Int
+  spaces      Record<SpaceDID, SpaceUsage>   # keys MUST be sorted
 }
 
-type ProviderUsage {
-  total     Int
-  spaces    Record<SpaceDID, UsageData>    # keys MUST be sorted
+type SpaceUsage {
+  total       Int
+  providers   Record<ProviderDID, UsageData>    # keys MUST be sorted
 }
 
 # UsageData is already defined and used by `usage/report`
@@ -173,7 +173,7 @@ type DID = string
 
 ```
 
-In all responses, the keys of the `providers` field in `AccountUsageGetSuccess` and the `spaces` field in `ProviderUsage` **MUST be sorted lexicographically** by their respective key (ProviderDID, SpaceDID). This ensures that the same query produces the same output each time.
+In all responses, the keys of the `spaces` field in `AccountUsageGetSuccess` and the `providers` field in `SpaceUsage` **MUST be sorted lexicographically** by their respective key (SpaceDID, ProviderDID). This ensures that the same query produces the same output each time.
 
 
 > example:
@@ -181,11 +181,11 @@ In all responses, the keys of the `providers` field in `AccountUsageGetSuccess` 
 ```json
 {
   "total": 5356848797,
-  "providers": {
-    "did:web:web3.storage": {
+  "spaces": {
+    "did:key:z6MkuxVKbEvYzXw89c9ESd3xoZ988MFrCgqT5JF5wtBvuYWe": {
       "total": 5356848797,
-      "spaces" {
-        "did:key:z6MkuxVKbEvYzXw89c9ESd3xoZ988MFrCgqT5JF5wtBvuYWe": {
+      "providers": {
+        "did:web:web3.storage": {
           "provider": "did:web:web3.storage",
           "space": "did:key:z6MkuxVKbEvYzXw89c9ESd3xoZ988MFrCgqT5JF5wtBvuYWe",
           "period": {
--- a/rfc/account-usage-get-capability.md
+++ b/rfc/account-usage-get-capability.md
@@ -1,7 +1,5 @@
 # Account Usage Get
 
-![draft](https://img.shields.io/badge/status-draft-yellow.svg?style=flat-square)
-
 ## Editors
 
 - Storacha team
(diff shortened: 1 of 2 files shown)


rfc/new-usage-capability.md
```diff
@@ -0,0 +1,210 @@
...
+
+  # Optional time period (Unix timestamps in seconds).
+  # If omitted, provider MUST return a current snapshot.
+  period optional Period
+}
+
+type Period struct {
+  from Int  # inclusive
+  to   Int  # inclusive
+}
```
alanshaw, line 51: You might want to use tuple representation here https://ipld.io/docs/schemas/features/representation-strategies/#struct-tuple-representation

rfc/new-usage-capability.md
```diff
@@ -0,0 +1,210 @@
...
+> - If we decide to keep it, we could gradually deprecate the current `usage/report`, since the same operation would be supported by the new one.
+> - On the other hand, if we choose to remove it, we can simplify the return type to only provide an overview of the total usage, while leaving the more detailed per-space reporting to the existing `usage/report`.
+> Let's discuss the options.
+
+
+> example: getting the total usage
+
+```json
+{
+  "iss": "did:mailto:web.mail:alice",
```
alanshaw: I don't think we ever do this do we? We just issue from the agent DID key and provide the attestation.

rfc/new-usage-capability.md
```diff
@@ -0,0 +1,210 @@
...
+  with AccountDID
+  nb optional UsageGetNB
+}
+
+type UsageGetNB struct {
+  space optional DID
+
+  # Optional time period (Unix timestamps in seconds).
+  # If omitted, provider MUST return a current snapshot.
+  period optional Period
```
alanshaw, line 45: I would make this required since it prevents the invocation from being idempotent - i.e. you make the same invocation and you will get a different answer if you don't provide the period.

rfc/new-usage-capability.md
```diff
@@ -0,0 +1,210 @@
...
+#### Invocation
+
+```ipldsch
+type UsageGet struct {
+  with AccountDID
+  nb optional UsageGetNB
+}
+
+type UsageGetNB struct {
+  space optional DID
```
alanshaw: Why not an array of spaces?

alanshaw: Can we define what happens when you provide a space DID that the account does not have access to? Does it error or simply return no results? I would guess the latter...

alanshaw (changes requested):
Love this - thanks for proposing!

rfc/new-usage-capability.md
BravoNatalie: That makes sense, thanks!

rfc/new-usage-capability.md
BravoNatalie: I understand the concern about idempotency, but I think leaving period optional is more user-friendly since it doesn’t require the user to know the billing period upfront. The receipt always includes the period that was applied. This way, anyone who needs idempotency can explicitly specify it, while keeping usage flexible for others. We can also inform users about the idempotency aspect and encourage them to provide a period when reproducibility is important.

rfc/new-usage-capability.md
BravoNatalie: I agree with omitting the inaccessible space from the result and treating the request as successful if at least one space is accessible. But, I’m concerned this might cause confusion when interpreting the result. Do you think returning a receipt without the space in the `spaces` map is sufficient, or should we add a dedicated field to account for errors? For example:

```json
{
  "total": 0,
  "spaces": {},
  "partial": true,
  "errors": {
    "did:key:zSpaceX": "not accessible",
    "did:key:zSpaceY": "not accessible"
  }
}
```

hannahhoward (commented):
Some thoughts:

- given that `usage/*` is generally scoped already to spacedid, perhaps this should be `account-usage/get`. honestly I think it would make sense for `usage/report` to be `space/usage/get` but we're a bit late for that I guess. I dunno.

- I am assuming the delegation you need is a mailto to the agent DID, with an attestation (`ucan/attest`) from Storacha, correct?

rfc/account-usage-get-capability.md
```diff
@@ -0,0 +1,218 @@
...
+
+type AccountUsageGetNB struct {
+  space optional [SpaceDID]
+
+  # Optional time period (Unix timestamps in seconds).
+  # If omitted, provider MUST return a current snapshot.
+  period optional Period
+}
+
+type Period struct {
```
hannahhoward, line 48: for consistency, I think we should make this representation Map (what is setup in the usage/report capability)

hannahhoward (approved):
see comment -- I am currently implementing in PR, so we should get @alanshaw 's sign off on the current version asap

hannahhoward:
I have now made an implementation of this capability in https://github.com/storacha/upload-service/pull/386

However, I did make a change in my implementation -- I think the usages should be divided by provider and then space. Did you have a reason to go the other way?

I also decided if you ask for a space you're not ultimately authorized for, the whole invocation fails.

rfc/account-usage-get-capability.md
```diff
@@ -0,0 +1,221 @@
...
+  from Int # inclusive
+  to Int # inclusive
+}
+```
+
+> example: getting the total usage
```
alanshaw: What is the "total usage"? What is the period in the returned data in this case?

rfc/account-usage-get-capability.md
```diff
@@ -0,0 +1,221 @@
...
+
+```json
+{
+  "iss": "did:key:z6MktfnQz8Kcz5nsC65oyXWFXhbbAZQavjg6LYuHOOOagent",
+  "aud": "did:web:storacha.network",
+  "att": [
+    {
+      "with": "did:mailto:web.mail:alice",
+      "can": "account-usage/get"
```
alanshaw: Can we not use a dash? Maybe we want to add capabilities for other account things in the future and allow delegating `account/*` to do that?

```suggestion
      "can": "account/usage/get"
```

Could even just be `account/usage`:

```suggestion
      "can": "account/usage"
```

rfc/account-usage-get-capability.md
```diff
@@ -0,0 +1,221 @@
...
+
+```json
+{
+  "total": 5356848797,
+  "spaces": {
+    "did:key:z6MkuxVKbEvYzXw89c9ESd3xoZ988MFrCgqT5JF5wtBvuYWe": {
+      "total": 5356848797,
+      "providers": {
+        "did:web:web3.storage": {
+          "provider": "did:web:web3.storage",
```
alanshaw, line 168: Maybe just a list instead of a provider map? Provider is an item for each space anyway.

Alternatively `{ [spaceDID]: Usage }`?

alanshaw (approved):
Please address feedback, but then LGTM.

rfc/account-usage-get-capability.md
```diff
@@ -0,0 +1,221 @@
...
+  out: Result<AccountUsageGetOk, AccountUsageGetError>
+}
+
+type AccountUsageGetError {
+    message: string
+}
+
+type AccountUsageGetOk {
+    total        Int
+    spaces       {String: SpaceUsage}   # key: SpaceDID
```
alanshaw: We should specify the keys are sorted so the same query produces the same output each time.

rfc/account-usage-get-capability.md
```diff
@@ -0,0 +1,221 @@
...
+}
+
+type SpaceUsage {
+  total     Int
+  providers {String: ProviderUsage}   # key: ProviderDID
```
alanshaw: We should specify the keys are sorted so the same query produces the same output each time.

rfc/account-usage-get-capability.md
BravoNatalie: updated the text with more context

PR #65, opened 2025-09-05
rfc: filepack

alanshaw:
Filepack is an archive format for transferring and storing content addressed data.

[📚 Preview](https://github.com/storacha/RFC/blob/ash/rfc/filepack/rfc/filepack.md)

[🎥 Demo](https://youtu.be/ApBkDlhyfuU?si=1w6YJYyMPYsll_jq)

### DIFF
--- a/rfc/filepack.md
+++ b/rfc/filepack.md
@@ -1,0 +1,55 @@
+# RFC: Filepack archive format
+
+**Author**: Alan Shaw (@alanshaw)
+
+**Date**: 2025-09-05
+
+**Status**: Draft
+
+## Abstract
+
+Filepack is an archive format for transferring content addressed data. It is comprised of two files: data archive and index. Both data and index MAY be split into multiple chunks.
+
+The format is an alternative to CAR, allowing smaller and faster uploads as well as quicker retrievals.
+
+
+## Format Definition
+
+### Data Archive
+
+A Filepack data archive is simply the concatenation of the bytes of 1 or more files and the bytes of 0 or more metadata items.
+
+```
+<file0-bytes>[...meta0-bytes][...fileN-bytes][...metaN-bytes]
+```
+
+For IPFS data, the metadata bytes may be the bytes of each non-leaf DAG node created in a UnixFS tree.
+
+Filepack data MUST be identified as a whole by a CID, allowing it to be referenced by the filepack index. The IPLD codec used in the CID MUST be “raw” since the data is simply concatenated raw file bytes.
+
+### Index
+
+A Filepack index adds content addressing to the data by specifying hashes, and byte offsets of files within the data archive. It allows the data to be made sense of either partially or as a whole.
+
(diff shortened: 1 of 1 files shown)


hannahhoward (approved):
Fundamentally, I love the filepack idea. to the extent this is a proposal for a thing we could do, I'm LGTM.

Also, I think we might go even farther and I'll make an RFC with some ideas.

What I'd like like to do now though is talk through the implications of implementation and if/when we'd actually do the work to put this in production.

Is it true to say we could ship Filepack in the upload-service and it would "just work" in freeway? I think so no? perhaps further optimization needed but it would work as is?

And then we'd need to modify guppy if we wanted things to be standard.

What I am ultimately wondering is: do we push on implementation  now, given all that we have to implement in the next few months? It is another thing to fit in. Or do we wait for a lull after warm storage ships and fit in UCAN 1.0 / tech debt

alanshaw:
> What I'd like like to do now though is talk through the implications of implementation and if/when we'd actually do the work to put this in production.

The only adverse implication I'm aware of is putting this data into filecoin, which is supposed to be content address data. i.e. it expects a CAR (well, the boost code expects the first few bytes of deal data to be a CAR). However we _already_ work around that by prepending an empty CAR piece to every aggregate.
 
> Is it true to say we could ship Filepack in the upload-service and it would "just work" in freeway? I think so no? perhaps further optimization needed but it would work as is?

Yes. In the demo (linked in the PR description) I showed retrieving data from freeway.

There would be a piece of work to optimise retrieval yes but it's compatible as is.

> And then we'd need to modify guppy if we wanted things to be standard.

IIRC the tracks are currently being laid in guppy. It might be preferable to implement while everyone's heads are "in the game" as it were.

> What I am ultimately wondering is: do we push on implementation now, given all that we have to implement in the next few months? It is another thing to fit in. Or do we wait for a lull after warm storage ships and fit in UCAN 1.0 / tech debt

Good question. What's nice about this is that we can split the work in two - writes and reads. We could do the writes work now and the reads work a lot later...and the reads work would still yield a speed up for the older data that we stored with the new format.

That said, it's all non-critical optimization work. I'd prefer we focused on more mission critical tasks _in general_, but the writes implementation would probably affect _only_ guppy at this point (since we have an [implementation in the JS client](https://github.com/storacha/upload-service/pull/409) already) so it's somewhat a question of whether @Peeja would be comfortable thrashing out the implementation in guppy while she's already knee deep in it...

The other thing to consider is that guppy is targeted to upload a substantial amount of data to the network, and IMO it would be preferable to enable an optimised retreival for that data, even if it is not realized immediately. We cannot "go back" and enable this optimization for all that data.

BravoNatalie:
I know it’s already in the demo video, but could you also share the tldr you presented here?

alanshaw:
Of course! https://www.tldraw.com/f/M2SIzY7S-scAnDgkuGs0e?d=v-50.-266.1366.775.G6_i5aitwPjqP9tzED4dj

PR #69, opened 2025-10-08
docs: signing service RFC

frrist:
📖 [Preview](https://github.com/storacha/RFC/blob/03f2f8966254bfddf15c6325ea66e6bf4853b463/rfc/signing-service.md)

🧌 [Completely unhinged vibe-coded alternative that we absolutely don't have the time to implement before launch, but included anyways just for fun](https://gist.github.com/frrist/72063c58fd73d1751f63c4a795df4826)

### DIFF
--- a/rfc/signing-service.md
+++ b/rfc/signing-service.md
@@ -1,0 +1,276 @@
+# Authorizing Proof Data Provider Operations
+
+## Problem
+
+Node providers on the Storacha network need to perform operations on proof sets - adding piece roots, deleting roots, and other state changes. 
+These operations require signatures from Storacha to be valid on-chain.
+
+Currently, we don't have a clean way to authorize these operations. Node providers need to:
+1. Prove they're authorized to request signatures from Storacha
+2. Get Storacha's signature for their operation payload
+3. Use that signature to submit blockchain transactions
+
+We need a system that:
+- Only allows authorized operators to request signatures
+- Prevents unauthorized access to the signing service
+- Maintains a verifiable chain of authority back to Storacha
+- Doesn't require manual management of operator allow-lists
+- Uses short-lived delegations instead of relying on revocations
+
+## Background
+
+### EIP-712: Typed Structured Data Signing
+
+The operations we're authorizing need to be submitted as blockchain transactions. 
+Specifically, we're calling smart contract methods like `addPieces()` and `deletePieces()` on Filecoin.
+
+These contract methods require signatures that prove the operations were authorized by Storacha. 
+We use [EIP-712](https://eips.ethereum.org/EIPS/eip-712) for this signing.
+
+EIP-712 is an Ethereum standard for signing typed, structured data. 
+Instead of signing raw bytes (which are hard for users to verify), EIP-712 lets you sign structured data with clear types and fields. 
+The signature proves:
+1. The exact data being signed (operation type, parameters, values)
+2. Which contract the data is intended for (domain separation)
+3. Who signed it (via signature verification on-chain)
+
+For example, an `AddPieces` operation might have a structure like:
+
+```solidity
+struct AddPieces {
+  uint256 clientDataSetId;
+  uint256 firstAdded;
+  bytes32[] pieceData;
+  bytes[] metadata;
+}
+```
+
+The EIP-712 signature of this structure can be verified on-chain by the smart contract, which checks that it was signed by Storacha's authorized key.
+
+This is why we need a signing service. 
+The private key that creates these EIP-712 signatures must be controlled by Storacha, but node providers need to be able to request signatures for their operations. 
+The signing service is the bridge - it holds Storacha's key and issues signatures to authorized operators.
+
+## Proposal
+
+### UCAN-Authorized Signing Service
+
+We'll create a signing service that uses UCAN delegations to authorize signature requests. 
+The flow works like this:
+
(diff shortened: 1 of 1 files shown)


travis (commented):
this makes sense to me!

travis (approved):
this makes sense to me!

alanshaw (approved):

rfc/signing-service.md
```diff
@@ -0,0 +1,276 @@
...
+│ Issuer:    did:web:storacha.network     │ ← Storacha's identity
+│ Audience:  did:key:PiriNode...          │ ← Operator's identity
+│ Capability:                             │
+│   with: "did:key:SigningService"        │
+│   can:  "pdp/sign"                      │
+│   nb:   {                               │
+│     operations: [                       │
+│       "AddPieces",                      │
+│       "DeletePieces",                   │
+│       ...                               │
```
volmedo, line 118: this would be easier to implement with a UCAN 1.0 policy.

How would this be implemented in UCAN 0.9? Would the capability caveats contain all the fields but use a different set of them in delegations and invocations (e.g. the delegation would state allowed operations in an `operations` field, but invocations would contain `operation` and `payload`, where `operation` needs to be contained in the delegation's `operations` list).

An alternative would be to define different capabilities for the different possible operations and delegate them independently.

rfc/signing-service.md
```diff
@@ -0,0 +1,276 @@
+# Authorizing Proof Data Provider Operations
+
+## Problem
+
+Node providers on the Storacha network need to perform operations on proof sets - adding piece roots, deleting roots, and other state changes.
```
volmedo, line 5: maybe we should start calling them data sets 🙂

volmedo (approved):
I very much like this!

I have some questions about the distribution of `pdp/sign` delegations to node providers. The RFC mentions there is no need for allow-lists, but I don't think that's true. Maybe it is not an actual list in the signing service, but the list needs to exist somewhere in some form. If it's automated when operators join, something or someone needs to allow them to join, same if they are issued through an admin interface. A similar statement applies to `access/authorize`. I think this is currently a property of the system.

It is also mentioned that Storacha delegations can be short-lived, which I think would be a desirable property, but wouldn't that rule out getting the delegation once when joining the network or the admin interface as mechanisms for delegation distribution?

BravoNatalie:
LGTM! Regarding the Storacha key, do we plan to support rotation? What happens if we need to change the key that we use to sign?

BravoNatalie (approved):

frrist:
@BravoNatalie Yes! The [Service contract](https://github.com/storacha/filecoin-services/tree/main/service_contracts/lib) supports a [Session Key Registry](https://github.com/FilOzone/SessionKeyRegistry/tree/e472ca2b525fb2396832216182b64a0c165cb49c) that can issue keys with scoped capabilities. The signing service could use a service key, issued by storacha that is expired and rotated. Though I have left this outside the scope of the RFC.

frrist:
@volmedo Yes, good points. An allow-list will probably need to exist somewhere since medium-to-short-lived delegations require on-demand distribution via `access/authorize`. 
(Note that initially said list could just be the [staging-warm-delegator-storage-provider-allow-list](https://us-east-2.console.aws.amazon.com/dynamodbv2/home?region=us-east-2#item-explorer?maximize=true&table=staging-warm-delegator-storage-provider-allow-list) we already maintain for issuing delegations from the delegator)

In the context of this RFC I see these as implementation details, and have left out of scope here. Where the allow-list lives and how it's checked will be determined whenever we build an "authorization service". My aim in this RFC is establish the authorization pattern. The mechanics can be left to implementation.

rfc/signing-service.md
frrist: > the delegation would state allowed operations in an operations field, but invocations would contain operation and payload, where operation needs to be contained in the delegation's operations list

Yes, exactly! This validation is covered in Step 3.4 under "Validate caveats":
>  - Is the operation valid (e.g., "AddPieces", "DeletePieces")?
>  - Does the payload have all required fields?

The signing service checks that nb.operation from the invocation is in nb.operations from the delegation. We could considered separate capabilities per operation but a single capability feels more flexible for the time being.

volmedo: yes, I get that. I just wanted to point out that we usually define the same set of fields for caveats in both delegations and invocations, without making a distinction when describing capabilities.

Not saying this approach is wrong, it is just not what we usually do AFAICT.

volmedo:
> An allow-list will probably need to exist somewhere since medium-to-short-lived delegations require on-demand distribution via access/authorize.
(Note that initially said list could just be the [staging-warm-delegator-storage-provider-allow-list](https://us-east-2.console.aws.amazon.com/dynamodbv2/home?region=us-east-2#item-explorer?maximize=true&table=staging-warm-delegator-storage-provider-allow-list) we already maintain for issuing delegations from the delegator)
>
> In the context of this RFC I see these as implementation details, and have left out of scope here. Where the allow-list lives and how it's checked will be determined whenever we build an "authorization service". My aim in this RFC is establish the authorization pattern. The mechanics can be left to implementation.

I agree it's an implementation detail. I just wanted to highlight that, strictly speaking, the proposed mechanism doesn't really avoid the need of an allow-list, whatever the implementation of such a list might be.

PR #72, opened 2025-10-28
docs: UCAN Command Definitions RFC

Peeja:
[📖 Preview](https://github.com/storacha/RFC/blob/command-definitions/rfc/command-definitions.md)

> This RFC proposes a definition and formal documentation format for UCAN Commands, to replace the manually created and inconsistently structured specification documents we currently use.
>
> This is largely a proposal for UCAN itself, but as it's motivated by needs arising in Storacha, it will at least begin as a draft RFC here for initial feedback. It is fully compatible with the existing UCAN 1.0 specifications, and could be adopted by UCAN itself or live on top of it. However, its utility is limited if it is not adopted outside of a single organization.

### DIFF
--- a/rfc/command-definitions.md
+++ b/rfc/command-definitions.md
@@ -1,0 +1,190 @@
+# RFC: UCAN Command Definitions
+Status: Proposed Standard
+
+## Introduction
+
+This RFC proposes a definition and formal documentation format for UCAN Commands, to replace the manually created and inconsistently structured specification documents we currently use.
+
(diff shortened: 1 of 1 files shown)


rfc/command-definitions.md
```diff
@@ -0,0 +1,190 @@
...
+  "aud": "did:example:alices-front-door",
+  "sub": "did:example:alice",
+  "cmd": "/home/access",
+  "nonce": "…",
+  "exp": null,
+  "prf": [{"/": "bafy..accessdel"}]
+}
+```
+
+`did:example:alices-front-door` is the smart lock on Alice's front door. Seeing that Bob has authority from Alice to "access" her "home", it accordingly unlocks and lets Bob in, where he has a very different kind of access than what Alice intended to delegate.
```
alanshaw, line 48: Shouldn't the _subject_ differentiate here? You're not accessing "alice", you're accessing the alice's house. The house should have a DID, as should the server space - this is the subject of the invocation.

PR #75, opened 2025-11-12
rfc: gateway authorization flow - migration

fforbeck:
[📖 Preview](https://github.com/storacha/RFC/blob/rfc/gateway-auth-migration/rfc/gateway-authorization-migration.md)

This RFC proposes a simplified authorization mechanism for migrating legacy spaces to the new Storacha gateway infrastructure. Legacy spaces lack private keys, and we cannot create valid UCAN delegations, requiring a special authorization flow using `ucan/attest`, but it fails. The proposal is that the gateway creates and signs delegations on behalf of legacy spaces, then trusts its own signatures without complex attestation validations.

### DIFF
--- a/rfc/gateway-authorization-migration.md
+++ b/rfc/gateway-authorization-migration.md
@@ -1,0 +1,223 @@
+# RFC: Gateway Authorization for Legacy Spaces Migration
+Status: Experimental
+
+## Authors
+
+- [Felipe Forbeck](https://github.com/fforbeck), [Storacha Network](https://storacha.network/)
+
+## Abstract
+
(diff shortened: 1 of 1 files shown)


hannahhoward (commented):
was never a huge fan of the original ucan/attest plan.

I'm a little fuzzy on why it's not working ... would want @alanshaw 's input before we go forward

question: can't this be done from the delegation we have for the account for most spaces on the upload service?

IOW, 
- we have a store of space/* to email, valid signature from the space key
- now we make a account email to gateway pub key, w/ space/content/serve, attested in the facts by the upload service
- this is exactly the flow we use for delegating between devices so I don't know why we'd not be able to do it.

essentially it's running the email validation flow but w/o the actual confirmation (don't love, but c'est l'vie)

hannahhoward:
FWIW I think that the ucan/attest in the email flow is in the facts, not the direct chain

PR #78, opened 2026-01-13
rfc: space-diff table refactoring

BravoNatalie:
This RFC proposes a refactor of the `space-diff` table to eliminate space usage calculation timeouts, prevent duplicate diffs, and improve the efficiency of billing usage calculations.

[📖 Preview](https://github.com/storacha/RFC/blob/786450f38b0450e2d941d114401a7dd5d7313742/rfc/refactor-space-diff-table.md)

### DIFF
--- a/rfc/refactor-space-diff-table.md
+++ b/rfc/refactor-space-diff-table.md
@@ -120,44 +120,18 @@ The additional cost is acceptable, especially since older diffs can be safely de
 
 ### Fix for problem 2: Usage calculation timeouts
 
-Introduce a new table (e.g. `space-usage-month`) keyed by `provider#space#YYYY-MM` that is updated atomically on each diff write, making billing reads **O(1)**.
+Generate snapshots more frequently.
 
-#### Core idea
+One option is to move snapshot generation to a daily cadence. There are two possible approaches:
 
-Maintain a running usage accumulator instead of scanning historical diffs.
+1. **Decouple snapshot generation from the billing cron**
 
-**Algorithm**
+  * Generate snapshots independently, without running the full billing pipeline
 
-1. Track `lastSize` and `lastChangeAt` per `(provider, space, month)`
-2. On each incoming diff:
-   * `usage += lastSize × (receiptAt - lastChangeAt)`
-   * `lastSize += delta`
-   * `lastChangeAt = receiptAt`
-3. At end-of-month billing:
-   * `usage += lastSize × (periodEnd - lastChangeAt)`
-   * Finalize and snapshot
+2. **Run the full billing process daily**
 
-**Additional fields**
+  * This would naturally produce more snapshots and also push usage reports to Stripe more frequently
 
-* `sizeStart`
-* `sizeEnd`
-* `lastReceiptAt`
-* `subscription` 
+Since both approaches are pretty similar and would need to iterate over all customers and spaces to generate snapshots anyway, the main extra work with running the full billing flow is the usage calculation, writing usage records, and reporting to Stripe.
 
-**Behavior**
-
-* `space-diff` remains for audit and idempotency
-* Billing reads exclusively from `space-usage-month`
-* `calculatePeriodUsage`:
-  * First tries the aggregator
-  * Falls back to a GSI scan if missing
-* Aggregator becomes the canonical source for the billing month
-
-**Retention**
-
-* Keep `space-diff` entries for N months using TTL
-* Archive older diffs to S3 (TBD)
-
-**Considerations**
-
-To avoid potential race conditions when two diffs for the same space read the current total at the same time, one option is to process diffs through a queue. This would also help preserve the correct ordering of diffs.
+Given the upside of reporting to Stripe more frequently and the fact that this is simpler than setting up separate infra just for snapshot generation, we’ll move forward with option two.
--- a/rfc/refactor-space-diff-table.md
+++ b/rfc/refactor-space-diff-table.md
@@ -115,11 +115,8 @@ The additional cost is acceptable, especially since older diffs can be safely de
    * Correct PK design
    * `cause` as SK
    * GSI for timestamp-based queries
-2. Export data from the existing table
-3. Deduplicate and transform records
-4. Import data into the new table
-5. Update application code to use the new schema
-6. Decommission the old table after validation is complete
+2. Enable dual-writes: on each diff event, write to both the existing table and the new table. Keep all readers (usage, reporting, billing) pointed at the existing table during January.
+3. Cut over in February: switch usage reporting and billing reads to the new table; keep the existing table as read-only historical storage.
 
 ### Fix for problem 2: Usage calculation timeouts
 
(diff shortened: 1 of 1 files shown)


rfc/refactor-space-diff-table.md
```diff
@@ -0,0 +1,172 @@
...
+* Use **`cause` as the sort key (SK)** of the `space-diff` table
+* This makes it impossible to insert two diffs for the same `(space, cause)` pair
+
+#### Open design concern
+
+Using `cause` as the SK removes natural chronological ordering.
+
+**Proposed solution**
+
+* Add a **GSI with a timestamp-based sort key**
```
travis, line 87: 🎉

rfc/refactor-space-diff-table.md
```diff
@@ -0,0 +1,172 @@
...
+
+The additional cost is acceptable, especially since older diffs can be safely deleted after the retention window.
+
+#### Migration plan (high level)
+
+1. Create a **new `space-diff` table** with:
+   * Correct PK design
+   * `cause` as SK
+   * GSI for timestamp-based queries
+2. Export data from the existing table
```
travis: I think we might even be able to just skip importing the old data - we can spin this system up alongside it, use it in parallel until february, and then cut over to the new table for usage reporting and billing and leave the old diffs in the old table

rfc/refactor-space-diff-table.md
```diff
@@ -0,0 +1,172 @@
...
+* Aggregator becomes the canonical source for the billing month
+
+**Retention**
+
+* Keep `space-diff` entries for N months using TTL
+* Archive older diffs to S3 (TBD)
+
+**Considerations**
+
+- The accumulator MUST process diffs for a space in **ascending** `receiptAt` order. If the write path can deliver out-of-order events and strict ordering cannot be guaranteed, this solution SHOULD be revisited. Pragmatic mitigations include:
```
travis: is this strictly necessary? I think we could just choose the later of "current accumulator date" or "new diff receiptAt" - this will ensure we always have the "latest" date for a particular accumulator, even in the case where we process diffs out of order

rfc/refactor-space-diff-table.md
```diff
@@ -0,0 +1,172 @@
...
+  * First tries the aggregator
+  * Falls back to a GSI scan if missing
+* Aggregator becomes the canonical source for the billing month
+
+**Retention**
+
+* Keep `space-diff` entries for N months using TTL
+* Archive older diffs to S3 (TBD)
+
+**Considerations**
```
travis: my only major concern here is the potential race condition if two diffs for the same space read the current total at the same time - we could mitigate this by having a single queue reader processing the diffs - it should go pretty fast and uploads aren't THAT high frequency so having a single queue reader process diffs and update the accumulator should be plenty - this would solve race conditions pretty conclusively I think

travis (approved):
looks great! I have a couple suggestions but generally very excited about this change!!

rfc/refactor-space-diff-table.md
BravoNatalie: I reviewed this again later and don’t think it’s a problem, especially since receiptAt is generated server-side when we add the diff entry to the table.

rfc/refactor-space-diff-table.md
```diff
@@ -163,10 +160,4 @@ Maintain a running usage accumulator instead of scanning historical diffs.
...
 **Considerations**
 
-- The accumulator MUST process diffs for a space in **ascending** `receiptAt` order. If the write path can deliver out-of-order events and strict ordering cannot be guaranteed, this solution SHOULD be revisited. Pragmatic mitigations include:
-  - Buffer within a small window and sort incoming diffs.
-  - Recompute a localized suffix by reading recent diffs via the time GSI and re-applying from the last stable checkpoint.
-
-- Alternative when strict ordering is infeasible:
-  - Use time-bucketed diffs (hour/day): persist per-bucket, order-independent aggregates (e.g., Σdelta and Σ(delta × (bucketEnd − receiptAt))). At billing time, iterate buckets in chronological order to compute exact monthly usage, where no event sorting required.
-  - Maintain a size-only monthly state (track `lastSize` and `lastChangeAt`) to accelerate space usage report. Note: this does NOT remove the need to iterate diffs for the billing run.
+To avoid potential race conditions when two diffs for the same space read the current total at the same time, one option is to process diffs through a queue. This would also help preserve the correct ordering of diffs.
```
hannahhoward: @alanshaw can confirm and you should chat briefly with him about this:

my understanding is:
1. the existing diffs are written in batches of 25 from a kinesis stream with enhanced fan-out. Enhanced fan-out means up to 20 invocations of 25 can be called in parallel to the lambda that writes space diffs
2. While you could apply a FIFO SQS Queue for a post process step to get maintain this running total, note that part of the heavy batching and parallelism for the kinesis stream, at least to my understanding, is make sure the lambda doesn't fall behind.

I think it may make sense to pursue a kind of daily auto-compaction instead, so that you process in aggregate once a day. Lambda invocations are non-zero cost per invocation.

Second, please make sure the usage/report & account-usage/report maintain the ability to pass a time range -- this is neccesary for Forge

hannahhoward (commented):
Left one comment with some thoughts and things to consider

rfc/refactor-space-diff-table.md
BravoNatalie: My understanding is that the scenario you’re describing only applies to `store/add` and `store/remove` receipts. Blob protocol receipts are written through a different path, via the Blob Registry `register` and `deregister` flows.
Since we’re planning to deprecate the Store Protocol, this shouldn’t be an issue. Can you confirm, @alanshaw ?

rfc/refactor-space-diff-table.md
```diff
@@ -0,0 +1,163 @@
...
+1. Create a **new `space-diff` table** with:
+   * Correct PK design
+   * `cause` as SK
+   * GSI for timestamp-based queries
+2. Enable dual-writes: on each diff event, write to both the existing table and the new table. Keep all readers (usage, reporting, billing) pointed at the existing table during January.
+3. Cut over in February: switch usage reporting and billing reads to the new table; keep the existing table as read-only historical storage.
+
+### Fix for problem 2: Usage calculation timeouts
+
+Introduce a new table (e.g. `space-usage-month`) keyed by `provider#space#YYYY-MM` that is updated atomically on each diff write, making billing reads **O(1)**.
```
alanshaw: This is basically the space metrics table - could you use that instead?

rfc/refactor-space-diff-table.md
alanshaw: Yes in the blob protocol the blob registry writes to the space diff table - the ucan stream is not used for this any more. My original idea was to remove it post store protocol removal.

The space snapshot table consolidates space diffs, could you use the existing code/infra (with small modification) to calculate a snapshot on a daily/weekly basis for spaces so that when you want to know current usage you don't have to look so far back?

alanshaw (changes requested):

rfc/refactor-space-diff-table.md
BravoNatalie: So yes, we could combine the space-metrics and snapshots with more frequent runs. However, we would still need to query the full list of customers and spaces every time to iterate though the diffs. This would reduce the time spent calculating usage and billing, but the proposed new table would consolidate currently dispersed data into a single place (snapshots, usage, and size) while also providing near real-time visibility into each space. This feels like a better approach for scalability to me, but I'd like to know if you have a different thought since you have a better view of the system. @travis might also have good input here.

@alanshaw I have a question about `space-metrics`: is it still actively used? I only see writes to it and no reads,  does it still make sense to keep it around?

rfc/refactor-space-diff-table.md
alanshaw: > my only major concern here is the potential race condition if two diffs for the same space read the current total at the same time

For incrementing? You'd use an update command with `ADD` which would consistently increment.

rfc/refactor-space-diff-table.md
alanshaw: Can you speak to "updated atomically on each diff write" - how is this applied? Is it lambda triggered on an insert event to the space diff table?

alanshaw: How do we transition from one month to the next, retaining the current total, so that concurrent updates succeed and are all applied as expected?

rfc/refactor-space-diff-table.md
alanshaw: I am unsure how this works for the transition from one month to the next though...

rfc/refactor-space-diff-table.md
alanshaw: > but the proposed new table would consolidate currently dispersed data into a single place (snapshots, usage, and size) 

Yes, but at the expense of another dynamo table (storage, writes) and lambda invocation per space diff insert. It's just more infra to run, maintain and (potentially) go wrong, whereas upping the cron frequency for calculating snapshots basically gets you everything you need...unless I'm missing something?

> while also providing near real-time visibility into each space

We already have this though...right? Latest snapshot + diffs since is the real time.

travis: > Latest snapshot + diffs since is the real time.

yea I think the issue here is that not all spaces can calculate this in a reasonable amount of time - if there are 200k diffs to add to a space snapshot it generally just doesn't work

PR #79, opened 2026-01-30
rfc: UCAN Authroized Retrieval via Roundabout

alanshaw:
A proposal for how to do UCAN authorized retrieval in the Forge network with Spade.

[📖 Preview](https://github.com/storacha/RFC/blob/ash/rfc/roundabout-ucan-retrieval/rfc/roundabout-ucan-retrieval.md)

Note: we could push this back further _into_ the filecoin pipeline - i.e. the manifest _it_ provides to Spade could include these headers...interested in opinions here.

### DIFF
--- a/rfc/roundabout-ucan-retrieval.md
+++ b/rfc/roundabout-ucan-retrieval.md
@@ -12,7 +12,7 @@ Roundabout currently just redirects to the found URL. There is no provision for 
 
 The Spade API provides agents (clients) a [frc58 "manifest"](https://github.com/filecoin-project/FIPs/blob/master/FRCs/frc-0058.md) from which they can build an aggregate piece from.
 
-It contains the aggregate piece CID, a list of _segment_ piece CIDs, and a URL (multiple supported, in practice only 1). It has the structure:
+It contains the aggregate piece CID, a list of _segment_ piece CIDs, and for each segment, a URL (multiple supported, in practice only 1) the data can be retrieved from. It has the structure:
 
 ```go
 type ResponsePieceManifestFR58 struct {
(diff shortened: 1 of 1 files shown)


Peeja (commented):
I think I'm missing some context on this. Do we control both ends of the Spade conversation? I had imagined that Spade implemented a common API, but are the (Filecoin) storage providers supposed to run a _Spade_ client specifically to fetch from our Spade service? Is that why we can extend it this way?

In that case, we could probably make the API know more about UCAN directly. Is the idea here not to do that to keep UCAN out of Spade? But HTTP headers are fair game because it's already dealing with HTTP URLs?

alanshaw:
> I think I'm missing some context on this. Do we control both ends of the Spade conversation? I had imagined that Spade implemented a common API, but are the (Filecoin) storage providers supposed to run a _Spade_ client specifically to fetch from our Spade service? Is that why we can extend it this way?

Yes we run the Spade API and the clients run a Spade client. The status quo is that SPs run various scripts and community created clients in order to interact with Spade. We will create our own Spade client (I'm calling this an agent, since it's a long running process, not just a client library) so that we can ensure quality of service.

> In that case, we could probably make the API know more about UCAN directly. Is the idea here not to do that to keep UCAN out of Spade? But HTTP headers are fair game because it's already dealing with HTTP URLs?

The Spade API already authenticates clients with it's own auth system (SPID), so in the interests of not rewriting the whole thing the idea where was to just extend the response format slightly so that it can basically operate how it currently does, with minimal changes. It's also backwards compatible with existing publically available data - just don't send any HTTP headers. It should also be pretty easy for any existing clients to be updated, should anyone ever want to do that.

PR #80, opened 2026-02-03
rfc: Proposal: Single-DID Service Entry

Peeja:
[📖 Preview](https://github.com/storacha/RFC/blob/rfc/single-did-service-entry/rfc/single-did-service-entry.md)

This is something that's been rolling around in my head for a while.

> We cart around a whole set of environment variables to point to the correct environment. Rather than set all of them separately, and potentially incorrectly, we can set a single value as an entry point and discover the rest of the service from there. DID documents offer us a great way to do exactly that. Thus, I am proposing that we list the required endpoints and services in the root service's DID document, and discover them from there.

### DIFF
--- a/rfc/single-did-service-entry.md
+++ b/rfc/single-did-service-entry.md
@@ -1,0 +1,91 @@
+# Proposal: Single-DID Service Entry
+
+## Authors
+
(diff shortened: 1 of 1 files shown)


alanshaw (approved):
❤️ Love all of this.

PR #83, opened 2026-03-02
doc: go container release process

frrist:
[Preview](https://github.com/storacha/RFC/blob/dcbd9a30543d62cdffd8082ec49cbcf9f4bd4e76/rfc/container-release-process.md)

### DIFF
--- a/rfc/container-release-process.md
+++ b/rfc/container-release-process.md
@@ -7,15 +7,14 @@ This RFC defines the standard process for building and publishing container imag
 ## Status
 
 - **Status**: Review
-- **Applies to**: piri, piri-signing-service, delegator, indexing-service, and more
+- **Applies to**: All deployable golang Storacha services
 
 ## Goals
 
 1. Multi-architecture support — containers should work on both `linux/amd64` and `linux/arm64`
 2. Consistent tagging across all services
-3. Clear separation of dev and prod images
-4. Immutable tags for reproducibility
-5. Mutable tags for convenience
+3. Immutable tags for reproducibility
+4. Mutable tags for convenience
 
 ## Container Registry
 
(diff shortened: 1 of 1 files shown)


rfc/container-release-process.md
```diff
@@ -0,0 +1,165 @@
+# Container Image Release Process
+
+## Summary
+
+This RFC defines the standard process for building and publishing container images across Storacha Go services.
+
+## Status
+
+- **Status**: Review
+- **Applies to**: piri, piri-signing-service, delegator, indexing-service, and more
```
alanshaw: I would use them as examples, and generalise to "Storacha services with a public API" or similar.

rfc/container-release-process.md
```diff
@@ -0,0 +1,165 @@
...
+| Tag | Target | Mutable | Description |
+|-----|--------|---------|-------------|
+| `main` | prod | Yes | Latest main branch (production build) |
+| `main-dev` | dev | Yes | Latest main branch (dev build with debug tools) |
+| `<sha>` | prod | No | Specific commit (production build) |
+| `<sha>-dev` | dev | No | Specific commit (dev build) |
+
+### Release (tag v*)
+
+A release tag should produce **prod** images only. Nobody ships a debugger to production on purpose.
```
alanshaw: I'd maybe include that container tag should match the git tag and obviously build from the same commit.

rfc/container-release-process.md
```diff
@@ -0,0 +1,165 @@
...
+
+| Tag | Target | Mutable | Description |
+|-----|--------|---------|-------------|
+| `<version>` | prod | No | Semantic version (e.g., `1.2.3`) |
+
+There is no `latest` tag. The `latest` tag is a convention that rewards ambiguity — it lets you deploy without knowing what you're deploying, which is the kind of convenience that generates incident reports. All references to images should specify an explicit version.
+
+### Pull Requests
+
+PR builds should validate that the Dockerfile compiles but need not push images anywhere. Building on a single platform (amd64) is sufficient here — the goal is fast feedback, not architectural completeness.
```
alanshaw: Is there any chance that building for a different arch could fail? My guess is no, or at least _very_ unlikely, but if we're concerned then we should probably build on both.

rfc/container-release-process.md
```diff
@@ -0,0 +1,165 @@
...
+ENTRYPOINT ["/usr/bin/app"]
+```
+
+The key insight: `$BUILDPLATFORM` is the CI runner's architecture (x86_64), and `$TARGETPLATFORM` is the final image's architecture. Go handles cross-compilation natively, so the compiler runs fast while producing binaries for whatever architecture you need.
+
+QEMU remains necessary for the runtime stages, where `apt-get` and friends must execute on the target architecture. These operations are lightweight compared to compilation, so the tradeoff is favorable.
+
+## CI/CD Considerations
+
+The workflow should live at `.github/workflows/publish-ghcr.yml` and be named `Container`.
```
alanshaw, line 196: Include a template?

alanshaw (approved):
LGTM - left some suggestions.

rfc/container-release-process.md
```diff
@@ -0,0 +1,165 @@
...
+## Dockerfile Guidance
+
+Each service should provide a multi-stage Dockerfile with at least two targets.
+
+### `prod` target
+
+The production target should be small, fast, and boring:
+
+- Stripped binary (`-ldflags="-s -w"`)
+- Minimal base image (`debian:bookworm-slim`)
```
volmedo: For Go I'd go with `scratch` for a really minimal image with the smallest attack surface possible. The caveat is that `scratch` has nothing in it, so it requires copying CA certificates from the builder image (we currently do this in our services deployed via storoku), and setting a non-root image up is a bit more involved.

Alternatively, `gcr.io/distroless/static` could be a good middle ground, as it is small but comes with CA certificates and a non-root user pre-baked in the image.

Of course, both alternatives work only for statically-linked binaries. But that is what we are doing currently.

rfc/container-release-process.md
```diff
@@ -0,0 +1,165 @@
...
+- Only what's needed to run and health-check: `ca-certificates`, `curl`
```
volmedo: To run a healthcheck on a different service? Having `curl` installed in a prod image it's an uncommon requirement IMO.

rfc/container-release-process.md
```diff
@@ -0,0 +1,165 @@
...
+
+The dev target exists for the days when something has gone wrong and you need to figure out what. It should include:
+
+- A debug-friendly binary (`-gcflags="all=-N -l"`)
+- Delve debugger
+- Enough diagnostic tools to be useful: `bash-completion`, `less`, `vim-tiny`, `procps`, `htop`, `strace`, `iputils-ping`, `dnsutils`, `net-tools`, `tcpdump`, `jq`
+
+The precise selection of debugging tools is a matter of taste. The principle is not: include everything that may be required.
+
+### Non-root user
```
volmedo, line 102: ❤️

rfc/container-release-process.md
```diff
@@ -0,0 +1,165 @@
...
+
+FROM debian:bookworm-slim AS prod
+RUN apt-get update && apt-get install -y ca-certificates curl && rm -rf /var/lib/apt/lists/*
+COPY --from=build /app /usr/bin/app
+ENTRYPOINT ["/usr/bin/app"]
+```
+
+The key insight: `$BUILDPLATFORM` is the CI runner's architecture (x86_64), and `$TARGETPLATFORM` is the final image's architecture. Go handles cross-compilation natively, so the compiler runs fast while producing binaries for whatever architecture you need.
+
+QEMU remains necessary for the runtime stages, where `apt-get` and friends must execute on the target architecture. These operations are lightweight compared to compilation, so the tradeoff is favorable.
```
volmedo: Avoiding the `RUN` command would also avoid the need of emulation completely. CA certificates can be installed in the builder and copied over to the runtime image (or use an image that comes with them pre-installed, such as `distroless/static`). Getting `curl` would be trickier though, but I'm not sure that should be a requirement.

rfc/container-release-process.md
```diff
@@ -0,0 +1,165 @@
...
+### Push to main branch
+
+A push to main should produce **both** dev and prod images:
+
+| Tag | Target | Mutable | Description |
+|-----|--------|---------|-------------|
+| `main` | prod | Yes | Latest main branch (production build) |
+| `main-dev` | dev | Yes | Latest main branch (dev build with debug tools) |
+| `<sha>` | prod | No | Specific commit (production build) |
+| `<sha>-dev` | dev | No | Specific commit (dev build) |
```
volmedo: These are short SHAs, correct? Might be worth making it explicit.

volmedo (approved):
Looks good overall. I left some suggestions.

rfc/container-release-process.md
frrist: Good idea, done.

rfc/container-release-process.md
frrist: The RFC now explicitly states this (lines 44-45):

Note that container tags omit the v prefix common in git tags. A git tag of v1.2.3 produces a container tag of 1.2.3. This matches the convention used by major container images (nginx, postgres, node, etc.).

rfc/container-release-process.md
frrist: Incorporated with a twist. The RFC now standardizes on `alpine:latest` as the single base image — it's ~5MB, includes wget and CA certificates, and provides a shell for Docker Compose healthchecks as well as entry point scripts which Piri and IPNI require.

rfc/container-release-process.md
frrist: Standardizing on `wget` instead which we get for free via https://github.com/storacha/RFC/pull/83#discussion_r2914315811

rfc/container-release-process.md
frrist: switching to alpine:latest should address this. Sorry was over complicating things here.

rfc/container-release-process.md
frrist: yea, now explicitly stated

PR #84, opened 2026-03-04
rfc: mutability & encryption for forge

### LINKED ISSUE #663: Plan privacy/mutability story
Gather requirements and beginnning planning endeavor for privacy and mutability.

fforbeck:
- [📖 Forge Encryption RFC](https://github.com/storacha/RFC/blob/rfc/privacy-mutability-forge/rfc/forge-encryption.md)
- [📖 Forge Mutability RFC](https://github.com/storacha/RFC/blob/rfc/privacy-mutability-forge/rfc/forge-mutability.md) 

### What We're Building

**Two complementary features for Forge enterprise customers:**

1. **Mutability** - Track and update content with stable paths
2. **Encryption** - Protect content with client-side encryption and access control

### Mutability ([📖 Forge Mutability RFC](https://github.com/storacha/RFC/blob/rfc/privacy-mutability-forge/rfc/forge-mutability.md))

**Problem:** Content on IPFS is immutable (CIDs change when content changes). Enterprise backup workflows need stable references that update to the latest version.

**Solution:** Extend existing Guppy commands (`upload source add`, `upload`, `retrieve`) with Pail + UCN integration.

**How it works:**
- **Pail** - A sharded data structure that maps paths to CIDs
- **UCN** - Publishes updates so other clients can resolve the latest version

**Commands:**
```bash
guppy upload source add <space> <path> --name backups   # Register source, init Pail bucket
guppy upload <space> backups                             # Upload + store path→CID in Pail
guppy retrieve <space> backups ./output                  # Retrieve by path (resolves via Pail)
```

### Encryption ([📖 Forge Encryption RFC](https://github.com/storacha/RFC/blob/rfc/privacy-mutability-forge/rfc/forge-encryption.md))

**Problem:** Enterprise customers need data protection before content leaves their systems.

**Solution:** Client-side encryption with UCAN-gated access control.

**How it works:**
- **Per-file encryption** - Each file gets its own key (DEK)
- **KMS integration** - Keys are wrapped and managed by Storacha KMS (KEK)
- **UCAN delegations** - Fine-grained access control (per-file or per-folder)

**Encryption modes**
| Mode | Use Case | Access Control |
|------|----------|----------------|
| KMS (production) | Enterprise | UCAN-gated |
| Local key (dev) | Testing | Anyone with key |

### How They Work Together

Encryption is an **opt-in flag** on the existing `upload source add` command:

```bash
# Plaintext source (mutability only)
guppy upload source add did:key:z6Mk... /my-backups --name backups

# Encrypted source (mutability + encryption)
guppy upload source add did:key:z6Mk... /my-secrets --name secrets --encrypt
```

**Key rotation** requires mutability because rotating keys creates new metadata CIDs that need to be tracked in Pail.

### Implementation Status

| Component | Status |
|-----------|--------|
| Block-level encryption (AES-256-CTR) | ✅ POC done |
| Pail library (Go) | ✅ Exists |
| UCN wrapper | 🔨 Needs Go impl |
| KMS integration | 🔨 TODO |
| Guppy command extensions | 🔨 TODO |

### Why This Approach?

- **Pail + UCN** = battle-tested components, already used in Storacha
- **Per-file encryption** = matches industry standard (AWS S3, Azure, GCP)
- **Minimal changes** = extends existing commands, encryption is just a flag

Closes https://github.com/storacha/project-tracking/issues/663

### DIFF
--- a/rfc/forge-encryption.md
+++ b/rfc/forge-encryption.md
@@ -1,0 +1,417 @@
+# RFC: Encryption for Forge (Guppy/Piri)
+
+**Status: Draft**
+
+## Authors
+
+- [Felipe Forbeck](https://github.com/fforbeck), [Storacha Network](https://storacha.network/)
+
+## Editors
+- [Alan Shaw](https://github.com/alanshaw), [Storacha Network](https://storacha.network/)
+- [Hannah Howard](https://github.com/hannahhoward), [Storacha Network](https://storacha.network/)
+- [Alex Kinstler](https://github.com/prodalex), [Storacha Network](https://storacha.network/)
+
+## Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC2119](https://datatracker.ietf.org/doc/html/rfc2119).
+
+## Introduction
+
+This RFC proposes the implementation of content encryption for Forge enterprise customers using the Guppy client. This enables:
+
+1. **Encryption at Rest**: Client-side encryption of file content before upload
+2. **UCAN-Gated Decryption**: Access control via delegations
+
+## Motivation
+
+Enterprise Forge customers need:
+
+- **Data protection**: Sensitive content must be encrypted before leaving the client
+- **Key management**: Integration with Storacha KMS or customer-managed KMS
+- **Access control**: Fine-grained control over who can decrypt content
+
+---
+
+## Industry Context
+
+The proposed encryption model encrypts **file content only**, leaving metadata (paths, filenames, sizes) visible. This is the standard approach used by major cloud storage providers:
+
+| Provider | Content Encrypted | Paths/Names Encrypted |
+|----------|-------------------|----------------------|
+| AWS S3 (SSE-S3, SSE-KMS) | ✅ | ❌ |
+| Azure Blob Storage | ✅ | ❌ |
+| Google Cloud Storage | ✅ | ❌ |
+
+## Implementation Status
+
+Based on Alan's POC ([storacha/guppy#376](https://github.com/storacha/guppy/pull/376)).
+
+| Component | Status | Notes |
+|-----------|--------|-------|
+| AES-256-CTR encryption | ✅ Done | `pkg/encryption/aes_256_ctr.go` |
+| Random IV per block | ✅ Done | Generated in `EncryptAES256CTR()` |
+| IV stored in metadata | ✅ Done | `pkg/preparation/dags/nodemeta/aes_256_ctr.go` |
+| Block-level splitter | ✅ Done | `pkg/preparation/dags/aes_256_ctr_splitter.go` |
+| Decryption | ✅ Done | `DecryptAES256CTR()` function |
+| KMS integration | ❌ Pending | `space/encryption/setup`, key unwrap |
+| Metadata format | ❌ Pending | `EncryptedMetadata` CBOR block |
+| UCAN delegation | ❌ Pending | `space/content/decrypt` handling |
+| Folder access control | ❌ Pending | `nb.prefix` validation |
+
+## Encryption Approach: Block-Level
+
+### How It Works
+
+Block-level encryption combines file-scoped DEKs with block-scoped IVs:
+
+- **1 DEK per file**: Each file gets a unique 256-bit AES key, generated randomly
+- **Random IV per block**: Each block within the file gets a unique 16-byte IV
+- **AES-256-CTR**: Counter mode encryption with unique keystream per block
+- **Incremental uploads**: Only changed blocks need re-encryption
+
+### DEK Lifecycle (Per File)
+
+```
+For each file to upload:
+
+1. Generate random DEK (256-bit AES key)
+        │
+        ▼
+2. For each block in file:
+   - Generate random IV (16 bytes)
+   - Encrypt block with DEK + IV
+   - Store IV in block metadata
+        │
+        ▼
+3. Wrap DEK with space's public key (RSA-OAEP)
+        │
+        ▼
+4. Store wrapped DEK in file metadata block
+        │
+        ▼
+5. Upload encrypted blocks + metadata
+```
+
+**Key insight:** The DEK is the same for all blocks within a file, but each block has a unique IV. This allows:
+- Incremental uploads (only changed blocks re-encrypted with same DEK)
+- File-level access control (revoke access to specific files)
+- TS client compatibility (same pattern as `@storacha/encrypt-upload-client`)
+
+### Security: IV Requirements
+
+**MUST**
+- Generate a new random IV (16 bytes) for every block
+- Never reuse an IV with the same DEK
+- Use cryptographically secure random number generator for IV generation
+
+These requirements apply to all encryption operations, including initial uploads and incremental re-uploads.
+
+### Why Block-Level (vs File-Level)
+
+| Approach | Incremental Uploads | KMS Calls | Metadata |
+|----------|---------------------|-----------|----------|
+| **File-level** | ❌ Re-upload entire file | 1 per file | 1 per file |
+| **Block-level** | ✅ Only changed blocks | 1 per session | IV per block |
+
+Block-level enables Guppy's existing incremental upload capability to work with encrypted content.
+
+
+## Metadata Format
+
+Encrypted content MUST include a metadata block compatible with `@storacha/encrypt-upload-client`:
+
+```typescript
+interface EncryptedMetadata {
+  encryptedDataCID: CID        // CID of encrypted content
+  encryptedSymmetricKey: string // Base64-encoded wrapped DEK
+  space: SpaceDID              // Space the content belongs to
+  path?: string                // File path (e.g., "/backups/server1/backup.tar")
+  kms: {
+    provider: string           // e.g., "storacha"
+    keyId: string              // KMS key identifier
+    algorithm: string          // e.g., "RSA-OAEP-256"
+  }
+}
+```
+
+The `path` field is RECOMMENDED for all new uploads. It enables folder-level access control via `nb.prefix` delegations (see Folder-Level Access Control section).
+
+## Streaming Support
+
+Encryption MUST support streaming for large files (1TB+) with O(1) memory usage.
+
+Go's standard library provides native support via:
+- `crypto/aes`: AES block cipher
+- `crypto/cipher`: CTR mode stream cipher
+- `io.Reader` wrapper pattern: encrypt/decrypt chunks on-the-fly without buffering entire file
+
+## Encryption Flow
+
+```
+1. guppy upload <source>
+        │
+        ▼
+2. Get space public key from KMS (space/encryption/setup)
+   → 1 KMS call per upload session
+        │
+        ▼
+3. For each file:
+   a. Generate random DEK (256-bit AES key)
+   b. For each file chunk:
+      - Generate random IV (16 bytes)
+      - Encrypt chunk with DEK + IV (AES-256-CTR)
+      - Store IV in block metadata
+   c. Build UnixFS DAG from encrypted blocks
+   d. Wrap DEK with space public key (RSA-OAEP)
+   e. Create metadata block with wrapped DEK + file path
+        │
+        ▼
+4. Upload encrypted blocks + metadata (blob/add)
+```
+
+**POC Status**
+- Step 2 (KMS): ❌ Pending
+- Step 3a (DEK generation): ❌ Pending (POC uses manually provided key)
+- Step 3b (IV + encryption): ✅ Done
+- Step 3c (UnixFS DAG): ✅ Done
+- Step 3d (DEK wrapping): ❌ Pending
+- Step 3e (metadata block): ❌ Pending
+- Step 4 (upload): ✅ Existing Guppy functionality
+
+## Decryption Flow
+
+Two decryption modes are supported:
+
+### Option A: Gateway-Side Decryption (Local Key Mode)
+
+For local/development use, the gateway can decrypt content on-the-fly:
+
+```bash
+guppy gateway serve --decryption-key /path/to/key.bin
+```
+
+```
+1. Client requests: http://localhost:3000/ipfs/<CID>
+        │
+        ▼
+2. Gateway fetches encrypted blocks from network
+        │
+        ▼
+3. Gateway decrypts each block using provided key
+   - Read IV from block (first 16 bytes)
+   - Decrypt with key + IV (AES-256-CTR)
+        │
+        ▼
+4. Gateway serves decrypted content to client
+```
+
+**POC Status:** ✅ Done (`--decryption-key` flag implemented)
+
+### Option B: Client-Side Decryption (KMS Mode)
+
+For production with access control, decryption happens client-side:
+
+```
+1. Fetch encrypted content via gateway
+   - Public: https://w3s.link/ipfs/<CID>
+   - Local: guppy gateway serve → http://localhost:3000/ipfs/<CID>
+        │
+        ▼
+2. Extract file metadata block
+        │
+        ▼
+3. Extract wrapped DEK from metadata
+        │
+        ▼
+4. Unwrap DEK via KMS (space/encryption/key/decrypt)
+   → Provide UCAN proof with space/content/decrypt delegation
+   → KMS validates nb.prefix against file path
+        │
+        ▼
+5. For each encrypted block:
+   - Read IV from block metadata
+   - Decrypt with unwrapped DEK + IV (AES-256-CTR)
+        │
+        ▼
+6. Reassemble file from decrypted blocks
+```
+
+**POC Status:**
+- Step 1 (gateway fetch): ✅ Existing infrastructure
+- Step 2-4 (metadata + KMS): ❌ Pending
+- Step 5 (IV + decryption): ✅ Done
+- Step 6 (reassemble): ✅ Standard UnixFS
+
+## Folder-Level Access Control
+
+Access control is enforced via the `nb.prefix` caveat on `space/content/decrypt` delegations:
+
+```typescript
+// Grant access to all files under /backups/server1/
+space/content/decrypt
+  with: did:key:zSpace
+  nb: { prefix: "/backups/server1/" }
+  audience: did:key:zRecipient
+```
+
+**Validation flow**
+
+1. User requests decryption with delegation containing `nb.prefix`
+2. KMS extracts `path` from encrypted metadata block
+3. KMS validates: `path.startsWith(delegation.nb.prefix)`
+4. If valid → unwrap DEK; if invalid → reject
+
+**Access control examples**
+
+| Delegation `nb.prefix` | File `path` | Access |
+|------------------------|-------------|--------|
+| `/backups/` | `/backups/server1/backup.tar` | ✅ Allowed |
+| `/backups/server1/` | `/backups/server1/backup.tar` | ✅ Allowed |
+| `/backups/server2/` | `/backups/server1/backup.tar` | ❌ Denied |
+| (none) | `/backups/server1/backup.tar` | ✅ Space-level access |
+
+**Backward compatibility**
+
+- Delegations without `nb.prefix` grant space-level access (all files)
+- Files uploaded without `path` field are treated as root (`/`) and accessible with any space-level delegation
+
+## Capabilities Needed
+
+| Capability | Status | Notes |
+|------------|--------|-------|
+| `space/encryption/setup` | Exists | Get space public key for wrapping |
+| `space/encryption/key/decrypt` | Exists | Unwrap space DEK |
+| `space/content/decrypt` | Exists | Authorization proof for decryption |
+
+## Key Management
+
+Guppy SHOULD support two key management modes.
+
+### Local Key Mode (Development/Testing)
+
+For development and testing, Guppy MAY use a locally-provided key:
+
+```yaml
+# ~/.storacha/guppy/config.yaml
+encryption:
+  enabled: true
+  mode: local
+  key: "base64-encoded-256-bit-key"  # Or path to key file
+```
+
+This mode does NOT provide access control — anyone with the key can decrypt.
+
+### KMS Mode (Production/Enterprise)
+
+For production, Guppy SHOULD use the Storacha KMS:
+
+```yaml
+encryption:
+  enabled: true
+  mode: kms
+  kms_url: "https://kms.storacha.network"
+  kms_did: "did:web:kms.storacha.network"
+```
+
+| Config | Description |
+|--------|-------------|
+| `UCAN_KMS_URL` | ucan-kms service endpoint |
+| `UCAN_KMS_DID` | ucan-kms service DID for UCAN audience |
+
+KMS mode enables:
+- UCAN-gated decryption
+- Folder-level access control via `nb.prefix`
+- Key rotation (see below)
+
+## Key Rotation
+
+Key rotation requires the **Mutability** feature (see [forge-mutability.md](./forge-mutability.md)) because:
+- Rotation creates new metadata CIDs (wrapped DEK changes)
+- The catalog/UCN must be updated to point to the new metadata CID
+- Without mutability, clients cannot discover the rotated metadata
+
+Guppy SHOULD support two types of key rotation via CLI commands (KMS mode only):
+
+### KEK Rotation (Space Key)
+
+Rotates the space-level Key Encryption Key without re-encrypting content:
+
+```bash
+guppy encryption rotate-kek --space <space-did>
+```
+
+**Process:**
+1. Generate new KEK in KMS
+2. For each encrypted file in space:
+   - Unwrap DEK with old KEK
+   - Re-wrap DEK with new KEK
+   - Create new metadata block (new CID)
+   - Update catalog entry to point to new metadata CID
+3. Content blocks remain unchanged
+
+**Use case:** Regular security hygiene, suspected KEK compromise
+
+### DEK Rotation (File Key)
+
+Rotates a file's Data Encryption Key by re-encrypting the file:
+
+```bash
+guppy encryption rotate-dek --file <path>
+```
+
+**Process:**
+1. Download and decrypt file with old DEK
+2. Generate new DEK
+3. Re-encrypt all blocks with new DEK + new IVs
+4. Create new metadata block with new wrapped DEK
+5. Upload encrypted blocks + new metadata
+6. Update catalog entry to point to new root CID
+
+**Use case:** Suspected file-level compromise, compliance requirements
+
+### Local Key Mode
+
+Key rotation is NOT supported in local key mode:
+- No catalog/UCN to track metadata CID changes
+- Users managing their own keys are responsible for rotation
+
+## Revocation
+
+Access revocation is handled by the `ucan-kms` service:
+
+1. User revokes a `space/content/decrypt` delegation by CID
+2. KMS checks revocation status on every decrypt request
+3. Revoked delegations are rejected — DEK is not unwrapped
+
+**Guppy CLI:**
+
+```bash
+guppy delegation revoke <delegation-cid>
+```
+
+**Note:** Revocation only prevents future decryption. If a user already has the DEK in memory, they can still decrypt. For full revocation, combine with DEK rotation.
+
+## Implementation Requirements
+
+| Component | Change |
+|-----------|--------|
+| `@storacha/capabilities` | Add `prefix` field to `space/content/decrypt` schema |
+| `ucan-kms` | Validate prefix in decrypt handler |
+| `guppy` | CLI command to mint scoped delegations |
+
+
+## Security Considerations
+
+1. **Client-side encryption**: Plaintext MUST never leave the Guppy client
+2. **Key management**: DEK is wrapped with space-specific KEK managed by KMS
+3. **UCAN-gated decryption**: `space/encryption/key/decrypt` delegation required
+4. **Unique IV per block**: Each block MUST have a unique random IV
+5. **Secure random**: IVs MUST be generated using cryptographically secure random
+6. **No IV reuse**: Same DEK with same IV = keystream reuse attack
+
+
+## References
+
+- [Block Encryption POC](https://github.com/storacha/guppy/pull/376)
+- [Storacha Encrypt Upload Client](https://github.com/storacha/upload-service/tree/main/packages/encrypt-upload-client)
+- [AWS S3 Client-Side Encryption](https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingClientSideEncryption.html)
--- a/rfc/forge-encryption.md
+++ b/rfc/forge-encryption.md
@@ -89,7 +89,7 @@ Block-level encryption combines file-scoped DEKs with block-scoped IVs:
 
 - **1 DEK per file**: Each file gets a unique 256-bit AES key, generated randomly
 - **Random IV per block**: Each block within the file gets a unique 16-byte IV
-- **AES-256-CTR**: Counter mode encryption with unique keystream per block
+- **AES-256-CTR**: Counter mode encryption with unique keystream per block. AES-256-CTR doesn't provide authenticated encryption on its own, but in our case its fine since we use CIDs. Any tampering with the encrypted data changes the block's CID, which propagates up and changes the root CID, so it won't go unnoticed.
 - **Incremental uploads**: Only changed blocks need re-encryption
 
 ### DEK Lifecycle (Per File)
--- a/rfc/forge-encryption.md
+++ b/rfc/forge-encryption.md
@@ -22,6 +22,8 @@ This RFC proposes the implementation of content encryption for Forge enterprise 
 1. **Encryption at Rest**: Client-side encryption of file content before upload
 2. **UCAN-Gated Decryption**: Access control via delegations
 
+**Scope:** File-level encryption only. Directory/folder encryption is not supported (each file is encrypted individually with its own DEK).
+
 ## Motivation
 
 Enterprise Forge customers need:
@@ -58,6 +60,27 @@ Based on Alan's POC ([storacha/guppy#376](https://github.com/storacha/guppy/pull
 | UCAN delegation | ❌ Pending | `space/content/decrypt` handling |
 | Folder access control | ❌ Pending | `nb.prefix` validation |
 
+## Terminology
+
+| Term | Description |
+|------|-------------|
+| **DEK** | Data Encryption Key 256-bit AES key used to encrypt file content |
+| **KEK** | Key Encryption Key space-level RSA key managed by KMS, used to wrap DEKs |
+| **IV** | Initialization Vector 16-byte random value, unique per content block |
+| **Content block** | A chunk of encrypted file data (e.g., 1MB), contains IV inline |
+| **Encrypted metadata block** | CBOR block containing wrapped DEK, path, KMS info. One per file, its CID is the "metadataCID" |
+| **Wrapping** | Encrypting a DEK with the space's KEK (RSA-OAEP) |
+
+## Encryption vs Access Control Granularity
+
+| Aspect | Granularity | Description |
+|--------|-------------|-------------|
+| **Encryption (DEK)** | Per file | Each file has its own DEK |
+| **Key wrapping (KEK)** | Per space | All DEKs in a space are wrapped with the same KEK |
+| **Access control** | Per file OR per folder | CID-based (`nb.resource`) or path-based (`nb.path`) |
+
+**Key insight:** Even though encryption is file-level, access control can be folder-level. A delegation with `nb.path: "/backups/"` grants access to decrypt **all files** whose path starts with `/backups/`, each with their own DEK.
+
 ## Encryption Approach: Block-Level
 
 ### How It Works
@@ -84,9 +107,10 @@ For each file to upload:
         │
         ▼
 3. Wrap DEK with space's public key (RSA-OAEP)
+   - "Wrapping" = encrypting the DEK with the space's KEK (Key Encryption Key)
         │
         ▼
-4. Store wrapped DEK in file metadata block
+4. Store wrapped DEK in encrypted metadata block
         │
         ▼
 5. Upload encrypted blocks + metadata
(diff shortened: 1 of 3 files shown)


hannahhoward (commented):
Overall, this is good, but there are some critical missing bits, that I think will emerge from working closely with @Peeja , @alanshaw and the existing go devs.

Specifically,

1. "Service layer" -- is this a server? a secondary process process on the dev machine? I don't either of these are a good idea, and the server approach would break a number of design principles about our system (namely that all CIDs should be generated on the client). Personally I think a language port is gonna be WAY faster (especially with AI aided dev) and product a way less complex system, so I'd argue strongly for a full port. These aren't complex libraries and I think we could have them ported in a week or two with the AI helping. And then we have a single process for a single machine, way simpler to maintain and reason about.

2. There's a bit of unspecified confusion about how Pail works in the Forge context, that I think you might need to embed with @peeja on guppy to really grok. So Guppy has a notion of "sources" -- i.e. data sources (usually large, deep directories) that get uploaded within a space. Each space has 1..n sources, and when you upload within a space, after the first upload of a source, only the "delta" gets updated-- Guppy knows how to upload just blocks to make a new updated UnixFS root. So with mutabiltiy:

- You have the list of sources which get updated, and you DEFINITELY want that to be represented by Pail + UCN.
- You have the directory tree structure within the sources itself. This is currently UnixFS and is updated properly each incremental upload.
- So the real question is about whether to use Pail for the whole directory tree, and I think that's a complicated question that merits further examination

Reasons not to use Pail:
1. These are extremely big complicated directories and Pail hasn't been tested at a scale even remotely close to working with these directories
2. The retrieval patters and general usage for Pail is totally different than for UnixFS -- so the downstream change implications of using Pail for the whole directory tree structure are unknown.

Reasons to use Pail:
1. Much more fine grained "multi-writer" capabilities are unlocked if you use Pail for everything. If you used pail for just the sources list, then you'd essentially have a last-writer-wins on a per-source level -- if source X is in state A, and two different guppies make several changes to the directory tree structure, written as UnixFS, then the directory structure would by default ONLY get the changes of the last client to write. Note: we could apply a smarter merge outside of PAIL, similar to the way I merge Markdown files in Clawracha. I actually believe this wouldn't be TOO hard.

Final sidebar: Current Guppy is also smart enough to only upload diff blocks for Files when they change.  Encryption will kill that ability I believe, unless there's some useful way to encode only changes that works for encrypted data. Worth a google.

rfc/forge-mutability-encryption.md
```diff
@@ -0,0 +1,424 @@
...
+│     → returns root CID (metadata block CID)                  │
+│                                                              │
+│  ─────────────── POST-UPLOAD (UCN + Pail) ───────────────    │
+│                                                              │
+│  8. Record in Pail: put(filePath, rootCID)                   │
+│     → returns new Pail root CID                              │
+│          │                                                   │
+│          ▼                                                   │
+│  9. Publish to UCN: Name.publish(pailRootCID)                │
+│     → mutable name now points to updated index               │
```
alanshaw, line 111: This is "pail without CRDT" - in the case of multiple concurrent updates to the same name, the UCN resolution is to just use the first of the alphabetically sorted CIDs (IIRC). It means if 2 users start with the same pail, and both make an update, only 1 wins.

The Pail CRDT library allows the two updates to be applied, only resorting to alphabetically sorted CIDs when the two updates have the same causal order and touch the same key.

rfc/forge-mutability-encryption.md
```diff
@@ -0,0 +1,424 @@
...
+│  3. Fetch CAR from gateway using content CID                 │
+│          │                                                   │
+│          ▼                                                   │
+│  4. Extract metadata block from CAR (CBOR)                   │
+│     - encryptedDataCID                                       │
+│     - encryptedSymmetricKey                                  │
+│     - KMS info                                               │
+│          │                                                   │
+│          ▼                                                   │
+│  5. Extract encrypted content from CAR using encryptedDataCID│
```
alanshaw, line 200: Why don't we encrypt each block?

alanshaw: POC: https://github.com/storacha/guppy/pull/376

fforbeck:
After the feedback received and the new POC completed by @alanshaw, I decided to break it into 2 RFCs:
- [📖 Forge Encryption RFC](https://github.com/storacha/RFC/blob/rfc/privacy-mutability-forge/rfc/forge-encryption.md)
- [📖 Forge Mutability RFC](https://github.com/storacha/RFC/blob/rfc/privacy-mutability-forge/rfc/forge-mutability.md)

rfc/forge-encryption.md
```diff
@@ -0,0 +1,417 @@
...
+5. For each encrypted block:
+   - Read IV from block metadata
+   - Decrypt with unwrapped DEK + IV (AES-256-CTR)
+        │
+        ▼
+6. Reassemble file from decrypted blocks
+```
+
+**POC Status:**
+- Step 1 (gateway fetch): ✅ Existing infrastructure
```
hannahhoward, line 276: no gateway in the forge context -- direct retrieval from SPs

rfc/forge-encryption.md
```diff
@@ -0,0 +1,417 @@
...
+## Key Rotation
+
+Key rotation requires the **Mutability** feature (see [forge-mutability.md](./forge-mutability.md)) because:
+- Rotation creates new metadata CIDs (wrapped DEK changes)
+- The catalog/UCN must be updated to point to the new metadata CID
+- Without mutability, clients cannot discover the rotated metadata
+
+Guppy SHOULD support two types of key rotation via CLI commands (KMS mode only):
+
+### KEK Rotation (Space Key)
```
hannahhoward, line 390: This is pretty cool that we can rotate the core RSA key without reencrypting all the files.

rfc/forge-encryption.md
```diff
@@ -0,0 +1,417 @@
...
+guppy delegation revoke <delegation-cid>
+```
+
+**Note:** Revocation only prevents future decryption. If a user already has the DEK in memory, they can still decrypt. For full revocation, combine with DEK rotation.
+
+## Implementation Requirements
+
+| Component | Change |
+|-----------|--------|
+| `@storacha/capabilities` | Add `prefix` field to `space/content/decrypt` schema |
```
hannahhoward: make sure to note we have to port over these capabilites to golang

rfc/forge-encryption.md
```diff
@@ -0,0 +1,417 @@
...
+2. KMS checks revocation status on every decrypt request
+3. Revoked delegations are rejected — DEK is not unwrapped
+
+**Guppy CLI:**
+
+```bash
+guppy delegation revoke <delegation-cid>
+```
+
+**Note:** Revocation only prevents future decryption. If a user already has the DEK in memory, they can still decrypt. For full revocation, combine with DEK rotation.
```
hannahhoward: This is worth documenting when we get to docs

rfc/forge-encryption.md
```diff
@@ -0,0 +1,417 @@
...
+# ~/.storacha/guppy/config.yaml
+encryption:
+  enabled: true
+  mode: local
+  key: "base64-encoded-256-bit-key"  # Or path to key file
+```
+
+This mode does NOT provide access control — anyone with the key can decrypt.
+
+### KMS Mode (Production/Enterprise)
```
hannahhoward, line 359: I believe all external forge users will be in this tier

rfc/forge-mutability.md
```diff
@@ -0,0 +1,169 @@
...
+
+Enterprise Forge customers need:
+
+- **Mutable references**: Backup workflows need stable names that update to point to the latest backup version
+- **Cross-client sync**: Multiple Guppy instances should be able to resolve the same reference
+- **On-network state**: Track what has been uploaded without relying on local-only databases
+
+## Approaches Under Consideration
+
+### Option A: Simple Catalog (Alex's Proposal)
```
hannahhoward: So:
I do not believe a single catalog file is a good idea for the size of Forge directories (potentially several thousands or up to a million files/folders) -- an update would mean a large upload just to change a single value.

Yes, you could start chunking past a certain size, but then you're chunking up a key value list and essentially reimplementing Pail from scratch. Pail already considers chunking and all that comes with it -- like inserts and deletes, as well as sorting and range querying.

And https://github.com/storacha/go-pail already exists

rfc/forge-mutability.md
```diff
@@ -0,0 +1,169 @@
...
+| **Catalog size** | Works for 100k+ files | Optimized for millions |
+| **Conflict resolution** | Manual (user re-uploads) | Automatic (CRDT merge) |
+
+**Recommendation:** Start with Option A for Guppy CLI (single-user tool). Option B becomes valuable when enabling team collaboration with concurrent uploads from multiple clients.
+
+### Option B: Go Ports (Hannah's Proposal)
+
+**How it works**
+
+- **Namespace:** UCN Names - ed25519 keypairs that can be delegated and shared
```
hannahhoward: UCN is a bit of a misnomer as a library one needs to port -- it's all of a few hundred lines of code and is really just scaffolding to provide a simple UI around clock/head + clock/advance.

Ultimately I think we can just use UCN + Pail

rfc/forge-mutability.md
```diff
@@ -0,0 +1,169 @@
...
+- **State Index:** Pail - sharded Merkle trie for `path → CID` mappings
```
hannahhoward: This is an interesting question as it relates to uploads -- we upload large folders as UnixFS -- this is important for retrieval as all software tends to assume UnixFS.

We've discussed various proposals for putting the whole index in Pail vs putting only the uploads in UnixFS. 

I think we could get by with just Pail for the upload list, or, if we want to do pail for everything, I think it would make sense to keep uploading in UnixFS (for traditional retrieval patterns) but also store all the file CIDs in Pail

hannahhoward (changes requested):
I'm pretty set on using Pail, especially since the go library exists, but if @alanshaw disagrees I'd be open to using the catalog approach.

rfc/forge-encryption.md
fforbeck: @hannahhoward, when we run the `guppy gateway serve`, is that how the SPs will allow retrievals? Through this "gateway".

rfc/forge-encryption.md
fforbeck: done

rfc/forge-encryption.md
fforbeck: sounds good. I will add a note about it.

rfc/forge-encryption.md
hannahhoward: yes

rfc/forge-encryption.md
```diff
@@ -0,0 +1,418 @@
...
+### Security: IV Requirements
+
+**MUST**
+- Generate a new random IV (16 bytes) for every block
+- Never reuse an IV with the same DEK
+- Use cryptographically secure random number generator for IV generation
+
+These requirements apply to all encryption operations, including initial uploads and incremental re-uploads.
+
+### Why Block-Level (vs File-Level)
```
alanshaw, line 135: IDK if worth mentioning but there is obviously a small overhead in storage size with adding a unique IV to each block.

rfc/forge-encryption.md
```diff
@@ -0,0 +1,418 @@
...
+|----------|---------------------|-----------|----------|
+| **File-level** | ❌ Re-upload entire file | 1 per file | 1 per file |
+| **Block-level** | ✅ Only changed blocks | 1 per session | IV per block |
+
+Block-level enables Guppy's existing incremental upload capability to work with encrypted content.
+
+
+## Metadata Format
+
+Encrypted content MUST include a metadata block compatible with `@storacha/encrypt-upload-client`:
```
alanshaw: I'm not clear if this is 1 metadata block per upload or per file?

rfc/forge-encryption.md
```diff
@@ -0,0 +1,418 @@
...
+        ▼
+4. Upload encrypted blocks + metadata (blob/add)
+```
+
+**POC Status**
+- Step 2 (KMS): ❌ Pending
+- Step 3a (DEK generation): ❌ Pending (POC uses manually provided key)
+- Step 3b (IV + encryption): ✅ Done
+- Step 3c (UnixFS DAG): ✅ Done
+- Step 3d (DEK wrapping): ❌ Pending
```
alanshaw, line 213: What does "wrapping" mean here? Encrypting the decryption key with a different key?

rfc/forge-encryption.md
```diff
@@ -0,0 +1,418 @@
...
+   - Read IV from block (first 16 bytes)
+   - Decrypt with key + IV (AES-256-CTR)
+        │
+        ▼
+4. Gateway serves decrypted content to client
+```
+
+**POC Status:** ✅ Done (`--decryption-key` flag implemented)
+
+### Option B: Client-Side Decryption (KMS Mode)
```
alanshaw, line 246: Do you mean `guppy retrieve`?

rfc/forge-encryption.md
```diff
@@ -0,0 +1,418 @@
...
+
+## Folder-Level Access Control
+
+Access control is enforced via the `nb.prefix` caveat on `space/content/decrypt` delegations:
+
+```typescript
+// Grant access to all files under /backups/server1/
+space/content/decrypt
+  with: did:key:zSpace
+  nb: { prefix: "/backups/server1/" }
```
alanshaw: Is this space wide? Do we not have to include the CID of the content?

rfc/forge-encryption.md
```diff
@@ -0,0 +1,418 @@
...
+  audience: did:key:zRecipient
+```
+
+**Validation flow**
+
+1. User requests decryption with delegation containing `nb.prefix`
```
alanshaw: What if you want to give access to just a single file?

rfc/forge-encryption.md
```diff
@@ -0,0 +1,418 @@
...
+2. KMS extracts `path` from encrypted metadata block
+3. KMS validates: `path.startsWith(delegation.nb.prefix)`
```
alanshaw: If I have a directory `/priv` and a file `/priv.txt`, does `prefix: '/priv'` give me access to both?

I feel like `prefix` should be `path` - i.e. it is a path to a file or directory, not just a prefix. So in the example `path: '/priv'` does _not_ match `/priv.txt`.

rfc/forge-encryption.md
alanshaw: Might be nice if this was a _list_? Then you don't need multiple delegations...

rfc/forge-encryption.md
```diff
@@ -0,0 +1,418 @@
...
+|------------------------|-------------|--------|
+| `/backups/` | `/backups/server1/backup.tar` | ✅ Allowed |
+| `/backups/server1/` | `/backups/server1/backup.tar` | ✅ Allowed |
+| `/backups/server2/` | `/backups/server1/backup.tar` | ❌ Denied |
+| (none) | `/backups/server1/backup.tar` | ✅ Space-level access |
+
+**Backward compatibility**
+
+- Delegations without `nb.prefix` grant space-level access (all files)
+- Files uploaded without `path` field are treated as root (`/`) and accessible with any space-level delegation
```
alanshaw: I don't understand what this means.

rfc/forge-encryption.md
```diff
@@ -0,0 +1,418 @@
...
+
+Rotates the space-level Key Encryption Key without re-encrypting content:
+
+```bash
+guppy encryption rotate-kek --space <space-did>
+```
+
+**Process:**
+1. Generate new KEK in KMS
+2. For each encrypted file in space:
```
alanshaw: Where is the list of encrypted files?

rfc/forge-encryption.md
```diff
@@ -0,0 +1,418 @@
...
+  nb: { prefix: "/backups/server1/" }
+  audience: did:key:zRecipient
+```
+
+**Validation flow**
+
+1. User requests decryption with delegation containing `nb.prefix`
+2. KMS extracts `path` from encrypted metadata block
+3. KMS validates: `path.startsWith(delegation.nb.prefix)`
+4. If valid → unwrap DEK; if invalid → reject
```
alanshaw: How does the service know what the wrapped decryption key is? Does it have to fetch the content? Does it need a retrieval delegation?

rfc/forge-encryption.md
```diff
@@ -0,0 +1,418 @@
...
+guppy encryption rotate-kek --space <space-did>
+```
+
+**Process:**
+1. Generate new KEK in KMS
+2. For each encrypted file in space:
+   - Unwrap DEK with old KEK
+   - Re-wrap DEK with new KEK
+   - Create new metadata block (new CID)
+   - Update catalog entry to point to new metadata CID
```
alanshaw: ...but the old metadata block is still in the network? How do we know that we shouldn't decrypt an wrapped DEK in old metadata?

fforbeck:
@hannahhoward & @alanshaw - I've updated the Mutability RFC and addressed Hannah's suggestions.

alanshaw (changes requested):

rfc/forge-encryption.md
fforbeck: I will add it. It is small but important.

rfc/forge-encryption.md
fforbeck: It is 1 metadata block per file. Each file has its own DEK and metadata.

rfc/forge-encryption.md
fforbeck: wrapping = encrypting the DEK (Data Encryption Key) with the space's public key (RSA-OAEP). The KEK (Key Encryption Key) encrypts the DEK (Data Encryption Key).

rfc/forge-mutability.md
```diff
@@ -0,0 +1,218 @@
...
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC2119](https://datatracker.ietf.org/doc/html/rfc2119).
+
+## Introduction
+
+This RFC proposes the implementation of mutability features for Forge enterprise customers using the Guppy client. These features enable:
+
+1. **Mutable References**: Stable pointers that update to the latest version of uploaded content
+2. **Content Catalog**: Track uploaded content (path to CID mappings) across clients
+
+**Related:** [forge-encryption.md](./forge-encryption.md) - Encryption key rotation depends on mutability to track metadata CID changes.
```
alanshaw, line 26: Perhaps worth being a bit more explicit about this in the encryption doc. Is it just this? It does not depend on mutability otherwise?

rfc/forge-encryption.md
fforbeck: Yes. Fixed.

rfc/forge-mutability.md
```diff
@@ -0,0 +1,218 @@
...
+```mermaid
+sequenceDiagram
+    participant User
+    participant Guppy
+    participant Workers
+    participant Storacha
+    participant Pail
+    participant UCN
+    
+    User->>Guppy: guppy upload <space> <source-path>
```
alanshaw: How do you choose the pail path? For example you probably do not want your pail paths to be `/Users/alan/my-private-datas`. I guess when you add a source to guppy you'd specify the "name" for the upload?

I have previously built a [CLI extension](https://github.com/alanshaw/w3cli-plugin-bucket/blob/main/index.js) where we literally treat the mutabaility stuff seperately. So you could maybe have `guppy bucket put <space> <bucket-path> <source-path>`?

rfc/forge-encryption.md
fforbeck: It is. We do need the CID of the content. The capability in the TS lib already has that, but I didn't include it here. Just fixed that.

rfc/forge-encryption.md
fforbeck: The existing `space/content/decrypt` capability uses `nb.resource` (CID-based) to control the decryption of that file. The idea with this new attribute is to enhance the capability with an optional `nb.prefix` field for path-based access control. So you don't need to set the prefix at all, you would delegate and provide the file/content CID.

rfc/forge-encryption.md
fforbeck: If a delegation specifies `nb.prefix` (I will rename it to `nb.path`), the file's metadata MUST contain a path field to validate against. Old files without a path in metadata cannot be accessed using path-scoped delegations, so we use a CID-only delegation instead (like we already do today).

rfc/forge-encryption.md
fforbeck: This enables granular revocation. Revoking access to one path doesn't affect others.

rfc/forge-encryption.md
fforbeck: Good catch. Thanks.
We will need to enforce that `nb.path` MUST end with `/` to ensure proper directory matching.

rfc/forge-mutability.md
```diff
@@ -0,0 +1,218 @@
...
+    Workers->>Workers: Pack into CAR shards
+    Workers->>Storacha: space/blob/add (shards)
+    Workers->>Storacha: space/blob/add (indexes)
+    Workers->>Storacha: upload/add
+    Workers-->>Guppy: rootCID
+    
+    Note over Guppy,UCN: NEW: Mutability integration
+    Guppy->>Pail: crdt.Put(sourcePath, rootCID)
+    Pail-->>Guppy: eventCID
+    Guppy->>UCN: name.Publish(eventCID)
```
alanshaw: I guess the problem is that you lose the ability to merge multiple concurrent changes within the _value_. Like, if I have a path `/path/to: <unixfs-cid>` and I 2 people change the unixfs data then there is no merge, only lucky ordered CID gets picked first.

If we want multiwriter that _includes_ the VALUE then we should either use Pail for the whole tree, or use unixfs for the whole tree. The latter involves writing code to CRDT merge a UnixFS directory structure.

However, I don't really know how it would work when 2 writers change the _value_ - you'd kinda have to sync the data to disk since guppy works from on-disk sources. Otherwise each writer would keep overwriting each others changes. So, with that in mind I'd suggest that we don't allow multiwriter that includes the value...and so I think Pail + UnixFS values seems sensible.

rfc/forge-mutability.md
```diff
@@ -0,0 +1,218 @@
...
+sequenceDiagram
+    participant User
+    participant Guppy
+    participant UCN
+    participant Pail
+    participant Locator
+    participant Storage
+    participant KMS
+    
+    User->>Guppy: guppy retrieve <space> <path> <output>
```
alanshaw: I would suggest `guppy bucket get <space> <path>` in order to not overload the retrieval command.

alanshaw (changes requested):

rfc/forge-encryption.md
fforbeck: I had in mind the file structure we have in the client that uploaded the data, but retrievals and rotation will probably not happen in the same client, so I believe we will need to store encrypted file paths in Pail - and that pushes us back to that idea of `root path` vs `all file paths` in Pail.

fforbeck: The current encryption is file-level only (no folder encryption support in the TS/JS side). We can preserve this idea for v1: each encrypted file gets one Pail entry (path -> metadataCID), and key rotation uses `Pail.Entries()` to iterate over all encrypted files.

rfc/forge-encryption.md
fforbeck: We could combine the `path` with the `DEK` before wrapping it, and then use the wrapped DEK in KMS. WDYT?

rfc/forge-encryption.md
fforbeck: Yes, still in the network, but the old KEK is deleted/rotated in KMS, so even if someone has the old metadata CID, they can't unwrap the old wrapped DEK because the old KEK no longer exists in KMS. In KMS, we can always use the latest KEK to perform unwrapping.

fforbeck:
@alanshaw & @hannahhoward - we talked about starting v1 storing only the root path in Pail to handle mutability, I think that makes sense for the non-encrypted files. However, for the encryption, we kind of need the full path to handle the key rotation process. 

For **encrypted uploads**, we need `filePath -> metadataCID` entries because:
- Each file has its own DEK (wrapped in its encrypted metadata block)
- KEK rotation needs to iterate all encrypted files to re-wrap DEKs
- Path-based access control (`nb.path`) needs to validate file paths

**Proposal**

| Upload Type | Pail Entry | Rationale |
|-------------|------------|-----------|
| **Raw** | `rootPath -> rootCID` | File structure in UnixFS, traverse from root |
| **Encrypted** | `filePath -> metadataCID` | Each file has its own DEK, needs individual entries for rotation & access control |


Does this dual strategy make sense, or should we unify?

rfc/forge-encryption.md
```diff
@@ -0,0 +1,469 @@
...
+
+## Encryption Approach: Block-Level
+
+### How It Works
+
+Block-level encryption combines file-scoped DEKs with block-scoped IVs:
+
+- **1 DEK per file**: Each file gets a unique 256-bit AES key, generated randomly
+- **Random IV per block**: Each block within the file gets a unique 16-byte IV
+- **AES-256-CTR**: Counter mode encryption with unique keystream per block
```
BravoNatalie: I think it’s worth pointing out that AES-256-CTR doesn’t provide authenticated encryption on its own, but in our case that’s fine since we use CIDs. Any tampering with the encrypted data changes the block’s CID, which propagates up and changes the root CID, so it won’t go unnoticed.

rfc/forge-mutability.md
fforbeck: I love the idea of having a new command to handle mutability and encryption independently. Yes, let's use that. I've updated the RFC with bucket create, put, and get commands.

rfc/forge-mutability.md
fforbeck: Reverted to keep it under the retrieve command.

rfc/forge-mutability.md
fforbeck: Reverted to keep it under the source and upload commands as discussed in the call.

alanshaw:
# RFC: An alternate interface for mutable content in Guppy

My main concern with adding a "name" to a "source" is that it's not an explicit action. It's a passive attribute of the source that we've decided will be used after the data is uploaded to record a name for the uploaded DAG. The name is attached to a collection of static files AND it is being used as a mutable pointer that MAY point to the DAG created by said source.

Lets walk though the proposed interface:

```sh
$ guppy upload source add did:key:zSpace "My Stuff" /path/to/alans/stuff
```

When I type `guppy upload did:key:zSpace` what do I expect to happen? All my named sources get uploaded and appear in my `guppy ls` output:

```sh
$ guppy upload did:key:zSpace
bafyAlansStuff

$ guppy ls did:key:zSpace
My Stuff    bafyAlansStuff
```

I delegate access to `did:key:zSpace` to Felipe.

Felipe runs:

```sh
$ guppy upload source add did:key:zSpace "My Stuff" /path/to/felipes/stuff
```

Then Felipe runs `guppy upload did:key:zSpace`:

```sh
$ guppy upload did:key:zSpace
bafyFelipesStuff

$ guppy ls did:key:zSpace
My Stuff    bafyFelipesStuff
```

What happens when if I want to update "My Stuff" to point back at _my_ stuff not Felipe's stuff? Can I run `guppy upload did:key:zSpace`? No, I believe guppy will just think everything is done already. Do I need to add a new source with the same name?

This is an odd situation.
 
We could make `guppy upload did:key:zSpace` always update the name to the value we think it should be...but then what happens with old sources that I _don't_ want to update? i.e. Felipe's stuff is the right stuff, but I just typed `guppy upload did:key:zSpace` to upload a _different_ source I added e.g. "My Pictures" and it overwrites "My Stuff" as well!

One possible solution to prevent issues like this is for `guppy upload` to ALWAYS take a file path/source directory.

However at this point I find myself questioning the exposure of sources in the CLI UX at all. Why do I need to create a source before uploading it? If I am always going to reference the file path in the _upload_ command, why doesn't it just create or use the source for the directory I provide to `guppy upload <space> <path>`?

...but if we don't have sources where does the name go? Well, what if we provide a name to `guppy upload`?

```sh
$ guppy upload did:key:zSpace "My Stuff" /path/to/alans/stuff
```

This feels better, _upload_ is an action and it looks like we're asking the name of this upload to be "My Stuff" and I feel it's more concievable that this will overwrite another upload with the same name. Note that this also allows the _same_ source to be put to multiple paths. e.g.

```sh
$ guppy upload did:key:zSpace "Backups/My Stuff/2026-04-20" /path/to/alans/stuff
```

> it's more concievable that this will overwrite another upload with the same name

IMHO it's better, and I could probably get behind this. However, the name for this upload is not an attribute, it's more like an ID, a path or a key i.e. if you upload with the same key it'll overwrite. The problem is we've mixed uploading files and managing buckets. The word "upload" makes it look like the name is an attribute not an identity. Also calling it "name" and not "key".

One way we could easily rectify this would be to switch to using bucket semantics for our CLI commands. i.e. instead of `upload` we prefer `put`, and instead of `name` we use `key`. I think this more accurately describes the action.

The magic is that when you provide a file path the CLI will automatically create a source and upload that data for you. This is expected, and is no different from say, the AWS CLI.

So then we can distil our command to simply:

```sh
$ guppy put <space> <key> <path>
```

* It's incredibly short.
* It accurately describes the action.
* It combines what previously was two commands (`guppy upload source add` and `guppy upload`) into one.
* It adds multabiliy as a first class citizen.

Side note, this also allows users to construct buckets that contain data that they didn't directly upload to a space e.g. we might in the future support `<path-or-cid>`.

## Proposal Summary

To Summarize, my proposal would be:

* Add `guppy put <space> <key> <path>`.
    * It'll add `<path>` as a source automatically if it does not exist.
    * It'll always set the value of name to the root CID of the DAG generated by `<path>`, even if the source thinks the DAG has been fully uploaded.
* Change `guppy ls` to list the keys.
* Add `guppy get <space> <key>` to retrieve a value for a key
* Remove `guppy retrieve` - mutable references are the default.
* Remove `guppy upload` - mutable references are the default.
* Remove `guppy upload source add` command in favour of sources being created as a result of a put action.
* Optionally retain source `ls`/`rm` to allow users to manually garbage collect their DB.
    * Move `guppy upload source` to `guppy source`.

fforbeck:
> # RFC: An alternate interface for mutable content in Guppy
> My main concern with adding a "name" to a "source" is that it's not an explicit action. It's a passive attribute of the source that we've decided will be used after the data is uploaded to record a name for the uploaded DAG. The name is attached to a collection of static files AND it is being used as a mutable pointer that MAY point to the DAG created by said source.
> 
> Lets walk though the proposed interface:
> 
> ```shell
> $ guppy upload source add did:key:zSpace "My Stuff" /path/to/alans/stuff
> ```
> 
> When I type `guppy upload did:key:zSpace` what do I expect to happen? All my named sources get uploaded and appear in my `guppy ls` output:
> 
> ```shell
> $ guppy upload did:key:zSpace
> bafyAlansStuff
> 
> $ guppy ls did:key:zSpace
> My Stuff    bafyAlansStuff
> ```
> 
> I delegate access to `did:key:zSpace` to Felipe.
> 
> Felipe runs:
> 
> ```shell
> $ guppy upload source add did:key:zSpace "My Stuff" /path/to/felipes/stuff
> ```
> 
> Then Felipe runs `guppy upload did:key:zSpace`:
> 
> ```shell
> $ guppy upload did:key:zSpace
> bafyFelipesStuff
> 
> $ guppy ls did:key:zSpace
> My Stuff    bafyFelipesStuff
> ```
> 
> What happens when if I want to update "My Stuff" to point back at _my_ stuff not Felipe's stuff? Can I run `guppy upload did:key:zSpace`? No, I believe guppy will just think everything is done already. Do I need to add a new source with the same name?
> 
> This is an odd situation.
> 
> We could make `guppy upload did:key:zSpace` always update the name to the value we think it should be...but then what happens with old sources that I _don't_ want to update? i.e. Felipe's stuff is the right stuff, but I just typed `guppy upload did:key:zSpace` to upload a _different_ source I added e.g. "My Pictures" and it overwrites "My Stuff" as well!
> 
> One possible solution to prevent issues like this is for `guppy upload` to ALWAYS take a file path/source directory.
> 
> However at this point I find myself questioning the exposure of sources in the CLI UX at all. Why do I need to create a source before uploading it? If I am always going to reference the file path in the _upload_ command, why doesn't it just create or use the source for the directory I provide to `guppy upload <space> <path>`?
> 
> ...but if we don't have sources where does the name go? Well, what if we provide a name to `guppy upload`?
> 
> ```shell
> $ guppy upload did:key:zSpace "My Stuff" /path/to/alans/stuff
> ```
> 
> This feels better, _upload_ is an action and it looks like we're asking the name of this upload to be "My Stuff" and I feel it's more concievable that this will overwrite another upload with the same name. Note that this also allows the _same_ source to be put to multiple paths. e.g.
> 
> ```shell
> $ guppy upload did:key:zSpace "Backups/My Stuff/2026-04-20" /path/to/alans/stuff
> ```
> 
> > it's more concievable that this will overwrite another upload with the same name
> 
> IMHO it's better, and I could probably get behind this. However, the name for this upload is not an attribute, it's more like an ID, a path or a key i.e. if you upload with the same key it'll overwrite. The problem is we've mixed uploading files and managing buckets. The word "upload" makes it look like the name is an attribute not an identity. Also calling it "name" and not "key".
> 
> One way we could easily rectify this would be to switch to using bucket semantics for our CLI commands. i.e. instead of `upload` we prefer `put`, and instead of `name` we use `key`. I think this more accurately describes the action.
> 
> The magic is that when you provide a file path the CLI will automatically create a source and upload that data for you. This is expected, and is no different from say, the AWS CLI.
> 
> So then we can distil our command to simply:
> 
> ```shell
> $ guppy put <space> <key> <path>
> ```
> 
> * It's incredibly short.
> * It accurately describes the action.
> * It combines what previously was two commands (`guppy upload source add` and `guppy upload`) into one.
> * It adds multabiliy as a first class citizen.
> 
> Side note, this also allows users to construct buckets that contain data that they didn't directly upload to a space e.g. we might in the future support `<path-or-cid>`.
> 
> ## Proposal Summary
> To Summarize, my proposal would be:
> 
> * Add `guppy put <space> <key> <path>`.
>   
>   * It'll add `<path>` as a source automatically if it does not exist.
>   * It'll always set the value of name to the root CID of the DAG generated by `<path>`, even if the source thinks the DAG has been fully uploaded.
> * Change `guppy ls` to list the keys.
> * Add `guppy get <space> <key>` to retrieve a value for a key
> * Remove `guppy retrieve` - mutable references are the default.
> * Remove `guppy upload` - mutable references are the default.
> * Remove `guppy upload source add` command in favour of sources being created as a result of a put action.
> * Optionally retain source `ls`/`rm` to allow users to manually garbage collect their DB.
>   
>   * Move `guppy upload source` to `guppy source`.

@alanshaw - I really liked the idea of having new commands to handle the bucket operations, that looks more natural and easy to reason about, plus the `key` attribute solution is quite elegant. I do think we should move into this direction, and I am updating the RFC with this proposal. In addition to that, I like the fact that it will be more similar to the S3/GCS bucket operations, which adds less overhead and improves the DX.

BravoNatalie:
> # RFC: An alternate interface for mutable content in Guppy
> My main concern with adding a "name" to a "source" is that it's not an explicit action. It's a passive attribute of the source that we've decided will be used after the data is uploaded to record a name for the uploaded DAG. The name is attached to a collection of static files AND it is being used as a mutable pointer that MAY point to the DAG created by said source.
> 
> Lets walk though the proposed interface:
> 
> ```shell
> $ guppy upload source add did:key:zSpace "My Stuff" /path/to/alans/stuff
> ```
> 
> When I type `guppy upload did:key:zSpace` what do I expect to happen? All my named sources get uploaded and appear in my `guppy ls` output:
> 
> ```shell
> $ guppy upload did:key:zSpace
> bafyAlansStuff
> 
> $ guppy ls did:key:zSpace
> My Stuff    bafyAlansStuff
> ```
> 
> I delegate access to `did:key:zSpace` to Felipe.
> 
> Felipe runs:
> 
> ```shell
> $ guppy upload source add did:key:zSpace "My Stuff" /path/to/felipes/stuff
> ```
> 
> Then Felipe runs `guppy upload did:key:zSpace`:
> 
> ```shell
> $ guppy upload did:key:zSpace
> bafyFelipesStuff
> 
> $ guppy ls did:key:zSpace
> My Stuff    bafyFelipesStuff
> ```
> 
> What happens when if I want to update "My Stuff" to point back at _my_ stuff not Felipe's stuff? Can I run `guppy upload did:key:zSpace`? No, I believe guppy will just think everything is done already. Do I need to add a new source with the same name?
> 
> This is an odd situation.
> 
> We could make `guppy upload did:key:zSpace` always update the name to the value we think it should be...but then what happens with old sources that I _don't_ want to update? i.e. Felipe's stuff is the right stuff, but I just typed `guppy upload did:key:zSpace` to upload a _different_ source I added e.g. "My Pictures" and it overwrites "My Stuff" as well!
> 
> One possible solution to prevent issues like this is for `guppy upload` to ALWAYS take a file path/source directory.
> 
> However at this point I find myself questioning the exposure of sources in the CLI UX at all. Why do I need to create a source before uploading it? If I am always going to reference the file path in the _upload_ command, why doesn't it just create or use the source for the directory I provide to `guppy upload <space> <path>`?
> 
> ...but if we don't have sources where does the name go? Well, what if we provide a name to `guppy upload`?
> 
> ```shell
> $ guppy upload did:key:zSpace "My Stuff" /path/to/alans/stuff
> ```
> 
> This feels better, _upload_ is an action and it looks like we're asking the name of this upload to be "My Stuff" and I feel it's more concievable that this will overwrite another upload with the same name. Note that this also allows the _same_ source to be put to multiple paths. e.g.
> 
> ```shell
> $ guppy upload did:key:zSpace "Backups/My Stuff/2026-04-20" /path/to/alans/stuff
> ```
> 
> > it's more concievable that this will overwrite another upload with the same name
> 
> IMHO it's better, and I could probably get behind this. However, the name for this upload is not an attribute, it's more like an ID, a path or a key i.e. if you upload with the same key it'll overwrite. The problem is we've mixed uploading files and managing buckets. The word "upload" makes it look like the name is an attribute not an identity. Also calling it "name" and not "key".
> 
> One way we could easily rectify this would be to switch to using bucket semantics for our CLI commands. i.e. instead of `upload` we prefer `put`, and instead of `name` we use `key`. I think this more accurately describes the action.
> 
> The magic is that when you provide a file path the CLI will automatically create a source and upload that data for you. This is expected, and is no different from say, the AWS CLI.
> 
> So then we can distil our command to simply:
> 
> ```shell
> $ guppy put <space> <key> <path>
> ```
> 
> * It's incredibly short.
> * It accurately describes the action.
> * It combines what previously was two commands (`guppy upload source add` and `guppy upload`) into one.
> * It adds multabiliy as a first class citizen.
> 
> Side note, this also allows users to construct buckets that contain data that they didn't directly upload to a space e.g. we might in the future support `<path-or-cid>`.
> 
> ## Proposal Summary
> To Summarize, my proposal would be:
> 
> * Add `guppy put <space> <key> <path>`.
>   
>   * It'll add `<path>` as a source automatically if it does not exist.
>   * It'll always set the value of name to the root CID of the DAG generated by `<path>`, even if the source thinks the DAG has been fully uploaded.
> * Change `guppy ls` to list the keys.
> * Add `guppy get <space> <key>` to retrieve a value for a key
> * Remove `guppy retrieve` - mutable references are the default.
> * Remove `guppy upload` - mutable references are the default.
> * Remove `guppy upload source add` command in favour of sources being created as a result of a put action.
> * Optionally retain source `ls`/`rm` to allow users to manually garbage collect their DB.
>   
>   * Move `guppy upload source` to `guppy source`.

Like Felipe, I agree this makes the DX more intuitive and clears up the confusion around name as an identity. That said, I think there are a couple of things worth thinking about before moving forward with this change.

The current two-step approach lets users add sources over time and then run uploads cumulatively, instead of just doing one-off uploads. Guppy also supports different configs per source (like shard size). That might not be a big deal to drop, but it really depends on whether users actually use different shard sizes per source, I’m not sure if they do.

